PR #4, opened 2017-09-12
Fix and update fr (French) translation

milouse:


gsantner:
Thank you very much!

Issue #10, opened 2017-09-17
Editor forgets contents when switching back from preview with a newly created note.


sjmduncan:
#### General information

* **Device:**   Moto G 4G
* **Android Version:**  Android 7.1 (LineageOS 14.1)
* **App version:** v0.1.1

#### Steps to reproduce

1. Create a new file, add some text, give it a name
2. Switch to preview
3. Switch back to edit

#### Expected result
The editor shows the information entered in step 1.

Instead the original file is saved but the editor shows a blank file as if you were creating a new one.

For existing notes this works as expected.

gsantner:
can be fixed by doing a save call before switching. feel free to send a PR

sjmduncan:
Alright, just waiting on the toolchain now.

PR #12, opened 2017-09-18
Fix the forgetful preview (#10) and some improvements to note saving and updating

### LINKED ISSUE #8: Changing case of the file name deletes the file
Version 0.1.1 (F-Droid)

Changing the case of one or more letters of the filename from within Markor will delete the file.

For example

Notes -> notes
Notes -> NoTes
Notes -> nOTES
...

Will delete Notes.md while

Notes -> Mynotes
Notes -> note
Notes -> Motes
...

Are safe renames.

sjmduncan:
Fix for #10.
Fix #8 for the case where the renaming is done by changing the title in the editor view.
Tidy up and clarify the `saveNote` method.


gsantner:
Thanks!
merigng.

PR #15, opened 2017-09-18
Better french translation

milouse:
Thanks to your previous release, I've been able to live test some of my previous fixes. Here is a second batch of translation fix.

I also allowed me to add my name in the Contributor file. Feel free to only cherry-pick the translation fix commit if you don't want my name to appear in the contributor file.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -8,7 +8,7 @@ Where:
   * Reference: E-Mail, Webpage
   * Text: Information about / kind of contribution
 
-
+  
 
 ## LIST OF CONTRIBUTORS
 -->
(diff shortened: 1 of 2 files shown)


gsantner:
Hi @milouse , thanks.
Please do revert the CONTRIBUTORS file to the current master, and just add your line below. Your change seemed to messed up others CONTRIBUTOR lines

gsantner:
@milouse  Are you okay with translations/localizations to be licensed Creative Commons Zero (https://creativecommons.org/publicdomain/zero/1.0/deed.de)? I'm using this license in all of my FOSS android apps and want to use it here too for translations. This allows me to swap some translation between projects :). 

Because french was done by many different people I cannot get the answer by everybody so I had to remove it. This means french has to be translated fully again, which will be then fully free :). Those new strings you recently added (wrote yourself completly) you could copy over if you want (not the modified ones). If you want and like to translate it I'm gonna merge it then.

Thanks!

milouse:
Hi!

I have no problem at all with CC0. I can also commit again a brand new french translated strings.xml, with the confirmation that there is no other way to translate the sentences, which can be found in both the old and new strings file. That way you'll blindly can license it under CC0. Just say it and I'll do it :)

gsantner:
Okay great, thanks!

Othe: A little difference in non-small strings always happens :D

gsantner:
German is now completly new translated by myself. Two other languages are avaialble too now.

Issue #1, opened 2017-08-26
Android >= M permissions


gsantner:
The app currently doesn't support the M permission system, therefor saving/reading from external storage does not work. Theres too no crash or dialog to confirm right now. The project was at API 22 at fork time, so it wasn't possible at that time. Currently one has to accept the permission manually in the system app settings

PR #17, opened 2017-09-19
Fix syntax highlighting #1 

sjmduncan:
* Adds plus and hyphen list markers
* Requires a space between list marker and list item text before applying highlighting
* Limits pound-sign header notation to 6 levels
* Require three or more hyphens or equals signs for 'udnerlined' header notation and prevent highlight conflicts with un-ordered lists using hyphens as list markers.

Auto-insert list markers when creating a newline at the end of an existing list item still only works some of the time, however when it does work it also works for plus and hyphen markers.


gsantner:
thanks!

You can add yourself to the contributor list

gsantner:
@sjmduncan If you are interested, maybe take a look into issue no21: `https://github.com/gsantner/markor/issues/21`

I feel like the current/old implementation of NoteActivity is broken by design. No question if something already exists, everything based on current UI text, lots of useless stuff... I think its best to rewrite it completly

Issue #11, opened 2017-09-17
Inconsistent behaviour & syntax highlighting when creating lists


sjmduncan:
#### General information

* **Device:**  Moto G 4G
* **Android Version:**  Android 7.1 (LineageOS 14.1)
* **App version:** v0.1.1

#### Steps to reproduce

1. Create a note
2. Add a list item (* + - or 1. ) 
3. Press enter at the end of the line

#### Expected result

1. Syntax-highlight the list marker
2. When creating a newline at the end of a list item, the next line should automatically start with the appropriate list marker, preserving indentation.

Syntax highlihghting doesn't work for the + or - markers. For the - marker every second item is highlighted with the header style. Both of those things work as expected in the markdown preview. 

Auto-inserting list markers only works for the * marker, only when editing existing notes, and only after you press enter at the end of a list item that already existed in that note when you loaded it.

gsantner:
The markdown highlighting is not modified yet since forking. I am thinking of maybe swapping it out completly instead of fixing the broken implementation. If you find something that works good and can be added (open source lib) I am open to include and replace the current implemetnation with it.

You can send hints for a good replacement, or a PR to improve the highlighting, but I would prefer to use some maintained markdown-highlighter instead of some outdated custom soluton.

sjmduncan:
There's [RxMarkdown](https://github.com/yydcdut/RxMarkdown) but I have no idea if it's any  good.

gsantner:
Seems nice, but very heavy / pulls in too much dependencies

sjmduncan:
For the moment I'll tidy up the existing one because I haven't found a decent replacement. It's not really *that* bad, mostly just some incomplete regexes.

It would be nice to have a renderer which is consistent with the syntax highlighter though. The current preview renderer has some issues of its own - no strikethrough, H1 with '===' but no H2 with '---', and sometimes weird list item spacing sometimes to name a few.

gsantner:
If you want and can, you can also take a look into my SimpleMarkdownParser. That works really good in many apps yet (for displaying). I dont know how much would need to be done to make it highlighting too

gsantner:
It's `net.gsantner.opoc.util.SimpleMarkdownParser`. If that can be expanded to highlighting too, I think it could be even more great. Basicially the infrastructure is there, it would just ned a custom "filter" (as you can see theres e.g. one for CHANGELOG especially)

sjmduncan:
I already have a PR to fix the most obvious problems.

[...]

PR #23, opened 2017-09-22
Highlighter matches renderer

sjmduncan:
Syntax highlighter now matches the render for most cases (#11)

There are some inconsistencies but addressing those with the current highlighter would be difficult to get right.

Auto-list-item for lists also only works for unordered lists for that reason.


gsantner:
thanks,
merging on desktop then

Issue #28, opened 2017-09-25
Feature request: sharing text as a file rather than as content 


imrosca:
#### General information

* **Device: Xperia Z1 compact**
* **Android Version:** 5.1.1
* **App version:** v0.1.2

Sharing text to seadroid is broken because seadroid doesn't seem to know how to handle text content, so I'm not sure if this is a markor  bug that the intent for sharing includes seafile, or if seadroid is registering for the wrong intent. 

Regardless of whose responsibility it is for a graceful failure, sharing markdown from markor as a file with the known markor filename (e.g. name.md) is what I actually want to do. 

#### Feature example

   Given seadroid is installed
   When I share a note from the rendered view
   Then I want an option to send a named file rather than the content of the file for the markdown and html formats
   When I select a file format
   And I choose seafile as the target 
   Then I should see the menu to choose a library in which to save the file

PR #34, opened 2017-09-29
Add share option from preview activity for sharing file instead of text

imrosca:
I attempted to fix issue #28 I raised a few days ago. I've not done any android development previously, but this small change seems to work for me.

Further work could include sharing multiple files after selection from the browser activity.


gsantner:
Use share as file instead of share as markdown. else it seems all good

imrosca:
Updated as a new commit.

gsantner:
🎉

PR #50, opened 2017-10-02
Fixed #32

### LINKED ISSUE #32: Back navigation when viewing a file within a folder returns to the overall screen rather than the folder
#### General information

* **Device:**  Oneplus 3T
* **Android Version:**  7.1.2
* **App version:**  v0.1.3

#### Steps to reproduce

1. Ensure the option 'Remember last folder location on startup' is **unchecked**
2. View file (either view or edit mode) within a folder
3. Press back button

Additional setting change from default: 'Preview first when opening notes' is **checked**

#### Expected result
Return to the file listing within the folder the file was in

**What do you see instead?**
The contents of storage folder selected in options

**Additional Info/Comments**  
It appears that the option to 'Remember last folder on startup' being unchecked additionally causes Markor to not remember the last folder when exiting the view/edit mode.

beast-incarnate:
Fixed by keeping a track of Last Opened Directory even if the user has deselected the option and checking for the same at the start of the MainActivity.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/MainActivity.java
+++ b/app/src/main/java/net/gsantner/markor/activity/MainActivity.java
@@ -71,8 +71,6 @@ public class MainActivity extends AppCompatActivity {
 
     private boolean _doubleBackToExitPressedOnce;
 
-
-
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
--- a/app/src/main/java/net/gsantner/markor/activity/MainActivity.java
+++ b/app/src/main/java/net/gsantner/markor/activity/MainActivity.java
@@ -71,12 +71,17 @@ public class MainActivity extends AppCompatActivity {
 
     private boolean _doubleBackToExitPressedOnce;
 
+
+
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         ContextUtils.get().setAppLanguage(AppSettings.get().getLanguage());
         if (AppSettings.get().isOverviewStatusBarHidden()) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/net/gsantner/markor/activity/MainActivity.java
```diff
@@ -71,13 +71,18 @@
 
     private boolean _doubleBackToExitPressedOnce;
 
+
```
gsantner: can be removed

app/src/main/java/net/gsantner/markor/activity/NoteActivity.java
```diff
@@ -74,6 +74,7 @@
     private String _initialContent = "";
     private String _initialFileName = "";
 
+
```
gsantner: no neee

app/src/main/java/net/gsantner/markor/activity/PreviewActivity.java
```diff
@@ -57,6 +57,7 @@
     private boolean _isEditIncoming = false;
     private MarkDownRenderer _renderer = new MarkDownRenderer();
 
+
```
gsantner: no need

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
beast-incarnate: Should I make another PR with the necessary changes?

gsantner: no, just reset those lines and make commit :)

gsantner:
thanks! merging on PC then

gsantner:
Thanks! here are some more if you are interested: <https://github.com/search?&q=label%3Ahacktoberfest+user%3Agsantner>

PR #54, opened 2017-10-04
Fixed #24

### LINKED ISSUE #24: Sort notes
#### General information
* **App version:** 0.1.2

#### Expected result
Feedback from gplay

beast-incarnate:
Added sorting functionality.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/NoteActivity.java
+++ b/app/src/main/java/net/gsantner/markor/activity/NoteActivity.java
@@ -42,7 +42,6 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStreamWriter;
-import java.lang.reflect.Type;
 import java.util.UUID;
 
 import butterknife.BindView;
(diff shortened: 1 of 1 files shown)


gsantner:
It seems like you didn't rebase on current master. Can you do a pull on master, merge/rebase on that in your branch and push that? (do a safety copy first, just to have an backup if something goes wrong)

app/src/main/java/net/gsantner/markor/activity/FilesystemListFragment.java
```diff
@@ -262,6 +256,93 @@ public void clearSearchFilter() {
         reloadAdapter();
     }
 
+    public void sortByName(){
+        int size = _filesCurrentlyShown.size();
```
gsantner: You don't need to do this all by hand! Nice work, but you don't need to that all manually, wow :D. There is already a lot of stuff in java standard lib that does this hard work for you. Just use  for example

```
Collections.sort(data, new Comparator(){

compareto(){
 return something.compareto(somethinelse)
}
```

beast-incarnate: Thanks for the info. I will make the necessary changes.

PR #55, opened 2017-10-04
Fixed #24

### LINKED ISSUE #24: Sort notes

beast-incarnate:
Added Sorting Functionality.


gsantner:
I looked into your solution, but it seems to only sort the data once when clicked. The target is that the sorting type is saved and done everytime something is browsed. So if I open folder "notes" the files are again sorted by what was chosen.

You could save e.g. sortType to a variable: Filled with something out of `SORT_TYPE_NAME`, `SORT_TYPE_DATE`, `SORT_TYPE_FILESIZE` and then call the proper method when opening a folder

I can merge this a first step to the solution, but it is not finished.

beast-incarnate:
If u want please merge this PR and then I can submit another PR with the sortType.

Issue #20, opened 2017-09-20
Markor translation & localization


gsantner:
[![Translate - with Stringlate](https://img.shields.io/badge/stringlate-translate-green.svg)](https://lonamiwebs.github.io/stringlate/translate?git=https%3A%2F%2Fgithub.com%2Fgsantner%2Fmarkor.git&name=Markor&web=https%3A%2F%2Fgithub.com%2Fgsantner%2FMarkor) [![Chat - Matrix](https://img.shields.io/badge/chat-on%20matrix-blue.svg)](https://matrix.to/#/#markor:matrix.org) [![Chat - FreeNode IRC](https://img.shields.io/badge/chat-on%20irc-blue.svg)](https://kiwiirc.com/client/irc.freenode.net/?nick=markor-anon|?#markor)

This issue will be used to notify translators of the app. Please subscribe to this issue if you want to start translating or already translated something and want to keep the strings updated for some language.

~~You can use e.g. [Stringlate](https://f-droid.org/packages/io.github.lonamiwebs.stringlate/) to contribute translations - directly on your phone.~~ This app is now translated on Crowdin! Join the project [here](https://crowdin.com/project/markor/invite) (free of course) and start translating!

List of translators that are on GitHub and contributed at least once translations to one of my projects _(unified across gsantner projects)_:
* French - @milouse, @Poussinou, @Quent-in, @SansPseudoFix, @gbaudic, @romainrossi, @NicolasAzzopardi, @nailyk-fr, @rain79, @Primokorn, @Bridouz
* Polish - @kRkk, @pskosinski, @Bjornskjald, @bla-ke, @thesebas 
* Spanish - @larjona, @Lonami, @AlexandreInsua, @sidhenidon
* Italian - @simocosimo, @Daraf-IT, @matteoarrighi
* Russian - @xredor, @guland2000, @trubitsyn, @antonv6
* Brazilian - @luisfsr, @TiagoDanin, @mannuscritto, @RockyTV, @smarquespt, @Valeyard1
* Ukrainian - @cl0ne
* Swedish - @spitlo
* German - @gsantner
* Turkish - @mobile46
* Hindi - @kartikarora
* Czech - @vakabus, @Golbinex 
* Japanese - @naofum
* Sardinian - @asereze
* Greek - @nicolasmanic
* Dutch - @TopdRob
* Indonesian - @freedomofkeima
* Catalan - @Zagur
* Norwegian - @comradekingu
* Hungarian - @maxigaz
[...]

PR #56, opened 2017-10-04
Hindi (India) translations of strings

kartikarora:
For issue #20

### DIFF
--- a/app/src/main/res/values-hi/strings.xml
+++ b/app/src/main/res/values-hi/strings.xml
@@ -1,0 +1,80 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="action_settings">सेटिंग्स</string>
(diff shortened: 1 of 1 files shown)


kartikarora:
@gsantner You might wanna have a look at [plurals](https://developer.android.com/guide/topics/resources/string-resource.html#Plurals) to manage quantity stirngs, such as on line 60 and 61 of `strings.xml`

gsantner:
Another leftover from the old codebase. You can send a PR for that if you want to

kartikarora:
I can, but then it'll create an issue for other language strings. Its better if you use that concept for future quantity strings and other languages adapt.

gsantner:
Would just add the new plural now, but not yet use it in code, and swap over to it at some point when more is translated

kartikarora:
Ah good idea! PR coming in 5 mins!

PR #57, opened 2017-10-04
You forgot about add ')' in changelog.md

kRkk:


gsantner:
I will add that brace when doing next release, thanks for noting :)

PR #61, opened 2017-10-05
Fixed #24 part 2

### LINKED ISSUE #24: Sort notes

beast-incarnate:
Added sortType variable to save the sortType selected.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/FilesystemListFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/FilesystemListFragment.java
@@ -62,7 +62,7 @@ public class FilesystemListFragment extends Fragment {
 
     private NotesAdapter _filesAdapter;
 
-    private int sortType = 0;
+    private int _sortType = 0;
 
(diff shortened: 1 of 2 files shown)


gsantner:
Thanks! Maybe go the way of combining to one sortAdapter() method accepting the sorttype as parameter. currently the first \~10 lines are duplicate in all.

example:
```
void sortAdapter(int sortType) {
         int size = _filesCurrentlyShown.size();
         int k=0;
         for(int i=0;i<size;i++){
             if(_filesCurrentlyShown.get(i).isDirectory()){
                 k++;
             }
         }
switch(sortType){
case 0:
Collections.sort(
break;
case 1:
...
}

reload
_sortType = sortType;
}
```

gsantner:
exactly ;)

btw: I noticed that sort by date and size is not working. Sort by Date shows e.g. a note from september than october than september again. Sort by size does nothing at all

PR #67, opened 2017-10-05
[WIP] Added allow to share Text and link 

nextybrown:
Text share working fine
and some design good practice

### DIFF
--- a/app/src/main/res/menu/main.xml
+++ b/app/src/main/res/menu/main.xml
@@ -11,6 +11,12 @@
         app:actionViewClass="android.support.v7.widget.SearchView"
         app:showAsAction="always" />
     <item
+        android:id="@+id/action_folder"
+        android:title="@string/add_new_folder"
+        app:showAsAction="always"
+        android:icon="@drawable/ic_create_new_folder_white_24dp"
+        />
+    <item
         android:id="@+id/action_import"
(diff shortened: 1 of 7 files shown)


app/src/main/res/menu/main.xml
```diff
@@ -11,6 +11,12 @@
         app:actionViewClass="android.support.v7.widget.SearchView"
         app:showAsAction="always" />
     <item
+        android:id="@+id/action_folder"
```
gsantner: This can totally be added, but I would like to ask you to make that optionally toggleabel, so its still showing the 2 fabs by default, and switching to the toolbar option way when enabled.

In general fab is just for primary action right, but got used to it in this app and quite like it in this case.

gsantner:
Thanks for PR!
This only allows to "create new" markor files, append to existing is not done yet, or?

app/src/main/res/menu/main.xml
nextybrown: ok cool , don't think putting int setting will be better , let just leave like that

gsantner: k

gsantner:
Are you still working on this?

gsantner:
Imported current state / part 1 for releasing new update

PR #68, opened 2017-10-06
Implemented Markdown Actions

### LINKED ISSUE #42: Replace markdownbar with markdown actions instead of just characters
E.g. a icon showing bold text -> makes selected text, or current word bold / not bold depending on current state.

Example UI from OctoDroid below

30575545-de5735e4-9d01-11e7-852a-e5dda1950155

KrishAmal:
fixes #42


gsantner:
nice, Loki ng into it after university

KrishAmal:
Yeah sure . I have used regular expressions to match the selection to check if any action has been applied( i.e to undo) , however i have only added alphanumeric and whitespace characters to the part of regular expression which matches the actual text .

KrishAmal:
And , the settings item for smart bracket can be removed i guess .

gsantner:
i think you can remove those smart/regular actions and just have one list of markdown actions

gsantner:
thought the same

app/src/main/java/net/gsantner/markor/activity/NoteActivity.java
```diff
@@ -206,47 +207,39 @@ protected void onPause() {
...
-
     }
 
-    private void appendSmartBracketShortcuts() {
-        for (String shortcut : Constants.KEYBOARD_SMART_SHORTCUTS) {
-            appendButton(shortcut, new KeyboardBarSmartShortCutListener());
-        }
+    private void appendExtraActions(){
+        for (int[] actions : Constants.KEYBOARD_EXTRA_ACTIONS_ICONS) {
+                appendButton(actions[0], new KeyboardRegularActionListener(Constants.KEYBOARD_EXTRA_ACTIONS[actions[1]]));
```
gsantner: doesnt this append always the same?

KrishAmal: yeah , i made a third category for actions which doesn't have a consistent syntax , such as `()[]` and `----` . Thought if in the future , actions for tables or images were brought , a third cateogry would be apt for such syntax's . Right now it uses the same keyboard listener of the Regular Action.

gsantner:
For images and links we could go the way of showing a dialog, e.g. with title and link, and then insert the result as markdown

KrishAmal:
Yeah , that sounds good . I'll start working on that .

app/src/main/java/net/gsantner/markor/activity/NoteActivity.java
gsantner: Yeah, even if they use the same, wouldn't it be better to append the current iterated action to ensure not making bugs later?
e.g.

```
for (int i=0; i < actions.length ; i+=2) {
append(action[i] , .... action[i+1];
}
```

KrishAmal: I didnt understand . All the actions are appended exactly once to the keyboard bar , isn't that the desired outcome ? . The constants KEYBOARD_EXTRA_ACTIONS_ICONS is of type `int[][]` , used to store a drawable and an int (which referes to its respective posiiton of character to be appended).

KrishAmal:
device-2017-10-07-120508

gsantner:
nice! you can even use the same dialog for both, only needed to change dialog title/hints

app/src/main/java/net/gsantner/markor/activity/NoteActivity.java
gsantner: I meant that you are iterating over actions[] but yoi always Insert actions[0] and action[1]

gsantner: Oh, its a multi array, right, my bad :D

KrishAmal: Yeah :-)

KrishAmal:
Yes , I'll push the updates soon

gsantner:
Thanks! Im merging this now, so its not getting too big and to not make a lot of merge conflicts with other PRs then :D

I may make some adjustments, so pull them in if you do/want to continue :)

KrishAmal:
Yeah cool :-D

PR #70, opened 2017-10-07
Update uk (Ukrainian) translation

cl0ne:


gsantner:
@Abrasha If you want you can also take a look on that

app/src/main/res/values-uk/strings.xml
```diff
@@ -5,10 +5,10 @@
     <string name="dark">Темна</string>
     <string name="delete">Видалити</string>
     <string name="rename">Перейменувати</string>
-    <string name="confirm_delete">Підтвердити видалення</string>
+    <string name="confirm_delete">Підтвердіть видалення</string>
```
Abrasha: I am not sure that this makes sense. If it is button title than it should be reverted. Because previous means action and new one - request. @gsantner When this one is used?

app/src/main/res/values-uk/strings.xml
```diff
@@ -19,59 +19,70 @@
     <string name="select_folder">Обрати директорію</string>
 
     <string name="search_hint">Пошук нотаток</string>
-    <string name="share_string">Поділитися з...</string>
-    <string name="share">Поділитися</string>
+    <string name="share_string">Надіслати до...</string>
+    <string name="share">Надіслати</string>
```
Abrasha: New one means 'Send'. As for me it is better to keep previous one.

app/src/main/res/values-uk/strings.xml
```diff
@@ -19,59 +19,70 @@
...
-    <string name="error_cannot_create_save_dir">Помилка: неможливо зберегти директорію</string>
+    <string name="pref_title__render_rtl">Відображати для писемності справа-наліво</string>
+    <string name="error_cannot_create_save_dir">Помилка: неможливо створити директорію</string>
 
     <!-- Font Size -->
 
     <!-- Hard coded string fix -->
     <string name="last_modified">Дата зміни: %s</string>
-    <string name="number_of_files">Кількість файлів: %d</string>
+    <string name="number_of_files">Число файлів: %d</string>
```
Abrasha: Why do you think new one is better? Old one seems to me more native.

Abrasha:
@gsantner Nice improvements provided by @cl0ne. I have only 3 remarks.

app/src/main/res/values-uk/strings.xml
gsantner: When you select an item, click on the delete button. then you asked if you REALLY want to delete that file(s)

Abrasha: It is appeared on BUTTON or dialog title / content?

cl0ne: it's in a dialog's title https://github.com/gsantner/markor/blob/d39f80116cad7daf209087a6f033f2494d0a256d/app/src/main/java/net/gsantner/markor/activity/FilesystemListFragment.java#L146

Abrasha: @cl0ne super. If so you, I am sorry.

gsantner:
Ping me when ready

app/src/main/res/values-uk/strings.xml
cl0ne: I think translation here is the matter of preference and 'Share' have different translations in different apps (Amaze, NextCloud and others).
I preferred to use 'Send' (to app/recipient) since I consider 'Share' more as 'social network oriented' action.

app/src/main/res/values-uk/strings.xml
cl0ne: Agree.

app/src/main/res/values-uk/strings.xml
cl0ne: @Abrasha np

app/src/main/res/values-uk/strings.xml
Abrasha: @cl0ne Makes sense.

Abrasha:
@cl0ne @gsantner LGTM

cl0ne:
@Abrasha thanks for review

PR #72, opened 2017-10-07
[WIP] Added in the skeleton for adding a notification. Partial Fix #44

### LINKED ISSUE #44: Reminder feature
Allow to set a reminder for a specific document. This would create e.g. a notification at 6PM if setted up. A click on the notification will open the document. 

* step 1: option for onetime
* step 2: options for  intervals (like every day at 5PM, every monday) .

commanderpepper:
I added a basic UI for adding notifications. I will try to get one time notifications in tomorrow.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/NotificationDialogFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/NotificationDialogFragment.java
@@ -1,0 +1,41 @@
+package net.gsantner.markor.activity;
+
+import android.app.Dialog;
+import android.app.DialogFragment;
+import android.app.TimePickerDialog;
+import android.os.Bundle;
+import android.text.format.DateFormat;
+import android.widget.TimePicker;
+import android.widget.Toast;
+
+import java.util.Calendar;
+
+/**
+ * Created by Humza on 10/7/2017.
+ */
+
+public class NotificationDialogFragment extends DialogFragment implements TimePickerDialog.OnTimeSetListener {
+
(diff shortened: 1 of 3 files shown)


commanderpepper:
This is for issue #44 
This is the first part of the code to add notifications. 
I need to make a class to handle the creation of notifications.

I do have some questions. 
What is the small icon that you would like to use for the notifications. I'm thinking of using the ic_launcher. 
Also can you elaborate on this line "Only for notes that reside in the app's Notebook (getSavePath())."

gsantner:
icon: can be launcher, yes

path: Allow reminder ONLY for files that reside in the app's save directory (that one returned by said method). So not allow to set reminders, when e.g. opening some file opened e.g. from filemanager

app/src/main/java/net/gsantner/markor/activity/NotificationDialogFragment.java
```diff
@@ -0,0 +1,41 @@
...
+import android.text.format.DateFormat;
+import android.widget.TimePicker;
+import android.widget.Toast;
+
+import java.util.Calendar;
+
+/**
+ * Created by Humza on 10/7/2017.
+ */
+
```
gsantner: I didnt look to much into code, but be sure to use AppCompat and NotificationCompat

gsantner:
do you keep working on it?

commanderpepper:
I didn't have time to work on the problem. I'll come back to this issue on Friday.

PR #77, opened 2017-10-15
[WIP] Extension Bug Fix

### LINKED ISSUE #37: Don't change file extension when saving
Currently file gets moved to *title.md* when it was called title.markdown. Resolving this will keep the original extension.

manmeet-22:
#37
Now supports multiple extension.
Rest you can add the extensions you want in the Constants.java file and then just write them in the _extensions array in NoteActivity.


app/src/main/java/net/gsantner/markor/model/Constants.java
```diff
@@ -18,8 +18,18 @@
...
     public static final String UTF_CHARSET = "utf-8";
 
-    public static final String MD_EXT = ".md";
-
+    public static final String MD_EXT1 = ".md";
+    public static final String MD_EXT2 = ".markdown";
+    public static final String MD_EXT3 = ".mkd";
+    public static final String MD_EXT4 = ".mdown";
+    public static final String MD_EXT5 = ".mkdn";
+    public static final String MD_EXT6 = ".pdf";
```
gsantner: markdown cannot be saved as picture or pdf. only be exported, so these should not be listed

app/src/main/java/net/gsantner/markor/activity/NoteActivity.java
```diff
@@ -126,6 +127,10 @@ public void onFocusChange(View v, boolean hasFocus) {
                 }
             }
         });
+
+        _extensions = new String[]{ Constants.MD_EXT1, Constants.MD_EXT2, Constants.MD_EXT3,Constants.MD_EXT4,
```
gsantner: could be pur in constants too, for reuse

app/src/main/java/net/gsantner/markor/model/Constants.java
manmeet-22: Oh that just  got out of my mind.

app/src/main/java/net/gsantner/markor/activity/NoteActivity.java
manmeet-22: Didn't get you. "Could you pur"? 
Rest you can tell me which all extensions you want and I will add them  or you can add it on your own?

gsantner: put

everything markdown like + txt

manmeet-22:
Hmm still not getting you. Please Elaborate. 
Is the way I am implementing ok with you? 
Rest for the extensions part you can add any extensions you want.

app/src/main/java/net/gsantner/markor/activity/NoteActivity.java
gsantner: to move the array to Constants and make it final static

manmeet-22:
Is it ok now?

vaskas:
I don't think it's the right solution to the problem.
I think what should happen is:
- When the file is new or there is no extension, append `.md`
- When the file has got *any* extension at all, leave it alone

This would make the editor suitable for any generic (text) files. Thoughts?

gsantner:
thats what it should do, yes

PR #80, opened 2017-10-18
Fixed share as image menu

### LINKED ISSUE #47: Fix share document as image
Currently exporting a markdown document as image just exports the visible part. Resolving this should export the whole document as image. Also the image currently contains a lot of whitespace on bottom (see image), which should be gone too. 

<table><tr><td></td></tr></table>

jvretamero:
Fix #47

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/PreviewActivity.java
+++ b/app/src/main/java/net/gsantner/markor/activity/PreviewActivity.java
@@ -11,6 +11,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.graphics.Bitmap;
 import android.graphics.Canvas;
+import android.graphics.Paint;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
@@ -24,6 +25,7 @@ import android.support.v7.widget.Toolbar;
 import android.text.TextUtils;
 import android.view.Menu;
 import android.view.MenuItem;
+import android.view.View;
 import android.view.WindowManager;
 import android.webkit.WebView;
 import android.webkit.WebViewClient;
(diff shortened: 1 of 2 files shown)


app/src/main/java/net/gsantner/markor/activity/PreviewActivity.java
```diff
@@ -60,6 +61,11 @@
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            WebView.enableSlowWholeDocumentDraw();
```
gsantner: can this be toggled before picture creation too?

jvretamero: If I understood your question correctly, there is no need to toogle this configuration, because the [documentation](https://developer.android.com/reference/android/webkit/WebView.html#enableSlowWholeDocumentDraw()) says _"This method should be called before any WebViews are created."_ and there is no method like _"disableSlowWholeDocumentDraw()"_.

gsantner: does it make a big speed impact when using this in markor?

app/src/main/res/layout/preview__activity.xml
```diff
@@ -1,16 +1,21 @@
...
+    android:orientation="vertical">
 
     <include
         android:id="@+id/toolbar"
-        layout="@layout/toolbar"/>
+        layout="@layout/toolbar" />
 
-    <WebView
-        android:id="@+id/preview__activity__webview"
+    <ScrollView
```
gsantner: Does scrolling work on older devices (minApi) too? Horizontal and vertical? <https://stackoverflow.com/questions/9718245/webview-in-scrollview>

app/src/main/java/net/gsantner/markor/activity/PreviewActivity.java
jvretamero: No

app/src/main/res/layout/preview__activity.xml
jvretamero: I just did other tests varying the implementation and also the devices, so:

1. It works with the current minApi (16)
2. As you can see, I did two commits, the ScrollView was needed to make the first implementation work. I have tested the pull request's current implementation and the ScrollView is not needed anymore

Do I create another commit without the ScrollView?

gsantner: yep, that may create bugs in some cases

jvretamero: Done

gsantner:
🎉 thanks

PR #116, opened 2017-11-09
Swedish translations

spitlo:
Nice work with the updates!

There were some strings I couldn't find in the app, but they're pretty general so I think it'll be ok.


gsantner:
Thanks!
Yep, there were  some general strings added, but not yet used

PR #121, opened 2017-11-09
Changed back to QuickNote

spitlo:


gsantner:
is there also something shorter for notebook? or does it fit?

spitlo:
It fits on my Nexus 4, and it has a pretty small screen. But perhaps the symmetry is a bit off. I tried to think of something shorter but couldn't come up with something conceptually close to "notebook" in fewer letters.

PR #123, opened 2017-11-09
Syntax highlighting for line breaks

sjmduncan:
Add a highlighting filter which changes the background to grey for two or more trailing whitespace characters at the end of a non-empty line of text.


gsantner:
wow, thx for fast response!

If you want to have some better conversatsion method than here for development, theres IRC, Matrix channel for markor and my personal accounts too (<http://gsantner.net/#contact>)

Issue #130, opened 2017-11-18
Insert Link/Image select/replaces whole line


gsantner:
#### General information

* **App version:**  v0.2.2
* **Android Version:**  7.1

#### Expected result
* [ ]  **(broken)** Add new link/image if selection/marker is in non-markdown-link element
* [ ] **(broken)** Modify link/image if selection/marker is in markdown-link element
* [x] Add new link/image if multiple characters are selected (selection comes after pressing long / holding shift on keyboard)

#### Steps to reproduce
The whole line gets selected when theres a single link somewhere in the line, or lines above/below if these are empty lines. 

See video: <https://github.com/gsantner/gh_tmp/blob/master/video_2017-11-18_14-00-29.mp4> (Note: Playback on Firefox seems to be broken, still picture)

gsantner:
@jamiet-dev  @sjmduncan

Could you/Do you want to take on this? Thanks 😄

sjmduncan:
Oops, that was some pretty lazy testing for #114 & issue #100.

#133 should fix this.

PR #133, opened 2017-11-20
Fix for #130

sjmduncan:
Don't replace the whole line when editing existing links (and don't try to edit links when the cursor is on the same line but not actually *in* the link)


gsantner:
thx for fast response!

PR #164, opened 2017-12-17
New swedish translations

spitlo:
I couldn't find where `search_hint__add_or_custom` was used, so I couldn't figure out a good translation for "Custom". Will add it the next time if I figure it out,


gsantner:
In the todo bottombar - at context and project (b2,3)

gsantner:
thx! merging

PR #176, opened 2018-01-05
Add Simplified Chinese Translation.

billchenchina:


gsantner:
thx!

i think it should be `values-zh-rCN`, read here too <https://stackoverflow.com/a/4249345>

billchenchina:
Done!
@gsantner

PR #179, opened 2018-01-12
Updated Polish translation

thesebas:


gsantner:
I may add some new strings when I have time to develop; I will add these after translation then; Leaving this open till

PR #225, opened 2018-04-25
Norwegian Bokmål translation

comradekingu:


gsantner:
Do all/most norwegians understand that dialect? Else would put it in "values-no" so every norwegian benefits

Error: `values-nb-rNO/strings.xml:186:50: Error: Element type "string" must be followed by either attribute specifications, ">" or "/>".`

gsantner:
As I do understand from <https://de.wikipedia.org/wiki/Bokm%C3%A5l> , by 85% bokmal is the official administrative language in norway, and ~90% write it in this language.

In this case I think it's best to set it as norwegian (values-no). As I understood many things are equal in the other dialects and some things may need to be overridden, which can be done then e.g. in ﻿Nynorsk translation.

comradekingu:
@gsantner The way to go is nb_NO, the other is nn_NO. You can set up fallback locales if you want something other than English for strings that arent there.

gsantner:
What would be nb alone?

comradekingu:
@gsantner New version of nb_NO. It is not as backwards-compatible.

PR #226, opened 2018-04-25
Spelling: Language rework

comradekingu:
Questions: 

"list" and "archive" verb or noun?
"context"as in?
"alignment" of?
"Page Up (Page Up)
Page Down (Page Down)" Why is the name shown twice, and should it be PgDn?


app/src/main/res/values/strings.xml
```diff
@@ -135,26 +135,26 @@
...
+    <string name="global">Program-wide</string>
     <string name="main_view">Main View</string>
     <string name="theme_description">The overall theme used in the app</string>
     <string name="load_last_directory_at_startup">Remember last directory</string>
     <string name="syntax_highlighting">Syntax highlighting</string>
     <string name="text_modules">Text modules</string>
     <string name="text_module">Text module</string>
     <string name="highlighting_delay">Highlighting delay</string>
-    <string name="highlighting_delay_summary">Delay in milliseconds for highlighting refresh. A lower value will start highlighting earlier, but will drain more battery and decrease performance.</string>
+    <string name="highlighting_delay_summary">Delay in milliseconds for highlighting refresh. A lower value will start highlighting earlier, but will drain more battery and improve performance.</string>
```
comradekingu: I don't see why it wouldn't improve performance, but I could be wrong.

app/src/main/res/values/strings.xml
```diff
@@ -135,26 +135,26 @@
     <string name="number_of_files" needed="false">Number of files: %d</string>
 
     <!-- New strings -->
-    <string name="global">Global</string>
+    <string name="global">Program-wide</string>
```
comradekingu: What does "global" mean here? It is very easy to get wrong in translation by assuming its meaning.

app/src/main/res/values/strings.xml
```diff
@@ -163,8 +163,8 @@
     <string name="right_to_left_rendering__abbreviated">RTL rendering</string>
     <string name="highlight_line_ending">Highlight line ending</string>
     <string name="highlight_line_ending_if_two_or_more_spaces">Highlight spaces at the end of lines containing two or more spaces</string>
-    <string name="use_monospace_for_code">Use Monospace for code</string>
-    <string name="use_different_fonttype_slow_down">Make use of a different font type. Changing the font type dynamically makes an impact on the performance of the app.</string>
+    <string name="use_monospace_for_code">Use monospace font for code</string>
```
comradekingu: Assuming here that the font isn't called "Monospace"

app/src/main/res/values/strings.xml
```diff
@@ -177,13 +177,13 @@
...
     <string name="start_on_bottom">Start on bottom</string>
-    <string name="start_on_bottom_summary">After a document gets loaded, position the cursor on the end</string>
-    <string name="filemanager_doesnot_supply_required_data">Markor only allows to edit local, offline available, documents and support file managers which supply the full filepath including filename. Apparently the app didn\'t supply all required information and can\'t be used with Markor :(.</string>
-    <string name="sync_to_local_folder_notice">Notice: If you use a sync app, the files may reside in the known or custom folders, you can open the files by changing Markor\'s document folder to it, or by opening the file via a different file manager.</string>
-    <string name="configure_in_the_apps_settings">Configure in the apps settings</string>
+    <string name="start_on_bottom_summary">Upon loading a document, position cursor at its end</string>
+    <string name="filemanager_doesnot_supply_required_data">Markor can only edit local, offline, documents and file managers supplying full filepath, including filename. The app didn\'t supply all required info and can\'t be used with Markor :(.</string>
```
comradekingu: I have no idea what this sentence means, and my attempt at improvement is likely wrong.

app/src/main/res/values/strings.xml
```diff
@@ -177,13 +177,13 @@
...
+    <string name="sync_to_local_folder_notice">Notice: If you use a sync app, files may reside in custom folders, open them by changing Markor\'s document folder accordingly, or by opening the file in another file manager.</string>
```
comradekingu: This is likely problematic, as my understanding of the source string is lacking.

app/src/main/res/values/strings.xml
gsantner: doing hard work more often means more usage, means decreased. it's correct as is

app/src/main/res/values/strings.xml
gsantner: correct

app/src/main/res/values/strings.xml
gsantner: where do you have this string out of? it is currently not more in the list to translate. do you have an old git version?

app/src/main/res/values/strings.xml
gsantner: Would leave it as previous

app/src/main/res/values/strings.xml
gsantner: then just leave it :)

app/src/main/res/values/strings.xml
gsantner: App wide

app/src/main/res/values/strings.xml
```diff
@@ -199,22 +199,22 @@
     <string name="center_text_desc">Edit text in center of the screen</string>
     <string name="clipboard">Clipboard</string>
     <string name="existing_document">Select existing document</string>
-    <string name="open_source_licenses">Open Source Licenses</string>
+    <string name="open_source_licenses">Libre Software Licenses</string>
```
gsantner: Open Source Licenes is just fine, and I don't want to start a discussion on oss/foss/floss/free-sw :D.

gsantner:
> "list" and "archive" "context"

nouns. see <https://github.com/todotxt/todo.txt>, don't know how/if to translate it, it's how it's called in the spec and clients call it.

> "alignment" of?

text, cannot think of anything else


> "Page Up (Page Up)
> Page Down (Page Down)" Why is the name shown twice, and should it be PgDn?

Key names. really depends on the keyboard

app/src/main/res/values/strings.xml
comradekingu: That would be efficiency. Performance is related to outcome, not effort. Like the performance governor for CPU scaling on Linux.

app/src/main/res/values/strings.xml
comradekingu: Could you explain the "filemanager_doesnot_supply_required_data" string?

app/src/main/res/values/strings.xml
comradekingu: Could you explain the "filemanager_doesnot_supply_required_data" string?

app/src/main/res/values/strings.xml
comradekingu: Could you explain the "sync_to_local_folder_notice" string?

gsantner: Markor can only edit with full filepath available. the given doesnt include everything or anything at all. I think it's already so that everybody understands that it cant be used.

gsantner:
Thanks @comradekingu ! merging

PR #240, opened 2018-05-05
Feature file info dialog

subkanthi:

### DIFF
--- a/app/src/main/java/net/gsantner/markor/ui/FileInfoDialog.java
+++ b/app/src/main/java/net/gsantner/markor/ui/FileInfoDialog.java
@@ -1,15 +1,18 @@
 package net.gsantner.markor.ui;
 
 import android.app.Dialog;
+import android.content.Context;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.v4.app.DialogFragment;
 import android.support.v4.content.ContextCompat;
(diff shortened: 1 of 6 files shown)


app/src/main/java/net/gsantner/markor/ui/FileInfoDialog.java
```diff
@@ -0,0 +1,90 @@
...
+        boolean darkTheme = AppSettings.get().isDarkThemeEnabled();
+        dialogBuilder = new AlertDialog.Builder(getActivity(), darkTheme ?
+                R.style.Theme_AppCompat_Dialog : R.style.Theme_AppCompat_Light_Dialog);
+        root = inflater.inflate(R.layout.ui__file_info_dialog, null);
+
+        dialogBuilder.setTitle(getResources().getString(R.string.info));
+        dialogBuilder.setView(root);
+
+        TextView textView = root.findViewById(R.id.ui__filesystem_item__description);
+        textView.setText(file.getName());
```
subkanthi: @gsantner I just wanted to keep a separate FileInfo class and populate it in a separate class and avoid adding file logic to this Dialog class.

subkanthi: @gsantner I didn't create it, I was just trying to find where the best place would be to add the file.Like /utils

PR #242, opened 2018-05-05
style blockquote

hungryseashell:
Changed the CSS styling of rendered blockquotes to make them more distinct from other parts of the text.
- added left border
- changed color to gray

_Thanks for maintaining this very nice project!_


gsantner:
do you have a screenshot how it looks now? I would prefer it to be quite like github's renderer does it, one continious thick line on the left side

gsantner:
make it so that you think it fits best. I'm letting the fine detail up to you to decide, screen looks good.

hungryseashell:
I added a bit more padding so it is like github's style now and doesn't take up much space. I think it looks better this way.

final

gsantner:
thx! finished? 

btw, open for other PRs ;)

hungryseashell:
yes, this is final.

I'm no android developer, but can help out in easy things like this. Thanks again for this project :)

gsantner:
well if you can compile, install on device and take screenshots thats already quite much, and very enought for the first :D.

Thanks to you too!

PR #243, opened 2018-05-07
Fix new blockquote style for RTL-rendering

ali-hardan:
Thanks all developers...


gsantner:
Hi there, thanks for PR!

This seems more like a quick workaround rather a fix to me. Would be better to move out the blockquote overwrite (e.g. HTML_050) and just not inject that, or inject it like it should be in RTL.

ali-hardan:
overwriting css doesn't make any problem, most RTL websites and WordPress themes use "rtl.css" file to overwriting past css.

But you want to be use another way, ok. I will edit my PR.

ali-hardan:
Is it okay now?

gsantner:
no, because replaceFirst does regex replace, not replace-first-string

ali-hardan:
ok

PR #245, opened 2018-05-08
Updated fr (French) translation

arnaud-jacquemin:


gsantner:
keep it open. ressources do not differ between the branches

PR #248, opened 2018-05-12
Feature fileinfo dialog

subkanthi:

### DIFF
--- a/app/src/main/java/net/gsantner/opoc/util/FileUtils.java
+++ b/app/src/main/java/net/gsantner/opoc/util/FileUtils.java
@@ -341,26 +341,27 @@ public class FileUtils {
             return null;
         }
     }
+
     /**
      * Function to get number of lines and characters.
      */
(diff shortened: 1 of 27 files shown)


app/src/main/java/net/gsantner/opoc/util/FileUtils.java
```diff
@@ -11,6 +11,8 @@
 package net.gsantner.opoc.util;
 
 
+import android.util.MutableInt;
```
gsantner: please don't use android classes in FileUtils - it should stay compatible with java-se

gsantner:
I am accepting this now and do minor modifications. 

Please do BEFORE going on with more development: Reset your repository to be exactly like `gsantner/markor`, make an git branch off that, and send a PR from that. Otherwise you get old commits in the PR, what you can see at this PR and above. 

For updating your repo, it's easiest to delete your fork and do refork now. Further use e.g. `git fetch upstream` and `git remote add` to keep your master updated.

PR #268, opened 2018-05-27
Updated Polish translation

kRkk:
Translated from 64.8% to 84.8%


gsantner:
thanks! 
you forgot to add the xmlns:tools line (top) hence failing check

Issue #284, opened 2018-06-14
Don't add seperator when sharing to linkbox


svemoe:
As stated in #259, I think no seperator should be added when you share a link to linkbox.
This way you get a textfile with on link per line.

Alternatively, there could be an option to turn off seperator completely.

gsantner:
Closing in favor #287

PR #286, opened 2018-06-19
Remove separator when sharing to linkbox and todo

svemoe:
Like I said in my issue #284, I think the linkbox should be an textfile with one link  per line.
You don't need a separator, because the link is a one-liner and so line-ending separates.

This also make the linkbox good for scripting, for example, on my linux machine, I pipe the
file into dmenu to give me a link-selector :smile:

You already made an exception when sharing to todo file, to add no ruler.
But I think there should also be no empty line before the new entry.
In the [Example](http://todotxt.org/todo.txt) from [Todo.txt](http://todotxt.org/) you kann see
one entry per line without empty lines.

This is just my opinion about this issue. If you think different, please tell me.
I would like to discuss.


gsantner:
See #287 what I expect rather just placing something hardcoded (which is empty string too).
If gets fixed please do it that way, rathrr this way.

Issue #317, opened 2018-08-31
Vimwiki support


timokau:
I currently manage my notes on the desktop using vim with the [vimwiki](https://github.com/vimwiki/vimwiki) plugin. It uses a very simple structure: a bunch of markdown files in a folder (flat structure, no subfolders). One of those files in `index.md` and is the starting point. Other files are linked from there. I've looked for an app to edit this on my phone for a while and Markor looks very promising. For basic vimwiki support I think two things are missing:

- an option to open `index.md` by default instead of launching in the folder view
- support for links in the two formats ([for
  reference](https://github.com/vimwiki/vimwiki/issues/318#issuecomment-290600538))
	- `[[some file]]` which equivalent to `[some file|some file.md]` and
	- `[text|some file]` which is equivalent to `[text|some file.md]`

Vimwiki also offers a "diary mode" which is basically just a seperate wiki with some convenience functions to quickly create new files with the current date as the name. Support for that would also be nice but supporting the above two features would already go a very long way to support basic editing on the phone.

timokau:
Great tip about the launcher shortcut, that also solves the diary mode :)

About the link format I decided to just switch to the standard markdown format. I like the explicitness better anyways. For now I'll have to create/fix those links manually until https://github.com/vimwiki/vimwiki/pull/529 is merged. The only remaining issue is that vimwiki uses spaces in links but Markor doesn't recognize those. Spaces are already [implemented](https://github.com/vsch/flexmark-java/issues/78) in flexmark-java and just need to be enabled (`Parser.SPACE_IN_LINK_URLS`).

gsantner:
can you please make a PR with addition of that option?

timokau:
Gladly! #318

Note however that I don't currently develop for android so I don't have a development enviornment set up and I don't want to do that for such a simple change. I made that change in a normal text editor without any testing.

timokau:
When the link thing is fixed (either bei vimwiki/vimwiki#529 or by vimwiki style link support in flexmark-java) you could also a dd "vimwiki" to the app description. I've already looked for an android app to edit vimwiki multiple times and found this app because it already has "vim" in the description. I'm sure I'm not the only one.

Thanks for your good work!

gsantner:
great that it fits your purpose too!

timokau:
For some reasons web-links with spaces (`[a test](www.google.com)`) now work but relative links to markdown files (`[a test](a.md)`) do not. Are they somehow handled differently?

[...]

PR #318, opened 2018-08-31
Allow spaces in markdown links

timokau:
This enables the space-in-links support of flexmark-java to support spaces in markdown links (especially relevant for filenames). See #317.


timokau:
Thanks :)

Issue #326, opened 2018-09-06
Math support in preview


Conceptron:
#### General information

* **App version:** 1.0.2
* **System:** Android 7

#### Description
So, I tried adding math rendering support in preview by using the `Preview->Inject to head` feature (like I did for TOC here: #192) using [katex](https://katex.org/docs/autorender.html) by inserting the below code:
```html
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.10.0-rc.1/dist/katex.min.css" integrity="sha384-D+9gmBxUQogRLqvARvNLmA9hS2x//eK1FhVb9PiU86gmcrBrJAQT8okdJ4LMp2uv" crossorigin="anonymous">
<script defer src="https://cdn.jsdelivr.net/npm/katex@0.10.0-rc.1/dist/katex.min.js" integrity="sha384-483A6DwYfKeDa0Q52fJmxFXkcPCFfnXMoXblOkJ4JcA8zATN6Tm78UNL72AKk+0O" crossorigin="anonymous"></script>
<script defer src="https://cdn.jsdelivr.net/npm/katex@0.10.0-rc.1/dist/contrib/auto-render.min.js" integrity="sha384-yACMu8JWxKzSp/C1YV86pzGiQ/l1YUfE8oPuahJQxzehAjEt2GiQuy/BIvl9KyeF" crossorigin="anonymous"></script>
<script>
    $(document).ready(function () {
        renderMathInElement(document.body, {
            "delimiters": [
                { left: "$", right: "$", display: false },
                { left: "$$", right: "$$", display: true },
                { left: "\\(", right: "\\)", display: false },
                { left: "\\[", right: "\\]", display: true }
            ]
        });
    });
</script>
```
This works in my laptop for a html file as follows:
```html
<!DOCTYPE html>
<html>

<head>
    <!-- <link rel="stylesheet" href="toc-style.css"> -->
    <script src="http://projects.jga.me/toc/dist/common.js"></script>
    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.10.0-rc.1/dist/katex.min.css" integrity="sha384-D+9gmBxUQogRLqvARvNLmA9hS2x//eK1FhVb9PiU86gmcrBrJAQT8okdJ4LMp2uv"
        crossorigin="anonymous">
[...]

Issue #192, opened 2018-02-16
Table of contents


contrapunctus-1:
It'd be nice if Markor auto-generated a ToC in the beginning of documents, both in preview and in export. Maybe only for files above a certain size/length.

gsantner:
Everybody can do this with the new Javascipt /CSS inject feature. I don't want to bloat Markor with lots of stuff. One can use whats in preference - an option is <http://projects.jga.me/toc/>.

Conceptron:
So, I was able to insert a TOC at the beginning of the preview by inserting the below javascript in the `Preview->Inject` settings option.
```html
<script src="http://projects.jga.me/toc/dist/common.js"></script>
<script>
    $(document).ready(function() {
        $("#toc").toc({
            'selectors': 'h1,h2,h3,h4'
        });
    });
</script>
```
and adding a `<div id="toc"></div>` at the beginning of the markdown file. To avoid editing the markdown file, an **Inject to body** setting could be added in which the `div` could be added.

Also, it would be nice if there was either an **`up` button** in the preview to scroll to the top of the document as clicking a link in the `TOC` scrolls to one heading and to navigate to a different heading I have to scroll all the way up again. Instead of the `up` button there could be **`back` and `forward`** button like in a browser too.

This only allows navigation in the preview though, It would be nice if there was a **navigate using toc** feature in the editor too.

gsantner:
Thanks for taking a look!

* Inject to body: Good idea! PR welcome, 90% same like `inject to head` in implementation.
* TOC javascript: I want all defaults (even uncommented includes) to work always and offline. Please don't include the above for now (also don't want to push javascript files to bundled assets, want it to be compatible on desktop too).
* Navigate to top - available from `Special keys` Text-Module. Thought this is maybe something often used, I'm unsure about if and where we would place it seperatly. We already have thaaaaat amount of toggles and actions. Currently more in the process of combining things. Ideas?

809825706:
is the function Navigate to top available？

PR #330, opened 2018-09-11
Added preference options to render math, show TOC, inject to body

Conceptron:
Addresses issues #326, #192.

Added preference options to render math and show TOC, also an `EditText` preference to inject to body.
Rendering math, showing TOC is done by Javascript libraries so they would work only with internet (mentioned this in the preference summaries)

Demo:
demo

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/TextConverter.java
+++ b/app/src/main/java/net/gsantner/markor/format/TextConverter.java
@@ -33,15 +33,33 @@ public abstract class TextConverter {
     protected static final String TOKEN_BW_INVERSE_OF_THEME = "{{ app.token_bw_inverse_of_theme }}";
     protected static final String TOKEN_TEXT_CONVERTER_NAME = "{{ post.text_converter_name }}";
 
-
+    protected static final String HTML_DOCTYPE = "<!DOCTYPE html>";
     protected static final String HTML001_HEAD_WITH_BASESTYLE = "<html><head>" + CSS_S + "html,body{padding:4px 8px 4px 8px;font-family:'" + TOKEN_FONT + "';}h1,h2,h3,h4,h5,h6{font-family:'sans-serif-condensed';}a{color:#388E3C;text-decoration:underline;}img{height:auto;width:325px;margin:auto;}" + CSS_E;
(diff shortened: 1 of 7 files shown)


gsantner:
Is MathTex/MathJax specific to Markdown and it's syntax, or supported in all languages?

E.g. Math in todo.txt is not possible /  wanted.

Conceptron:
Yeah, MathTex/Katex is generally used in Markdown. Also the TOC js lib is for markdown. So, should I add `instanceof` checks in `putContentInTemplate`?

gsantner:
@Conceptron 

I don't like it that way, would like to have markdown stuff in only the markdown converter. But in this case IMO thats the easiest, but we may refactor that at some point. 
Put everything new that can be moved into `markdown/MarkdownTextConverter.java`. Things like `HTML_KATEX_HEADERS`. The `HTML_DOCTYPE` can stay in TextConverter. 

Regarding `onPageLoaded` - it MUST stay, else user injected stuff doesn't work anymore. I see that you have done that in the end of `onPageLoaded1`, but it took me some time. can you please rename it to `onPageLoaded_markor_private` and add comment that it invokes user injected func?

Regarding preferences: Move the options to `Formats -> Markdown` as it is Markdown specific. Also rewrite description to say that it requires an active internet connection and others may track the use.

gsantner:
I changed the opinion on including the stuff. Before Markor gets non-offline features or includes external libs, it's better to have it bundled with app. Could you download the libs, and put it into assets folder? (I will help  to set it up in gradle etc if you can't find out how). Please go that way, than we also can skip the online and tracking string part.

gsantner:
@Conceptron 
We can now easily include parser specific css/js, and run parser specific javascript. See <https://github.com/gsantner/markor/commit/978e84cbc6d04c71addfae639747f624977c6e30#diff-40e019f435d222fc334d28c28b9fe409R109> ;). It is now `putContentIntoTemplate(context, converted, onLoadJs, head);`

Adding math is now just this in `MarkdownTextConverter.java`:
```
        if (appSettings.isMarkdownMathEnabled()) {
            head += HTML_KATEX_INCLUDE;
            onLoadJs += JS_KATEX;
        }
```

PR #340, opened 2018-09-25
Updated zh-rTW (Chinese (Taiwan)) translation

wlctw11:


gsantner:
Hey! Thanks!!

You seem to not have updated Markor in Stringlate. Unfourtunatly some weeks ago there was a nearly 100% percent rework of translations making this unmergable. As about 90% is translated into this language, my please is to just delete Markor in your Stringlate and re-download it  and then send again :)

PR #370, opened 2018-10-03
Improve original string wording

nightrainlily:


gsantner:
@nightrainlily As said in issue, I will apply probably most changes but will that do by hand. So not merging this. Also not forgetting the contributors line change.

Issue #372, opened 2018-10-03
Permission dialog appears twice


MikeMikerson:
App version: 1.2.2
System: Huawei P9 Lite, Android 7.0 API 24

When using the app without giving storage permission, the "Need permission for storage to save documents" dialog appears twice when choosing either "cancel" or "OK"

**To reproduce error**
1. Without storage permission, click either the "To-Do" or "QuickNote" tabs on the bottom navigation.
2. Click "Cancel"

**Expected behavior**
After clicking "Cancel" the first time, the dialog should not pop up again.

MikeMikerson:
Pull request created here: https://github.com/gsantner/markor/pull/373

gsantner:
Hi! This is intended behavior I don't want to change. I sometimes click somewhere to get the dialog just away with sometimes not reading fully.  I want people to understand or at least always have the information available why this is requested.

PR #373, opened 2018-10-03
Remove storage permission dialog from appearing a second time

MikeMikerson:


MikeMikerson:
For issue #372

gsantner:
Hi! This is intended behavior I don't want to change. I sometimes click somewhere to get the dialog just away with sometimes not reading fully.  I want people to understand or at least always have the information available why this is requested.

PR #379, opened 2018-10-04
Fix crash when no calendar app is found. Use toast message to inform …

MikeMikerson:
…the user of the issue


gsantner:
thanks for testing and fix!

Please make the method return true/false and do the toast on the caller side

gsantner:
thanks!

:tada:

Issue #378, opened 2018-10-04
Fix exception: WrFilesystemListFragment IndexOutOfBounds


gsantner:
report from gplay below

![grafik](https://user-images.githubusercontent.com/6735650/46469368-73d27780-c7d3-11e8-9a93-a34a1e218869.png)

notepad104:
What is the behaviour required in case of `_selectedItems` does not have any item for rename or menuInfo item in `onActionItemClicked`

gsantner:
do nothing. Rename should not even be visible for zero items

gsantner:
Hi!

No, copy paste errror log from gplay.

PR #380, opened 2018-10-05
Fixed ArrayIndexOutOfBoundsException for WrFilesystemListFragment

notepad104:
Fixed the ArrayIndexOutOfBoundsException for onActionItemClicked, returning the default behavior in case of no item selected for context_menu_rename and context_menu_info

Issue : https://github.com/gsantner/markor/issues/378


gsantner:
:tada:

Issue #364, opened 2018-10-02
TextAction: color picker


gsantner:
#### General information

* **App version:** v1.2.2
* **System:** Android v7.1.2

#### Description
Add a way to add a color to the text using an color picker. The first step is to have some textaction available which pops up an dialog for color selection.
I don't like to have a big bunch of libs and depencies included, please search for something that does the job without any huge extras  ( like 10 ways to integrate into fragment/activity/preference)..we really just need something simple :smile: .

jakubAerik:
I'll work on it.

gsantner:
@jakubAerik  Tell me when a question arises

jakubAerik:
Please may i ask you, where can i find the id of the text which i write for example in QuickNote activity? What's the name of this id or name of the xml file where the current id is? Thank you for your help :) .

gsantner:
It's bound to _hlEditor in DocumentEditActivity.

For inserting text you usually do   _hlEditor.getText().insert()

jakubAerik:
Please, which method or action cause when i choose bold action(when i click on icon which represent bold text) for example in QuickNote activity on some text and then i click on preview button, text is bold?
I was thinking that's that regexes in SimpleMarkdownParser, but I was wrong.
Thanks.

gsantner:
As I expect it to be it available in multiple formats (plaintext,markdown) I would place the code in CommonTextActions. Just see how other stuff is added (Example: open browser link) and add our new color picker R.drawable and string.

Then also reference the same in the real format specific TextAction (so a button in bottom bar gets generated). The runCommonAction method will execute then.

Refs:

https://github.com/gsantner/markor/blob/master/app/src/main/java/net/gsantner/markor/format/plaintext/PlaintextModuleActions.java#L46


https://github.com/gsantner/markor/blob/master/app/src/main/java/net/gsantner/markor/format/plaintext/CommonTextModuleActions.java#L22

jakubAerik:
Hi :) 
I have almost done your issue, but one problem left...if i select some text for example in plain text and i want to set some color using the color picker, the selected text change the color but only for one milisecond and then, text has again same color as at the beggining :/ i dont know if some method or what has invoke after when i choose color for text from color picker and set deffault color for text in current mode(dark/white)...can you help me with this?

gsantner:
ooh, I think you are doing way more than it was planned! :D It would be enough if you just insert the color into text: ala `#af0622`. 

[...]

PR #394, opened 2018-10-14
color picker

jakubAerik:
Hi, i am creating pull request on issue "TextAction: color picker #364".


jakubAerik:
Okay :) but as you written, you didn't want any other libraries to add into project so i decided for this simple/basic result, using alertDialog :) but okay, i will use this library for your project, no problem :) ;)

gsantner:
Oh, I think have described it little too less. I meant that we dont need a big spectecular thing. I know there are HUGE libs out with like 5 different modes, preference for normal and compat, theming fragments and on and on and on. 
Thats what we dont need, just something simple. Something where you get an simple color selection view you can emebed into app, and select colors from.

jakubAerik:
Hi ;) 
I've pushed some changes so please check it and reply me if u'll have time :)

gsantner:
@jakubAerik Thanks for help! And concrats to first-time merged contribution! :) :tada:

PR #411, opened 2018-10-27
Update string-not_translatable.xml

def14nt:
Changed some small things with the values, see commits

### DIFF
--- a/app/src/main/res/values/string-not_translatable.xml
+++ b/app/src/main/res/values/string-not_translatable.xml
@@ -54,7 +54,7 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="editor_basic_color_scheme_gruvbox" translatable="false">gruvbox</string>
     <string name="auto" translatable="false">Auto</string>
     <string name="hexcode" translatable="false">Hexcode</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/string-not_translatable.xml
```diff
@@ -82,14 +82,11 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     \n}
     \n</style>
     \n
-    \n
```
gsantner: the additional empty lines are there to faster see that these are different blocks. keep it.

as pf JS, keep it with 2 indent. I don't like 2space either, but on mobile you hardly see something in one line with spaces in front.

def14nt:
@gsanter comments? What's wrong with it. Just a small annoyance

gsantner:
I dont want it to be like you commited and said that.

def14nt:
Oh shit, actually didn't see that review. Surprised the FastHub team haven't fixed that yet :/

def0x099:
@gsantner 

> I dont want it to be like you commited and said that

What does this mean, by the way? I'm actually still curious whenever I come across this PR. :+1:

Issue #409, opened 2018-10-27
Cannot edit a HTML file without making it an MD file


def14nt:
I found this out when modifying a HTML file on my device. If I modify a HTML file: `about.html`, it is modified to `about.html.md` which forces a rename on my part.

jakeoneijk:
Hello, If nobody is working on this, I want to try this.

def14nt:
@jakeoneijk go for it mate, thx :)

PR #426, opened 2018-11-04
fix the problem of changing file.ext as file.ext.md(fixes issue #409)

jakeoneijk:
I added the code at loadDocument of DocumentIO(fixes issue #409).
If you want different solution for this, please let me know 👍

### DIFF
--- a/app/src/main/java/net/gsantner/markor/util/DocumentIO.java
+++ b/app/src/main/java/net/gsantner/markor/util/DocumentIO.java
@@ -104,12 +104,18 @@ public class DocumentIO {
                 document.setFormat(TextFormat.FORMAT_PLAIN);
             } else {
                 String oldTitle = document.getTitle();
-                int lastIndexOfDot = oldTitle.lastIndexOf(".");
+                if(oldTitle.contains(".")){
+                    int lastIndexOfDot = oldTitle.lastIndexOf(".");
 
-                //divide oldTitle to document title and file extension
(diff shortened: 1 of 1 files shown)


jakeoneijk:
thank you for feedback!!
In case 1(no extension), Is it better to change filename to README.md after saving ?? or keep the name as README(no extension) ??

gsantner:
There are two reasons it can be README:
* When you open an existing file as such -> don't rename
* When you just make a new document (+) -> input README from titlebar, here yes .md gets added automatically

jakeoneijk:
thank you !!
you said you will remove "auto-title" but 
Is it still ok for me to work on this problem??

gsantner:
It's not written in stone yet how stuff will go then, just have some ideas around. [Started on that](https://github.com/gsantner/markor/tree/fslist_rework) but it slowly progresses ^^..

Yes it's fine if you change this now, this won't go away from today to tomorrow.

jakeoneijk:
thank you so much for quick reply :)

jakeoneijk:
Hello I fix the problem.
but In case 2 (When you just make a new document (+) -> input README from titlebar, here yes .md gets added automatically) 
I didn't have to solve the problem because ,In case2,



extraPathIsFolder become true



so this code be activated.



and file has ".md" on it.



so it will go to this case and will not reach to code that I made

jakeoneijk:
@gsantner Is there anything that I should do more of modify??

gsantner:
My computer broke , sorry.

PR #429, opened 2018-11-13
Md checkbox action

jdckr:
Hi,
I have added a button for checkbox action:
- if there's no checkbox `- [ ]` is added
- if line stats with `- [ ]` it is replaced with `- [x]`

Furthermore I adapted `MarkDownAutoFormat.java` as auto insert of list didn't work for `- [ ]`.


app/src/main/java/net/gsantner/markor/format/markdown/MarkdownAutoFormat.java
```diff
@@ -67,7 +67,7 @@ private String createIndentForNextLine(Spanned dest, int dend, int istart) {
                 // This is for any line that is not the first line in a file
                 Matcher listMatcher = MarkdownHighlighterPattern.LIST_UNORDERED.pattern.matcher(dest.toString().substring(iend, dend));
                 if (listMatcher.find()) {
-                    return dest.subSequence(istart, iend) + Character.toString(dest.charAt(iend)) + " ";
```
gsantner: have you checked that all textactions using this stilll do work?

jdckr: I have checked that auto completion works for unordered list (-, *) and ordered list (1., ...) as well as for checkboxes (- [ ], - [x]). As I don't know what the function covers else I am not sure if that's all. Hoped tp get a hint/finding from your side during the review :-) Is that all what the function covers or is there something missing?

gsantner:
@jdckr You recently worked on that area, I think honestly you know better how it works now :=).
I tried it and have not noticed something major. Only found one issue, that is if you have a list, and go nested it always starts with beginning of line (see example below). This can be resolved later too and I'm fine with this happening for now. Merging ;).

Example (with underscore being a space):
```
Currently on git-master:
- text
- more text
  - subitem press enter
  - this is the new item


Now:
- text
- more text
  - subitem press enter
- this is the new item
```

PR #436, opened 2018-11-24
fixed indentation bug for auto insertion in case of nesting lists

jdckr:
Fixed bug introduced with PR #429 where indentation was not considered for list auto insertion.


gsantner:
thanks :)

Issue #375, opened 2018-10-04
TextAction: Date & Time insertion


gsantner:
#### General information

* **App version:** 1.1.2
* **System:** android 7.1.1

#### Description
The target is to have an TextAction (a button in the bottom bar) which opens a new **dialog** with features below:
* Selection of a date/timeformat, with some default options like `hh:mm`, also there should be an EditText so the user can do an own custom format. (resulting from #260)
* Show below all formats what would be result at the current time object
* Have an button to show an date/timepicker which allows you to select a different time than the current time object. So you could select 05:00 in the timepicker, and then press `hh:mm:ss` as format and you get `00:05:00` inserted into text.
* Add on very bottom a checkbox which changes insertion to text. When checked insert the format string. When not (default) then insert the formatted time (what you get from `SimpleDateFormat.format(dateObject..)`. (resulting from #374)

IzzySoft:
:+1: low prio (as in 90%+ I expect people want to have "now"): consider a date/time picker that also lets you input date/time manually (text-input instead of "month view"). I just had to initialize a new device with LOS, and on selecting current date/time had that stupid picker where you could only skip month-wise – initalized with a day from the year 1971. Happy skipping :scream: 

As said, very low prio – and only if it doesn't cause too much trouble. After all, in the rare cases that would be needed, one could do so manually in the text itself :wink:

gsantner:
@IzzySoft  already on the list :D. (point 3)

tengucrow:
My solution was just palliative for my own notes :) When taking notes on phone I just put them to QuickNotes and separate them by datatimestamps.

It would be nice get date/time picker button or two in toolbar (one button - "get now date/time", and second - "get date/time from calendar", and for both buttons date/time "format" field in settings section (something like for strftime library i.e. "@ %Y-%m-%d %H:%M"). So avrage user can use default values or customize his own format.

gsantner:
Target is to have one option, don't two or more. The reason is that the bar is already quite full, and nobody wants to scroll forever to get the right thing :).

Yeah, the thing with "just paste NOW to text" is what I meant with "have some defaults like mm:ss" :). Should be just one tap to open our datetime dialog and then press the "Now (Time)" button there

petraspeter:
Hi, I would like to pick this one up, if nobody else is working on it.

gsantner:
Seems like nobody started working on it, happy about help ;-)

gsantner:
[...]

PR #454, opened 2018-12-10
TextAction: Date & Time insertion #375

petraspeter:
- combo box for select & edit formats
- view of currently selected format
- date & time picker
- check box for insert format over datetime
- submit / cancel buttons


app/thirdparty/res/layout/time_format_dialog.xml
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
```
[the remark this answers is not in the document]
petraspeter: datetime_picker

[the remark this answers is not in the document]
petraspeter: ahh, I didn't notice I choose wrong folder :)

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextModuleActions.java
```diff
@@ -489,4 +509,190 @@ public void onClick(DialogInterface dialog, int id) {
                 });
         builder.show();
     }
+
```
gsantner: I think this would be worth to place in a seperate class, as this is quite a big amount now of TextActions :D. It's also planned to integrate this picker into all formats (so you also see this date-option e.g. in Plaintext format). So I think it would fit best in `format -> general -> something.java`. 
What do you think?

As of callbacks etc, you can use `Callback.a1<String>` for example, it's very easy fitting in one line :D!

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextModuleActions.java
```diff
@@ -489,4 +509,190 @@ public void onClick(DialogInterface dialog, int id) {
...
+    /**
+     * method 'll add text output to current index in editor
+     * as default will be added datetime based on given DT format
+     * optionally 'll be added DT format
+     *
+     */
+    @SuppressWarnings("RedundantCast")
+    private void addDateTime() {
+        android.support.v7.app.AlertDialog.Builder builder = new android.support.v7.app.AlertDialog.Builder(getActivity());
+        final View view = _activity.getLayoutInflater().inflate(R.layout.time_format_dialog, (ViewGroup) null);
```
gsantner: please call it `root`, `viewRoot` or like that, makes it lots easier to work on this :)

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextModuleActions.java
```diff
@@ -489,4 +509,190 @@ public void onClick(DialogInterface dialog, int id) {
...
+                                datetimeTextView.getText().toString(),
+                                timeFormatEditText.getText().toString()));
+                    }
+                });
+
+        builder.show();
+    }
+
+
+    /**
```
gsantner: JavaDoc, nice :+1:

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextModuleActions.java
```diff
@@ -489,4 +509,190 @@ public void onClick(DialogInterface dialog, int id) {
...
+     * @param isDatetimeSelected
+     *          {@link Boolean} information if we want datetime or format
+     * @param datetime
+     *          selected datetime as {@link String} based on given format
+     * @param format
+     *          {@link String} pattern used to convert datetime into text output
+     * @return
+     *          @datetime or @format, based on @isDatetimeSelected
+     */
+    private String getOutput(Boolean isDatetimeSelected, String datetime, String format) {
```
gsantner: Please use braces, even if not required. 
You can also use ternary operator if you want (` return condition ? datetime : format`), but no must.

gsantner:
@petraspeter Added some comments :).
Thanks for posting the gif, looks very nice to me. Didn't expect it to be already that advanced :tada: .

BTW, all is discussible. Be it above or below, or above. Things are not like 100% fixed.

* From UI view I recommend the following: To use a two-lined item, and display the datetime there. So everybody can see what he gets when scrolling through the list of predefined options. It has not to be millisecond accurate.
* I think if the minute is correct that's fine. Guess nobody cares if it's a minute behind, so guess ~30 secs interval should be very fine. Whats your opinion?

petraspeter:
Some short recapitulation :)
- drawable files should be moved
- text  should be used from strings
- for UI should be used two-lined item ->> (smaller font for preview)
- added custom class for dialog

I was thinking about updating time in select, but honestly I don't see reason why we should do that(also I'm not sure how would I implement it). I thought that it suppose to just show how each format looks like.  
So I add check box use actual time, which will always get current time when user select format & add text 'preview' before actual preview.

Quick look at actual state
datetime_picker2

gsantner:
@petraspeter Wow, looks very good!

If you want to do more improvements and have time feel free todo so, but it's not required!

Some suggestions:
* Reduce margin/padding between dialog elements little (the two checkboxes e.g. make me think it's a layout bug :D)
* `Use actual time`: What does this do? At any time you press, insert the date of press? So if you select that, wait 5 minutes you still have `now` inserted? Else the time of 5 minutes ago?
* Ideas regarding UI view visibility
  * Hide Pick Time/Date when either of the two checkboxes are selected
* Preview
  * ComboBox: Don't prefix with `preview:`, think it just makes it more hard to pick the one user is search
  * Dialog itself: Would put the resulting text on `bottom|end` of dialog, so it's above the  `OK` button
* ComboBox hint: Short so the whole hint is visible

petraspeter:
- ComboBox hint: it was whole hint :)
- Use actual time: it gets you time when you confirm format, or when you stop editing it on your own
Quick view:
datetime_picker3

gsantner:
@petraspeter 
Thank you very much for working on this, it was actually in need quite some time!
I'm merging this now and it will be included in the next update.

I see you have a good feeling for the projec and talented, would be happy to welcome you again in the project. So if you have some time and want todo something feel free to. Can be someting you found self (idea/issue), but asweel the stuff listed on bug tracker.

Gregor

gsantner:
:tada:

PR #509, opened 2019-03-09
Add app shortcuts

### LINKED ISSUE #407: Add 'App Shortcuts' for launchers
There's a new kid on the block: [quick actions](https://developer.android.com/guide/topics/ui/shortcuts/). Would be cool if someone with Java know-how would send a pr for this, looks great.

Whhoesj:
Hi,

This pull request adds shortcuts to the app icon in the launcher.
The shortcuts contain 2 recent documents, the to-do list and QuickNote.
Updated when a new document is added to the recent document list.

[Preview](https://i.imgur.com/C61zcF0.png)

See https://developer.android.com/guide/topics/ui/shortcuts

Fixes #407


app/src/main/java/net/gsantner/markor/util/ShortcutUtils.java
```diff
@@ -0,0 +1,131 @@
...
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N_MR1) return;
+
+        ShortcutManager shortcutManager = context.getSystemService(ShortcutManager.class);
+        List<ShortcutInfo> newShortcuts = new ArrayList<>();
+
+        // Create the to-do shortcut
+        Intent intentToDo = new Intent(context, OpenEditorTodoActivity.class);
+        intentToDo.setAction(Intent.ACTION_VIEW);
+
+        ShortcutInfo shortcutToDo = new ShortcutInfo.Builder(context, ID_TO_DO)
```
gsantner: Know, the recommendation is 4, but the API supports 5 options.
Can you please add/try LinkBox to the static options?

app/src/main/java/net/gsantner/markor/util/ShortcutUtils.java
```diff
@@ -0,0 +1,131 @@
...
+                .setShortLabel(createShortLabel(context.getString(R.string.todo)))
+                .setLongLabel(createLongLabel(context.getString(R.string.todo)))
+                .setIcon(Icon.createWithResource(context, R.mipmap.ic_shortcut_todo))
```
gsantner: Can you please try to use `ic_launcher_linkbox / todo / quicknote` for the static options? This means way less assets :D. If it looks OK, please remove the new quicknote/todo/linkbox pngs

app/src/main/java/net/gsantner/markor/util/ShortcutUtils.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+     * Update the app shortcuts.
+     * The list will contain a link to to-do, QuickNote and 2 recent documents.
+     *
+     * @param context Context
+     */
+    public static void setShortcuts(@NonNull Context context) {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N_MR1) return;
```
gsantner: Please use 
```
if () {
code
}
```

app/src/main/java/net/gsantner/markor/util/ShortcutUtils.java
Whhoesj: Just tried it. It's too much :(
`Caused by: java.lang.IllegalArgumentException: Max number of dynamic shortcuts exceeded`

gsantner: haha, ok thanks for trying :D.

Whhoesj: Also tried defining the static 3 shortcuts in XML and then adding 2 dynamic shortcuts. But then the static shortcuts get overwritten.
So now I've added LinkBox and removed one recent file.

app/src/main/java/net/gsantner/markor/util/ShortcutUtils.java
Whhoesj: It works, but it's not according to the design guidelines. That's why I created new ones.
If the existing resources should be used, they should be upgraded to mipmaps. Those are more fit for launchers.

[Here is a preview](https://i.imgur.com/AYmqLnS.png).

gsantner: Honestly: Prefer the right one, because it's like the launcher :D.
For the normal document I agree that the other should be used

PR #512, opened 2019-03-10
Fix typo

chohner:
just a letter :)


gsantner:
thx!
can you check other strings as well if you can find more? :D

chohner:
gave it another pass, lmk what you thing

PR #528, opened 2019-03-24
README Refactor

def0x099:
I've completely refactored the README. I think this looks a **lot** better, and it removes some of the code that may be harder to read (e.g `<br />` separated list -> `<ul>` list). I've heard you talk about F-Droid having an issue with READMEs before, so tell me if that happens and I'll happily modify it.

[Here it is](https://github.com/def0x099/markor/blob/def0x099-readme-v2/README.md).

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -37,3 +37,4 @@ Where:
 * **[Jawkwon Im](https://github.com/jakeoneijk)**<br/>~° Add keep screen on
 * **[Mark Goldman](https://github.com/goldmanm)**<br/>~° Added sorting for todo.txt, improved wiki
 * **[natanelho](natanelho@gmail.com)**<br/>~° Hebrew translation
+* **[def0x099](https://github.com/def0x099)**<br />~° Refactored README
--- a/README.md
+++ b/README.md
@@ -1,3 +1,12 @@
+<!-- markdownlint-disable MD033 MD041 MD002 -->
+
+<div align="center">
+  <h1>Markor</h1>
+  <img src="./app/src/main/ic_launcher-web.png" width="100">
+  <br />
+  <b>Text editor - Notes &amp; ToDo (for Android)</b>.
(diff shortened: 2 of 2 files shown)


def0x099:
@gsantner Do you think the Google Play and F-Droid images are *really* needed? I'm getting tired of seeing them not loading. What about static images?

gsantner:
Hi again!

Sorry didn't get from the conversation that you want to overhault README itself.

There is a reason for this to be like this - So I can copy-paste it **as is** to F-Droid metadata, Google Play and on my blog. One always doesn't understand markdown hence it's pure html there.

Please remove changes before the ` ### QuickNote` heading. You can improve the wording if needed in the points, but keep emoji and the br stuff, required for both stores (fdroid/gplay).

def0x099:
@gsantner Could you expand on this?

> So I can copy-paste it as is to F-Droid metadata, Google Play and on my blog.

A `<ul>` list *is* valid HTML. I'm not sure I follow. 

p.s Could you send these in as reviews for clarity?

def0x099:
Honestly if Google Play or F-Droid can't understand `<ul>`s, that's a fault on their behalf and an issue should be raised over there.

gsantner:
Dont expect Gplay or fdroid fix this because of this minor request on a short time, if at all.
Deal with, thats it. ;)

def0x099:
Other people can do it on Google Play, though. :thinking: 

Screenshot_20190324-233506

gsantner:
Next up try it in F-Droid.

Please just leave that part, and the Screenshot section as is. This is tested and just works for all targets. Those who read readme or app description don't care about the Format.

def0x099:
@gsantner Why not use [this](https://en.m.wikipedia.org/wiki/Interpunct) instead of emojis that provide no additional information?

gsantner:
because there is a thing called App store optimization.

Sorry but I want to keep that part as is. Maybe new or updated lines, but in general that format.

As said, the markdown showcase document should anyway be different than the readme.

def0x099:
@gsantner That's great. You have provided no reasons as to why, but OK.

Issue #549, opened 2019-04-28
Minor bug - file information


aubreyz:
#### General information

* **App version:** 1.7.5 (67)
* **System:** Moto g6

#### Description

When a single file is selected and file "Details" requested from menu, the details report shows the name of the folder where it should show the name of the file (in screenshot below the actual file name is barcelona.md not travel - which is the folder name)

![Screenshot_20190428-115406](https://user-images.githubusercontent.com/25100168/56863362-4aff1300-69ad-11e9-847c-5330adf0f1eb.png)

aubreyz:
I confirm now working well on 1.7.7 - thanks

PR #555, opened 2019-05-03
This commit corrects the filename in the file dialog (see #549)

melkor0:
There was a `getParentFile()` too many,


gsantner:
you can include this in the other merge request, so you dont get merge conflicts

PR #560, opened 2019-05-05
Archive Todo: Add line break between old/new tasks

olistroemer:
trim removes the trailing new line character from the archive file. Therefore the new tasks are glued directly to the last task of the archive:

x 2019-05-02 2019-05-02 Spargel holenx 2019-04-22 2019-05-04 Unterhose wechseln

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
+++ b/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
@@ -199,7 +199,7 @@ public class TodoTxtTextActions extends TextActions {
                                 if (doneFile.exists() && doneFile.canRead()) {
                                     doneFileContents = FileUtils.readTextFileFast(doneFile).trim() + "\n";
                                 }
-                                doneFileContents += TextUtils.join("\n", move);
+                                doneFileContents += TextUtils.join("\n", move).trim() + "\n";
                                 if (FileUtils.writeFile(doneFile, doneFileContents)) {
                                     // All went good
                                     _hlEditor.setText(TextUtils.join("\n", keep));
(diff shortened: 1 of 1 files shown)


gsantner:
thanks, however I cant reproduce your issue. I'm using the archiving quite often and have a clean list with all x in front ;).

If you do add \n, please trim the appended text, to ensure there wont be a whole empty line

olistroemer:
That's really strage :D. I just tried it again with a different todo-file: same problem. This "bug" bugged me since I started using this app. I can't really see, where in the code the new line character could come from.
```
String doneFileContents = "";
if (doneFile.exists() && doneFile.canRead()) {
    doneFileContents = FileUtils.readTextFileFast(doneFile).trim();
}
doneFileContents += TextUtils.join("\n", move);
```
join() only adds \n in between the parts, doesn't it? There shouldn't be a reason to trim() again...

Or am I crazy :D?

olistroemer:
@gsantner I'm not sure if you automatically got a notification regarding my reply. Therefore I mention you in this comment just to be sure ;).

gsantner:
@olistroemer  there is always this one editor (on any OS) that inserts one or two empty lines on ending.
thats why I asked to trim everything, and only have existingtext.trim + \n + newtext + \n

olistroemer:
@gsantner How about this :D?

> @olistroemer there is always this one editor (on any OS) that inserts one or two empty lines on ending.

Actually that would be an interesting issue in its own. Markor doesn't end the last line with an \n, if you don't put it there on purpose, which isn't the unix-way for text files[1]. Is this the way you prefer or don't you hold a strong opinion on this matter?

[1] https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

gsantner:
I prefer ending files with newline overall, thats the common way.
Only exception is the todo.txt file itself where the pointer should start on the last char of the last line.
For done file this doesnt matter, so just end it with newline.

olistroemer:
@gsantner Good :) I added my new proposal commit to this pr. I hope it suits your expectations :+1:

gsantner:
:tada:  thanks for your first time contribution. Help is always welcome, so feel free to further make PRs ;)

Issue #600, opened 2019-07-03
File browser: Foldername/Title not visible/updated sometimes


ema-pe:
#### General information

* **App version:** v.1.8.3 (75)
* **System:** LineageOS 14.1 (Android 7.1.2)

#### Description

As show in the attached video, when I move to another section and go back so `Files`, the app's title name suddenly disappears.

Video: https://streamable.com/s/13m5u/vcwvwc

PR #606, opened 2019-07-06
Fix folder title not visible sometimes #600

tomdaly:
Ensures that 'Files' tab has title set correctly by getting current
folder of FilesystemViewerFragment before setting

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/MainActivity.java
+++ b/app/src/main/java/net/gsantner/markor/activity/MainActivity.java
@@ -84,7 +84,7 @@ public class MainActivity extends AppActivityBase implements FilesystemViewerFra
     private ActivityUtils _contextUtils;
     private ShareUtil _shareUtil;
 
-    private String _currentTitle;
+    private String _cachedFolderTitle;
 
 
     @Override
@@ -231,6 +231,8 @@ public class MainActivity extends AppActivityBase implements FilesystemViewerFra
         } else {
             getWindow().clearFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
         }
+
+        cacheCurrentFolder();
     }
 
     @Override
@@ -254,7 +256,7 @@ public class MainActivity extends AppActivityBase implements FilesystemViewerFra
                     } else {
                         _toolbar.setTitle("> " + currentDir.getName());
                     }
-                    _currentTitle = _toolbar.getTitle().toString();
+                    _cachedFolderTitle = _toolbar.getTitle().toString();
                     return;
                 }
             }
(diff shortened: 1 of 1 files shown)


app/src/main/java/net/gsantner/markor/activity/MainActivity.java
```diff
@@ -394,7 +395,8 @@ public void onViewPagerPageSelected(int pos) {
         (_lastBottomMenuItem != null ? _lastBottomMenuItem : menu.getItem(0)).setChecked(false);
         _lastBottomMenuItem = menu.getItem(pos).setChecked(true);
         updateFabVisibility(pos == 0);
-        _toolbar.setTitle(new String[]{_currentTitle, getString(R.string.todo), getString(R.string.quicknote), getString(R.string.linkbox), getString(R.string.more)}[pos]);
+        FilesystemViewerFragment fragment = (FilesystemViewerFragment) _viewPagerAdapter.getFragmentByTag(FilesystemViewerFragment.FRAGMENT_TAG);
```
gsantner: it may not happen often, but it is possible that FilesystemViewerFragment is null. can you fallback to _currentTitle in that case? (`frag == null ? _currentTitle : getCurrent..`)

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
```diff
@@ -394,7 +395,8 @@ public void onViewPagerPageSelected(int pos) {
...
+        _toolbar.setTitle(new String[]{fragment.getCurrentFolder().getName(), getString(R.string.todo), getString(R.string.quicknote), getString(R.string.linkbox), getString(R.string.more)}[pos]);
```
gsantner: it may not happen often, but it is possible that FilesystemViewerFragment is null. can you fallback to _currentTitle in that case? (`frag == null ? _currentTitle : getCurrent..`).

tomdaly: Thanks for your reply I will make those changes, but while testing I found that the reason for the bug is that `_currentTitle` is never set (so is blank) as lines 246-261 are never run in the BroadcastReceiver. The title is actually set at line 428 when the folder changes.

Instead of changing it to `_currentTitle` if null, should I change it to `getResources().getString(R.string.app_name).toLowerCase()` so that if FilesystemViewerFragment is null, it displays the app name (_markor_) as a backup?

gsantner: sounds good, but you can also fix the cache to save at correct point :D

tomdaly: Updates made including a cached folder variable and fallback to app name :)

gsantner:
@tomdaly I see you pushed, but don't know if you completed. Is it finished?

tomdaly:
@gsantner the edits are complete, it should be ready to merge.

I have compiled the app and tested it on my phone (Samsung Galaxy S8) and it appears to be working properly.

gsantner:
:tada: 
Thanks for your contribution :+1: . More is always welcomed, if you like to. The project really needs some contributors :D. Doing this project pretty much 99% alone since the beginning.

Issue #610, opened 2019-07-08
Sort todo by date (created||finished)


esmoak:
Adding the ability to sort by date in addition to tag, priority, and category would greatly improve the power of the todo format for things like workout or reading logs.

Related to #211

tomdaly:
Hi @gsantner I think I can complete this enhancement. Shall I assign myself to this issue and develop it?

gsantner:
@tomdaly 
Hey nice thanks!
Let me know when you need some help.

There should be some kind of comparable/comparator, where you need an additional one (see other todo.txt sort stuff in SttCommander

PR #614, opened 2019-07-16
Add ability to sort todo by date ascending or descending

tomdaly:
Solves enhancement request #610
Also fixed typo 'serach_or_custom' -> 'search_or_custom' in all files
'Date (ascending)' and 'Date (descending)' uses date in strings.xml so that it works in all languages

### DIFF
--- a/app/src/main/java/net/gsantner/opoc/format/todotxt/SttCommander.java
+++ b/app/src/main/java/net/gsantner/opoc/format/todotxt/SttCommander.java
@@ -13,6 +13,7 @@ package net.gsantner.opoc.format.todotxt;
 import net.gsantner.opoc.format.todotxt.extension.SttTaskWithParserInfo;
 
 import java.text.SimpleDateFormat;
+import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.Date;
@@ -378,6 +379,10 @@ public class SttCommander {
                     difference = compare(x.getProjects(), y.getProjects());
                     break;
                 }
+                case "date": {
+                    difference = compare(x.getCreationDate(), y.getCreationDate()); 
+                    break;
+                }
                 default: {
                     return 0;
                 }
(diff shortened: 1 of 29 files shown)


gsantner:
Thanks! I'm currently on vacation and can't look into it much (dont have a computer with me).

app/src/main/java/net/gsantner/opoc/format/todotxt/SttCommander.java
```diff
@@ -412,5 +417,15 @@ private int compare(List<String> x, List<String> y) {
             return x.get(0).compareTo(y.get(0));
 
         }
+        private int compare(String x, String y) {
+            SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd");
+            try {
+                Date xAsDate = sdf.parse(x);
```
gsantner: any of both may be null or a invalid format. In that case it will be either nullpointer or invalidformatexception.

You could do e.g. two try catches, and assign 0 to x/y in case of catch clause.

app/src/main/java/net/gsantner/opoc/format/todotxt/SttCommander.java
```diff
@@ -412,5 +417,15 @@ private int compare(List<String> x, List<String> y) {
             return x.get(0).compareTo(y.get(0));
 
         }
+        private int compare(String x, String y) {
+            SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd");
```
gsantner: Use class constant `DATEF_YYYY_MM_DD`, saves object construction on each comparision step.

app/src/main/java/net/gsantner/opoc/format/todotxt/SttCommander.java
gsantner: Or even better, simple string comparision should be enough!

As the format is yyyy - mm - dd, its already in good shape for default (aphabetical) sort. The format prefixes also 0 e.g. `07`, so ideal.

app/src/main/java/net/gsantner/opoc/format/todotxt/SttCommander.java
tomdaly: Fixed

app/src/main/java/net/gsantner/opoc/format/todotxt/SttCommander.java
gsantner: Date format in todo.txt is part of specification and not to be changed But it's fine

gsantner:
Thanks again!

tomdaly:
No problem, happy to contribute to a great app :)

PR #644, opened 2019-08-07
WIP: File browser selection bugs

### LINKED ISSUE #641: File browser: Minor selection issues
###  General information

- **Android version**: 8.1.0
- **Markor version**: 2.0.4

### First bug

If you select a file and then press the back button without deselecting that file, you go to the previous level in the file system with that file still selected, which means that you cannot enter any folder because you're in select mode and any press just makes that folder selected.

I think the best solution would be to deselect everything when you move to an upper path.

### Second bug
When you move a file to another directory, the file stays selected (in the original location), and you have to manually update the folder for the changes to appear. In fact, you can still choose to move that same file again, though it doesn't actually have any effect.

It probably would be less confusing to deselect all the files that have been moved and update the current folder afterwards.

ivenk:
Fixes #641 (mostly)

This PR fixes the issues described in #641 by clearing the current file selection at the appropriate times. 

**Files not deselected after directory switch**
This is fixed by deselecting everything when the user presses the back button while in selection mode.

**Files stay selected after moving and are still visible in origin**
Once the FileSystemViewerDialog completes(not canceled) the current selection is cleared and the selection mode terminated. The item is still visible in the original location but is crossed out.

Im still looking into hiding the moved files instantly.

I tested all changes locally.

### DIFF
--- a/app/src/main/java/net/gsantner/opoc/ui/FilesystemViewerAdapter.java
+++ b/app/src/main/java/net/gsantner/opoc/ui/FilesystemViewerAdapter.java
@@ -361,6 +361,7 @@ public class FilesystemViewerAdapter extends RecyclerView.Adapter<FilesystemView
     public boolean goUp() {
         if (canGoUp()) {
             if (_currentFolder != null && _currentFolder.getAbsolutePath() != null && _currentFolder.getParentFile() != null && !_currentFolder.getParentFile().getAbsolutePath().equals(_currentFolder.getAbsolutePath())) {
+                unselectAll();
                 loadFolder(_currentFolder.getParentFile());
                 return true;
             }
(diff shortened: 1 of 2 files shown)


gsantner:
Thanks!

Please let me know when it's ready to merge :)

ivenk:
In order to fix the last issue i took a look at file deletion. After a file ist deleted the entire folder gets reloaded with FilesystemViewerAdapter::reloadCurrentFolder. I now use the same for moving files. It fixes the issue and seems consistent with the rest.

ivenk:
@gsantner should be done now.

gsantner:
@ivenk 
Thanks for contributing!
I'm merging now.

You are very welcome for more contributions, features and fixes :).

ivenk:
@gsantner no problem. Glad i could help !

Issue #620, opened 2019-07-23
Disallow renaming to existing file (overwrites target with source content)


junior291492659:
#### General information

* **App version:**  2.0.1
* **System:** Android 6.0 (can be reproduced on an Android emulator 6.0 and a Redmi Note 4X phone )

#### Description
    
  I install the version 2.0.1 and start the app, click the floating button to create a new file named "test1", set the type as "Plain Text", click ok then it get to the "edit" page of "test1.txt" and I set text with "Test1", a file "test1.txt" containing "Test1" is created successfully.
    
  Then I create a new file named "new" again, set the type as "None", click ok then it get to the "edit" page of "new" and I set text with "New", a file "new" containing "New" is created successfully as well.
    
  But then I long click the "new" file, click the "rename" icon in the action bar, and set its name as "test1.txt", click ok and the  previous "test1.txt" is overwritten since its content is "New" instead of "Test1".
    
  I have noticed that there is a hint  written in the dialog when creating a new file or folder, which tells user that files and folders are not overwritten. Indeed, if I set a file name and file type which is the same as one of the existed file, it will get to the "edit" page of the existed file to avoid overwriting. But the scenario I described above is in conflict with this principle. This can be a bug somehow and perhaps a pop-up hint would be helpful.

#### Reproducing video
[https://youtu.be/o-uy-n_RwGY?t=34](https://youtu.be/o-uy-n_RwGY?t=34)

I have reproduced it successfully in the latest version 2.0.2 in the google play.
It would be really appreciated to get your concern on this.

     
## Notes
* Pre state
  * new: "new"
  * test1.txt: "Test1"
* Post state (rename new -> test1.txt)
  * ~~new~~ file deleted
  * test1.txt: "New"

gsantner:
Goals, requires modification of of rename dialog class:
* Validate input, check if input renamed file exists
* In case exists, show red error description in dialog content. Do nothing on click or disable the `ok/rename` button.
* Validate if input filename is valid. (i.e. contains no illegal character)

tomdaly:
Hi @gsantner, I'd like to have a go at adding this enhancement if you can assign me :)

gsantner:
You can always start, I'm happy to review again. thanks

tomdaly:
@gsantner almost completed the fix but need to add a new string "File already exists! Please choose a different filename", can you recommend a good way to translate this new string and add it to each language strings.xml?
Thanks

gsantner:
[...]

PR #646, opened 2019-08-08
Fix renaming overwriting existing file/folder

tomdaly:
Fixes issue #620 

Modifications made to check for filename clash, and if filenames clash the 'OK' button is disabled and a red text error is displayed informing the user that the file/folder names clash, which disappears when the filename is changed again. Error string added to strings.xml.


gsantner:
Thanks! I'm having a look on computer later

gsantner:
I'm on mobile and see couple of minor things I would change, but need PC for that. 
Thanks for your improvements its gonna be merged with slight changes.

Issue #652, opened 2019-08-19
Hugo/Notable markdown template


daniele-athome:
I'd like to request adding of a new Markdown template based on [Hugo front matter](https://gohugo.io/content-management/front-matter/), also used by [Notable](https://github.com/notable/notable):

```
---
tags: [tag1, tag2, tag3]
title: Note title
created: '2019-07-29T14:47:15.537Z'
modified: '2019-07-29T14:47:36.296Z'
---

# Note title

Note content
```

Would you be interested in a pull request? I can implement the feature myself. Also since this kind of template needs access to user data, I could also modify `getTemplateContent` (and maybe the UI in the future too to input tags) to work with those. What do you think?

gsantner:
> Would you be interested in a pull request? I can implement the feature myself.

Of course! Thanks for PR! :) :smile: All help is welcome

PR #653, opened 2019-08-19
[WIP] Add template for Hugo

daniele-athome:
This is a basic idea of what I had in mind in #652. Before continuing I would really like some feedback from the team. Thanks!


gsantner:
Is it possible to have date in filename at hugo, like Jekyll has? `yyyy-mm-dd-title-here.md` 

I would prefer to have date out of content and if possible with as low amount/0 of parameters as possible. At file creation its important to be simple so every user can easily create files

daniele-athome:
> Is it possible to have date in filename at hugo, like Jekyll has? `yyyy-mm-dd-title-here.md`

I kept the timestamp inside (created field in the front matter) to be compatible with Hugo and Notable. How about having both formats? I could add another template using the format you described (without the created field in the front matter). But remember that neither Hugo nor Notable will understand the creation date in the filename.

> I would prefer to have date out of content

Other templates have the timestamp in content already :thinking: 

P.S. those templates should definitely be user-definable by the way :-)

gsantner:
> How about having both formats


No, please have only one that works for both. reusability and covering multiple usecases is better :-). 


> P.S. those templates should definitely be user-definable by the way :-)


I know. And thats somewhere on the roadmap, but no fixed milestone. This all is a pre-phase for that, but in any way I want to keep it simple KISS.

daniele-athome:
> No, please have only one that works for both. reusability and covering multiple usecases is better :-).

Ok then, that would mean:

* created field in front matter with current timestamp
* filename with format: yyyy-mm-dd-title.md with current date

I see that you don't process the filename in any way so title would have to be a valid filename, but that's ok by me.

I'll rename the template to "2029-01-01-front-matter.md" to follow the Jekyll post one, is that ok?

gsantner:
The point is, does the date matter in the filename at all? At jekyll it's a MUST so the post gets published

daniele-athome:
> The point is, does the date matter in the filename at all? At jekyll it's a MUST so the post gets published

No. The filename in Hugo and Notable is completely ignored. All metadata is read from the front matter.

gsantner:
Ok, then keep date out of filename

daniele-athome:
Ok so nothing else to do in this case, right?
I've added myself to the contributors list (may I? :-) )

gsantner:
sure who contributes fits there, doesnt matter how much.

just I didnt have a look into code yet, wiol let you know on computer then

thanks for contributing to the project!

gsantner:
In Jekyll I found dates in format like this `date: '2017-05-20 17:45:55'`. Does that work at your both cases too?

Than I would prefer a dtformat that works for all 3

daniele-athome:
Oh damn... I'm sorry I misread Hugo documentation... actually Hugo and Notable are not compatible :-(

Hugo uses `date` and `lastmod` for creation and last modified time.
Notable uses `created` and `modified` instead.

[I've opened a feature request with Notable](https://github.com/notable/notable/issues/787) to see what they think about it. Please hold this PR until I get back to you - let's see what happens.

daniele-athome:
Sorry, the two front matter formats are incompatible and it would be close to impossible to make them agree on following the same front matter standard - honestly I don't have the time or strength to follow such a process. One day there will a standard, that day is just not today.

That being said, I'll modify the source code once more so I can have 2 templates: one for Hugo and one for Notable - so I can use Markor for both.
I can see you'd like to avoid creating 2 different templates, but I don't want to convince you to follow that path just because I need Markor to do that :-) I'll just use my fork on my device. You can of course merge this PR and use it as you see fit.

app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
```diff
@@ -177,24 +180,40 @@ private void callback(boolean ok, File file) {
...
+                break;
             }
             case 3: {
-                return "---\nlayout: post\ntags: []\ncategories: []\n#date: 2019-06-25 13:14:15\n#excerpt: ''\n#image: 'BASEURL/assets/blog/img/.png'\n#description:\n#permalink:\ntitle: 'title'\n---\n\n\n";
+                template = "---\nlayout: post\ntags: []\ncategories: []\n#date: 2019-06-25 13:14:15\n#excerpt: ''\n#image: 'BASEURL/assets/blog/img/.png'\n#description:\n#permalink:\ntitle: 'title'\n---\n\n\n";
+                break;
             }
             case 4: {
-                return "# Title\n## Description\n\n![Text](picture.png)\n\n### Ingredients\n\n|  Ingredient   | Amount |\n|:--------------|:-------|\n| 1             | 1      |\n| 2             | 2      |\n| 3             | 3      |\n| 4             | 4      |\n\n\n### Preparation\n\n1. Text\n2. Text\n\n";
+                template = "---\ntitle: {title}\ntags: []\ncreated: '{timestamp}'\n---\n\n# {title}\n";
```
gsantner: When finished template, please also put a normal (file) copy in the git->samples folder. That gets used for testing and quickly showing people how things look.

daniele-athome:
> Please let's make this a Hugo template as that's the more common use case. I think Notable fits better for custom template later on. I assume that the project will not adapt to other standards on the fast, or even more get a release out to users fast.

Ok, I'll make it a template for Hugo.

> Is there some kind of special filepath part in hierachy at notable? If it's lets say /path/notable/stuff/, I would be open to have e.g. if (folder.getAbsPath().contains(/notable/)) replace(date, otherdatekey) in the switch-case 4.

I have a synced folder via 3rd party software that has both Hugo and Notable content, in two different folders.
Hugo files are just sparse in different subfolders.
Notable has a dedicated folder which has two subfolders (this is forced by Notable itself):

* notes: the actual markdown files
* attachments: non-markdown files referenced by the notes

[offtopic]I can see I cannot choose a subfolder when I create a new file, that would be a nice addition too by the way (I mean when you create a new note from a sharing intent). And you should definitely remove the new folder button, it's very confusing.[/offtopic]

> Can you also confirm me that you do use notable, but not hugo?

I use both, for different purposes. That's why I wanted to make 2 templates.
I use Notable as a note taking app and Hugo to write stories and diaries (processed via Hugo for web reading and pandoc to generate an e-book).
[Notable has a mobile app in their plans](https://github.com/notable/notable/issues/23), but it would be a non-native (and young) app so I would prefer to continue to use Markor which is an excellent app.

Thanks for the review! I'll implement your suggestions in the next 2-3 days.

gsantner:
Create file `.notabledir` on your device to have structure like this

```
./.notabledir
./notes
./attachments
```

and do `if file(file.getparent, .notabledir) .exists` -> replace

gsantner:
@daniele-athome 
I implemented it as wanted on my own as there was no further response.

Create a file or folder `.notabledir` above your posts folder and front matter-hugo files wil be created with notabledir syntax.

daniele-athome:
I'm sorry @gsantner something came up and the time I allocated for this vanished. Thank you :-)

gsantner:
np!

Issue #367, opened 2018-10-03
TextAction: Add voice to note (audio recording)


gsantner:
#### Description
Often spoken is the easiest way to take notes. It would be great to just be able to press a microphone button and a start/stop audio recording thing popup. 

Target of this:
* Create new audio recording button
* Popup dialog which has options for start/stop and cancel
* Save recorded audio in same folder as current document
* Add filename to current cursor position in text

### Suggestion
What I liked greatly so far is [this project](https://android-arsenal.com/details/1/6797), which basically resolves most important points. Sadly it doesn't allow set save paths. But as 99% of the important things is done [in a single file](https://github.com/IvanSotelo/RecordDialog/blob/master/recorddialog/src/main/java/com/deskode/recorddialog/RecordDialog.java) I think the best option is to copy/fork the file and adapt it to Markor needs. We don't need stuff like Hangouts ringetone when pressing start or fancy start animation for example:-).

![https://raw.githubusercontent.com/IvanSotelo/RecordDialog/master/20180521_093731.gif](https://raw.githubusercontent.com/IvanSotelo/RecordDialog/master/20180521_093731.gif)

stupidly-logical:
That library is very outdated.

stupidly-logical:
Can I use some other library if it satisfies your specifications?

gsantner:
Yes, but I had some research and all seem to have a fixed savepath sadly :/

poutrosgit:
Could it be possible to select what format to save in? Given that it's a voice note, a lossless (like wav as shown, or flac since it's open) or a lossy format that is good for voice (like opus) would be good to work with. Maybe flac would better by default because of needing less space storage than wav. Support is almost universal and voice notes aren't expect to be played anywhere else other than the phone of origin so saving in flac shouldn't be an issue for interoperability.

gsantner:
when the lib/methods support it yes. But for sure we won't have something like ffmpeg here :-).

ivenk:
I would like to give this a try. I have already started on but it might take a couple of days until i can submit a pr.

gsantner:
Nice! :D

ivenk:
Okay back on it! Hat some professional stuff to sort out

gsantner:
Closing as implemented :tada:

PR #663, opened 2019-08-25
[WIP] TextAction: Record audio

ivenk:
This is my attempt at  #367 .

Basically I implemented a simple dialog with the dialogfragment with 3 Buttons Play/Stop, Save, Cancel. I do the recording with androids mediaRecorder. The output format currently is .3gp but that should be simple to change. The audio files are stored in the same folder as the text file. The name is : {original_file_name_without_extension}_audio.3gp . In case of multiple audio files for the same text file the names become : {original_file_name_without_extension}_audio_{increasing_number}.3gp .

The current version satisfies the 4 original targets:
- Create new audio recording button
- Popup dialog which has options for start/stop and cancel
- Save recorded audio in same folder as current document
- Add filename to current cursor position in text

Im open for any ideas how to improve this solution.
What output format would be the best choice ?


ivenk:
presentation

gsantner:
@ivenk ogg is smaller so I would recommend to go for that, otherwise wav is fine too. But I would not go for 3gp.

For .wav see the [recorddialog](https://github.com/IvanSotelo/RecordDialog/blob/master/recorddialog/src/main/java/com/deskode/recorddialog/RecordDialog.java#L199) for reference how it works. For .ogg it shold be as simple as setting a flag.

ivenk:
@gsantner thank you for your feedback i will try and change it to ogg. 
Currently looking into the layout of the dialog.

I will format the code and merge/rebase once im happy with the rest.

gsantner:
OK! Just as note, simple UI is enough - start/pause/stop and on bottom Cancel/OK

ivenk:
Thank you @gsantner !
I will keep it simple.

gsantner:
sorry that the code review bot is not very mercy 😅

gsantner:
if I have to choose between (ogg + no extra lib) OR (wav + extra lib) I prefer ogg.

ivenk:
No problem! I will fix the formatting issues.

I looked into ogg. However mediarecorder support for ogg was only added in api level 29. [link](https://developer.android.com/reference/android/media/MediaRecorder.OutputFormat.html#OGG)
That is why i decided to go with .wav instead. I could still look into ogg but i don't see a way to avoid an extra lib either way. What are your thoughts on this @gsantner ?

gsantner:
Oh. Right, Sorry.

I just checked out the library and its not maintained since years. Gradle-android is at api24 which tells me not much gonna happen there anymore.

Do you think this sample works? https://stackoverflow.com/a/29148087

gsantner:
Oh. Right, Sorry.

I just checked out the library and its not maintained since years. Gradle-android is at api24 which tells me not much gonna happen there anymore.

Do you think this sample works? https://stackoverflow.com/a/29148087

ivenk:
Don't worry about it ! 😄 
I will take a look at the sample and try it out.

gsantner:
Hey!

I'm working on an own solution, based on OmRecorder too. I'm closing this.

Thanks for checking this out! :)

PR #681, opened 2019-09-20
Avoid showing permission denied warning when editing files on private storage

gnu3ra:
In some cases users may choose to store notes on the internal storage (/data/user/0 for instance), which adds protection against apps reading /data/media/0. 

ShareUtil.java mistakenly marks this location as non-writable, forcing the warning dialog to be shown in error. This patch adds private storage as a writable location.


gsantner:
Thanks!

Issue #584, opened 2019-06-20
TextActions: Indent/deindent buttons


frozencemetery:
#### General information

* **App version: v1.8.1 F-droid** 
* **System: Android 8.0.0** 

#### Description

It would be nice to have two buttons, one "indent" which adds four spaces to the start of the line, and one "deindent" which removes four from the start of the line (if they exist).

In addition to nesting lists, this also makes it slightly easier to work with blockquotes.  (This is my attempt to file #582 again, hopefully in a more useful and lightweight form this time.)

Thanks!

carlosolimpio:
Hello, I've made a PR for this issue, could you help to review it? Thanks

PR #713, opened 2019-10-03
TextActions: Indent/Deindent (feature #584)

carlosolimpio:
Hello, just making this PR for the hacktoberfest! Hope it helps.
#584


gsantner:
Hello! Thanks :)

Can you please backup your repository local (zip e.g.) and then run `git rebase origin/master`?

carlosolimpio:
@gsantner done! :)

gsantner:
@carlosolimpio 
Can you please check in the logic if current text contains "sw=2" or "shiftwidth=2", and use 2 instead of 4 spaces in that case?

This is used at VIM to configure space amount on a file level. for advanced users this allows to control it also in Markor by a small change.

https://github.com/gsantner/geany-plugin-gsantner-utils/blob/master/geanygsantnerutils.c#L20

app/src/main/java/net/gsantner/markor/ui/hleditor/HighlightingEditor.java
```diff
@@ -150,6 +150,18 @@ public void insertOrReplaceTextOnCursor(String newText) {
...
     }
 
+    public void deIndentCurrentLine() {
+        int start = getSelectionStart();
+        int end = getSelectionStart() + 4;
+        String fourFirstChars = getText().toString();
+
+        if (end <= fourFirstChars.length()) {
+            fourFirstChars = fourFirstChars.substring(start, end);
+            if ("    ".equals(fourFirstChars))
```
gsantner: use braces, less error prone

carlosolimpio:
Hi @gsantner can you clarify that sw check?
I've found in crowdin.yml the line `# vim: sw=2 ts=2 noexpandtab:`. So, you want me to read this info from this file and make the logic according to sw value?

gsantner:
@carlosolimpio 
Yeah but only in very simple manner:
```
String text = getText.toString()
int sw=4;
if (text.contains("sw=2") || text.contains("shiftwidth=2")) {
sw=2;
}
if (text.contains("sw=8") || text.contains("shiftwidth=8")) {
sw=8;
}
---your code that works with 4 so far -----
-----
-----
```
Please don't think about parsing or alike, simple string contains is very enough.

gsantner:
Do you need more information @carlosolimpio ?

carlosolimpio:
Yes, I need @gsantner. Do you have any idea how I'm supposed to read that from the file? Is there any specific piece of code or just use the java way of reading files?

gsantner:
I pasted snippet above.

You dont read that from any file. Just from what the edittext currently contains, thats all you need and you have it available.

carlosolimpio:
Hello @gsantner, is everything with my code ok?

gsantner:
@carlosolimpio 
Thanks for your help! I'm merging this now.

Congrats on your first PR on markor :tada: :smile: .
Would be happy about more contributions, be it during hacktoberfest or the anytime at year.

Issue #699, opened 2019-09-30
Editor & Viewer: Reading auto-scroll


gsantner:
#### General information

* **App version:**  v2.1
* **System:** v9

#### Description
Currently Markor's focus is mainly on writing and navigating text. This a task out of this years main tooic for hacktoberfest: **Reading**.

The goal of this task is to add a menu toolbar button which enable auto scrolling of text.

* This should be available for both View mode (webview) and Edit mode (EditText)
* Goal is to jump every X milliseconds by a page (or e.g. 20 lines of text).
* The speed (milliseconds interval) should be configurable. Easiest is in settings.

Please let me know when YOU start on this, need information or any help.

UtnansN:
Hi, I would like to attempt this. One question though - should both the jump time and the jump line count be adjustable? Or should it just be the millisecond interval?

gsantner:
@UtnansN 
For now use constants like
```
void func(param1, param2, ...) {
  final int interval = 123;
  final int countPages / countLines = 123;

  othercode;
}
```
I don't know yet how it will land in Markor in the end, for now just want to add it as a experiment. Thus no settings and no string ressources etc required. (i.e. hardcoded strings). (Else translators ask what this is about and where they can find it.)

So pretty much, the easiest implementation without cross referencing would be the best. At same time guess this will also be the easiest way for you to get into the project as you don't need to care about most things.

gsantner:
@UtnansN 
Thanks!

Please make a pull request as soon as you start, that makes it easier to detect bigger issues and you don't do certain things multiple times :).

gsantner:
Open again, no changes yes :smile:

PR #729, opened 2019-10-08
[WIP] Auto scrolling edit/view (feature #699)

UtnansN:
For issue #699

### DIFF
--- a/app/src/main/java/net/gsantner/opoc/util/CoolExperimentalStuff.java
+++ b/app/src/main/java/net/gsantner/opoc/util/CoolExperimentalStuff.java
@@ -144,8 +144,9 @@ public class CoolExperimentalStuff {
         }
     }
 
-    public static void startAutoScroller(final Activity activity, final boolean isWebView) {
+    private static Thread autoScrollerThread = null;
 
+    public static void startAutoScroller(final Activity activity, final boolean isWebView) {
         View view;
         if (isWebView) {
             view = activity.findViewById(R.id.document__fragment_view_webview);
@@ -156,7 +157,7 @@ public class CoolExperimentalStuff {
         final int scrollAmount = view.getHeight() / 3;
         final int millisecondsDelay = 5000;
 
-        Thread s = new Thread() {
+        autoScrollerThread = new Thread() {
             @Override
             public void run() {
                 try {
@@ -168,9 +169,18 @@ public class CoolExperimentalStuff {
                 }
             }
         };
-        s.start();
+        autoScrollerThread.start();
+    }
 
-        // TODO: Add some listener to cancel the scrolling
+    public static void stopAutoScrollerIfExists() {
+        if (existsAutoScroller()) {
+            autoScrollerThread.interrupt();
+        }
+        autoScrollerThread = null;
+    }
+
+    public static boolean existsAutoScroller() {
+        return autoScrollerThread != null && autoScrollerThread.isAlive();
     }
 
 }
--- a/app/src/main/java/net/gsantner/opoc/util/CoolExperimentalStuff.java
+++ b/app/src/main/java/net/gsantner/opoc/util/CoolExperimentalStuff.java
@@ -18,8 +18,6 @@ import com.github.mertakdut.Reader;
 import com.github.mertakdut.exception.OutOfPagesException;
 import com.github.mertakdut.exception.ReadingException;
 
-import net.gsantner.markor.R;
-
 import java.io.File;
 import java.util.Locale;
 import java.util.concurrent.atomic.AtomicReference;
@@ -27,6 +25,8 @@ import java.util.concurrent.atomic.AtomicReference;
 // License Public domain/CC0 for now
 // Just some experimental code
 public class CoolExperimentalStuff {
+
+    private static Thread autoScrollerThread = null;
 
     public static String convertEpubToText(final File filepath, final String translatedStringForPage) {
         //final String filepath = new File("/sdcard/epub.epub");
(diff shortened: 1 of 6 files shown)


gsantner:
Thanks!
Just let me know when it's ready for having a look into changes :smile:

gsantner:
Oh, don't mind the comments by bot for now. You can when you getting near finish :smiley: .

UtnansN:
@gsantner Hi again, I've submitted some commits to the PR. The basic auto-scrolling functionality is there, but I am not quite sure how to figure out the line count for the scroll, so for now it scrolls for 1/3rd of the fragment height every 5 seconds. Auto scroll can be cancelled with a repeated press of the auto scroll button or by switching between edit and view modes, as well as backing out to the folder selection activity.

I am not convinced my implementation is optimal, so feedback would be greatly appreciated. Please look into my changes whenever you have the time.

app/src/main/java/net/gsantner/opoc/util/CoolExperimentalStuff.java
```diff
@@ -141,4 +143,44 @@ public void run() {
             w.setLayout(WindowManager.LayoutParams.MATCH_PARENT, WindowManager.LayoutParams.MATCH_PARENT);
         }
     }
+
+    private static Thread autoScrollerThread = null;
+
+    public static void startAutoScroller(final Activity activity, final boolean isWebView) {
```
gsantner: Change to View, then check 

`instanceof EfitText` and `WebView`.

Pass from fragment hlEditor or webview (depending on current shown). 

This way, you dont need to lookup anything, and dont need the boolean argument.

app/src/main/java/net/gsantner/opoc/util/CoolExperimentalStuff.java
```diff
@@ -141,4 +143,44 @@ public void run() {
...
+        final int scrollAmount = view.getHeight() / 3;
+        final int millisecondsDelay = 5000;
+
+        autoScrollerThread = new Thread() {
+            @Override
+            public void run() {
+                try {
+                    while (view.getVisibility() == View.VISIBLE) {
+                        Thread.sleep(millisecondsDelay);
+                        view.scrollBy(0, scrollAmount);
```
gsantner: doeys that workgood and same for edittrxt and webview?

(i dont kmow, question)

app/src/main/java/net/gsantner/opoc/util/CoolExperimentalStuff.java
```diff
@@ -141,4 +143,44 @@ public void run() {
...
+                    }
+                } catch (Exception ignored) {
+                }
+            }
+        };
+        autoScrollerThread.start();
+    }
+
+    public static void stopAutoScrollerIfExists() {
```
gsantner: I think it would be fine to set a static bool flag to false, and at the loop check for that flag.
Then its nearly impossible to run into exception, thus less error prone.

app/src/main/java/net/gsantner/opoc/util/CoolExperimentalStuff.java
```diff
@@ -141,4 +143,44 @@ public void run() {
...
+    private static Thread autoScrollerThread = null;
+
+    public static void startAutoScroller(final Activity activity, final boolean isWebView) {
+        View view;
+        if (isWebView) {
+            view = activity.findViewById(R.id.document__fragment_view_webview);
+        } else {
+            view = activity.findViewById(R.id.document__fragment__edit__content_editor__scrolling_parent);
+        }
+
```
gsantner: For the logic, I dont know whats the best for average user.
Have some tries what you think fits.

So the idea is, slow & fast enough that you can finish reading the upper half screen within time.
Thats my thought on that, some things may be bigger or smaller which moves from the 50% but thats ok.

So stuff you have also is editor & viewer font size. You can include that in calculation somehow if you think it helps.

app/src/main/java/net/gsantner/opoc/util/CoolExperimentalStuff.java
UtnansN: Oh I believe the edit view never goes to another state than View.VISIBLE, this check only works on webview, as the edit view never switches to View.GONE. I think this while loop can be converted into a while (true), as the thread is set to be always killed when setDocumentViewVisibility is invoked.

UtnansN: I could convert it to kill the thread when it scrolls to the bottom, too

gsantner:
(I am posting this to all pull requests)

Hey! Just want to know how you are doing. Do you have issues or have questions? If you wont work on this anymore also let me know please, so I can announce to others its free to take again.
Thanks!

UtnansN:
Hi, I have been unable to work on this issue for the past couple of days, so feel free to reassign this issue to someone else as I am currently blocking this issue from being completed for Hacktoberfest. I can remove my pull request if you wish.

gsantner:
@UtnansN 
It doesn't matter how fast it's finished, doesn't have to be during hacktoberfest. I just want to know if you still plan to work on this, or you are like "nah, don't want to waste time here" :D .

UtnansN:
@gsantner In that case I am willing to continue work on this, I'll try and figure out the correct way to scroll by line count the next week and then do some refactoring and it should be good to go. :D

UtnansN:
Hi, just wanted to let you know that I won't be able to work on this issue in the near future, so feel free to remove my assignment for it. Sorry for taking the assignee spot for so long because I was uncertain if I could work on it anytime soon :P

Issue #740, opened 2019-10-16
Also hide bar when textactions are disabled


ohenepee:
#### General information
Textactions bar still shows without the buttons when you hide it in the settings, thus still stealing screen real estate for apparently no purpose.

* **App version:**  2.1.5 (97)
* **System:** Android 7.1

#### Description
I guess the obvious expectation is a hidden textactions along with the empty bar. No need to keep the bar still visible.

gsantner:
with the move of "Representation Page" to the new "View Overlay Mode" the hide functionality was not added.

This is a very easy change, one line of code required. Leaving this to hacktoberfest community.

Irfan-S:
Hi, could you specify the exact steps required to reproduce the issue? I'm unable to trace the exact cause of the bug.

gsantner:
Settings - General - TextActions. Disablee that and you see a empy bar in editor. that instead should be hidden and editor take its place too.

WendlerTech:
@gsantner mentioned that the functionality hadn't been added yet, but the code seemed to show that it was there. I spent _way_ too much time figuring out why it wasn't executing as expected; turns out someone just added a bottom margin to the document fragment.

gsantner:
Yeah guess then its two lines, setMargin  and Setvisibility

PR #742, opened 2019-10-17
Hide bar when textactions are disabled.

WendlerTech:
Solves #740

### DIFF
--- a/app/src/main/res/layout/document__fragment__edit.xml
+++ b/app/src/main/res/layout/document__fragment__edit.xml
@@ -24,7 +24,6 @@
         android:layout_width="match_parent"
         android:background="@android:color/transparent"
         android:layout_height="match_parent"
-        android:layout_marginBottom="@dimen/textactions_bar_height"
         android:fillViewport="true">
 
         <net.gsantner.markor.ui.hleditor.HighlightingEditor


gsantner:
Sorry but thats not the solution.

This way text is below the textactions bar.

Dont remove the margin in xml. But set margin in dp in code for editor, plus setVisibility on textactions.

WendlerTech:
Per your edit, @gsantner, I see you found that the setVisibility was already in place. As I said earlier, I spent a good chunk of time going through all the code & I was unable to find what exactly the problem was. This is most likely due to lack of experience in general &/or lack of experience with this (fairly large) codebase, since I feel like I'm missing something obvious. 

From what I found, the setBarVisible method in TextActions is called appropriately to hide the parent, followed by the DocumentEditFragment class calling applyTextFormat to remove the child. Both are using View.GONE but are behaving as if they're using View.INVISIBLE. 

Also may be worth mentioning that prior to the margin fix, while viewing a saved note (from the Files tab, not QuickNote) & having the bar hidden, the area that the bar takes up would actually duplicate the text underneath it, causing the duplicated text to follow along as a user scrolled up & down until they pulled the keyboard back up. This no longer happens after I removed the margin.

gsantner:
can you find a case where text from editor is below the bar?

WendlerTech:
Prior to removing the margin? Yeah, I just create a new note from the Files tab, using the default settings. Add a lot of text (spanning multiple lines, enough to make the user need to scroll & reaching below where the keyboard comes up to), hide the keyboard, then click within the note to re-activate the keyboard. The bar will now have some "ghosting" from the characters that were underneath it.

gsantner:
Weird :D.
And now its all good?

WendlerTech:
Yep, just by removing this line in the parent ScrollView:

`android:layout_marginBottom="@dimen/textactions_bar_height"`

Everything works. The bar hides appropriately, there's no ghosting, & due to always being able to scroll down, the text is never covered/made inaccessible due to lack of a margin.

gsantner:
OK thanks for your findings and fixes! :)

WendlerTech:
Happy to help! I definitely felt a bit lazy committing only 1 deleted line, but at least I got to explain all the work that went into tracking that 1 line down!

gsantner:
Hehe, its early in morning and I'm only on phone, no computer :D. Didnt have look into codebase, just from what I remember.

gsantner:
@WendlerTech have you expierenced similar like #782  ?

WendlerTech:
@gsantner Hmm, yeah, that does happen. I didn't catch it because in my testing I was only adding a few characters & clicking enter, rather than typing all the way through to the next line. Seems like it only happens when the textactions bar is visible.

Prior to the fix, not only did it leave that awkward space, but it was causing text to duplicate when scrolling, so I still think removing it was the right move. It just needs to be re-added when the bar is showing.

I feel like the ideal fix would be to have a flag that, when the textactions bar is set to hidden, would remove the bottom margin & conversely would re-add it when the bar was set to shown again. However, I don't know of an easy way to edit the xml via code, so we may have to come up with something else.

PR #753, opened 2019-10-28
A11y improvements

### LINKED ISSUE #613: Make Markor more accessible (accessibility)
#### General information

* **App version:** 2.0.1 F-Droid
* **System:** Android 8.1, Moto G4 Plus

#### Description

Currently the buttons for text editing and other controls are not accessible to users of assistive technology, e.g. Android Accessibility Suite (TalkBack). Even if announcement of button ID's is turned on in TalkBack settings, most buttons are still recognized as unlabeled, making them quite useless. This could be improvid by adding alternative texts for blind users.

pvagner:
Add content descriptions to unlabeled buttons (fixes #613 )


gsantner:
Thanks! I'm adding some comments

app/src/main/res/values/strings.xml
```diff
@@ -336,4 +336,35 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="table">Table</string>
     <string name="how_much_columns_press_table_button_long_to_start_table">How much columns? Press table button long to start a table (insert header), press short to add a row.</string>
     <string name="example_of_a_markdown_table">Example of a Markdown table</string>
+    <string name="common_delete_lines">Delete lines</string>
+    <string name="common_attach_something">Add attachment</string>
+    <string name="common_unordered_list_hyphen">Unordered list</string>
+    <string name="common_ordered_list_number">Ordered list</string>
+    <string name="common_checkbox_list">Check list</string>
+    <string name="common_special_key">Special key</string>
```
gsantner: Please reuse strings where possible, in constrast to the `tmaid` stuff this is really about text and not about identifiers. 

"Special key" for example is already defined. In general, use here the same text for name as the value is, just with underscore_and_lower_case.

For Markdown for example you can use "Markdown Heading 1" and have markdown_heading_1. 

For all todo.txt stuff, no todo.txt in title nor value is required. (We might get HTML heading 1 in future, but for sure there is no "priority" in HTML, so no need to differntiate)

pvagner: Thank you for the review.
I'll go through it and edit all the strings.
You know I have just copied and pasted the identifiers.

gsantner:
@pvagner 
Thanks much!

do you still have something in the queue, uncommited or alike? or are you finished with what you wanted to do?

pvagner:
@gsantner I'm now looking what to tweak in order to add a content description for the back button found in the settings activity.
It's not obvious to me. So if you can, please give me a hint otherwise I'm determined to find it out and I'll be looking into it these days.
Other than that I can't find other accessibility issues right now.
If I discover something else I can create follow up PR next time.

gsantner:
That one on the top left edge to leave the settings?

pvagner:
Hmm. So perhaps the correct way is to alter its style or something.
Other apps do usually have this button labelled.

Issue #747, opened 2019-10-23
Feature: Search in View


532910:
#### General information

* **App version:** 2.1.5 97
* **System:** 16.0-20191017-microG-opnepus3

#### Description

    "Hey! Theres no search for view available.
    I only found a button in view which does go to editing instead.
    Can you please add search for View mode? Thanks."

Anca07:
I would like to work on this issue :)

gsantner:
@Anca07 
Great, thanks! Sure, go for it!

Have you used the app before, and in general understand what this is about? I'm sure @532910 will give you details when needed :smile: .

(From code pov: it's about webview and it's find/findNext methods)

Anca07:
Hi @gsantner 

I played around with the app today, but since I liked it I think I will start using it from now on :smiley: 
If I understood correctly, the issue was that in view mode there is no search functionality (please correct me if I am wrong)
I have implemented a POC for search in view mode functionality (screenshots provided below).
Let me know if this is what is expected or if you have any suggestions.

![76730709_2178828135751202_625868385334329344_n](https://user-images.githubusercontent.com/18334788/67636113-d82c1180-f8d5-11e9-9bcb-5b10931d0114.jpg)
![73167841_449987422309431_3558630565106155520_n](https://user-images.githubusercontent.com/18334788/67636020-b8e0b480-f8d4-11e9-88aa-026d28196429.jpg)
![73019442_2613809975325052_4403713621666824192_n](https://user-images.githubusercontent.com/18334788/67636021-bda56880-f8d4-11e9-9d7f-3142e7485095.jpg)

gsantner:
In browser you usually have a option to search current text. 
In the View mode there should be used the web search (webview.find(text)) method to search text instead of the editor text

Anca07:
Okay, I got it now. Sorry for the misunderstanding.
I will make the changes accordingly.

Anca07:
Hello again,

I have changed the search functionality according to your indications. 
Searching a string will highlight all the matches and  by pressing the arrow button the next occurrence will be selected.
Let me know if this is what was expected.

![74163828_483273262265845_3213825754450100224_n](https://user-images.githubusercontent.com/18334788/67639256-58626f00-f8f6-11e9-85aa-b4d8e805e685.jpg)
[...]

PR #754, opened 2019-10-28
Add search in view mode

Anca07:
Solves #747 
- add searchview component
- implement search functionality in webview

Requested screenshots with light theme and markdown template file provided below:

73712538_1230374187146998_845365901631946752_n
72843522_534226883790108_7242509519094284288_n


gsantner:
Where does that button come from? 
grafik

So the Enter button as well the arrow jumps to next result?

Anca07:
The arrow button is a component of the SearchView, activated by doing this:

`searchView.setSubmitButtonEnabled(true);`

And yes, they both have the same functionality of jumping to the next result.

gsantner:
OK!

Can you please also add onBackButtonPressed -> close search?

gsantner:
Am I wrong or did you have the same also for EditText? How did it work there?
Potentially we can use this search also in edittext

Anca07:
If you refer to the search in edit mode, I did not change that. 
I only implemented this type of search for view mode.

PR #785, opened 2019-11-15
Use correct plurals in delete confirmation dialog

mymedia2:
It fixes plural form in the delete confirmation dialog: _these_. It also makes possible to get consistent translation here.

Please avoid joining of UI strings from small pieces like the case was here, as it makes harder to localize them.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -16,7 +16,10 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="appearance">Appearance</string>
     <string name="rename">Rename</string>
     <string name="confirm_delete">Confirm Delete</string>
-    <string name="do_you_really_want_to_delete_this_witharg">Do you really want to delete this %s?</string>
+    <plurals name="do_you_really_want_to_delete_this_document">
+        <item quantity="one">Do you really want to delete this document?</item>
+        <item quantity="other">Do you really want to delete these documents?</item>
+    </plurals>
     <string name="confirm_overwrite">Confirm Overwrite</string>
     <string name="file_already_exists_overwerite">File already exists, overwrite?</string>
     <string name="browse_filesystem">Browse filesystem</string>
--- a/app/src/main/java/net/gsantner/opoc/ui/FilesystemViewerFragment.java
+++ b/app/src/main/java/net/gsantner/opoc/ui/FilesystemViewerFragment.java
@@ -560,7 +560,7 @@ public class FilesystemViewerFragment extends GsFragmentBase
     ///////////////
     public void askForDeletingFilesRecursive(WrConfirmDialog.ConfirmDialogCallback confirmCallback) {
         final ArrayList<File> itemsToDelete = new ArrayList<>(_filesystemViewerAdapter.getCurrentSelection());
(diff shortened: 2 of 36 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -16,7 +16,10 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="appearance">Appearance</string>
     <string name="rename">Rename</string>
     <string name="confirm_delete">Confirm Delete</string>
-    <string name="do_you_really_want_to_delete_this_witharg">Do you really want to delete this %s?</string>
+    <plurals name="do_you_really_want_to_delete_this_document">
+        <item quantity="one">Do you really want to delete this document?</item>
```
gsantner: please have for both a separate string ID in strings.xml

gsantner: please have for both a separate string ID in strings.xml

app/src/main/res/values/strings.xml
```diff
@@ -16,7 +16,10 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="appearance">Appearance</string>
     <string name="rename">Rename</string>
     <string name="confirm_delete">Confirm Delete</string>
-    <string name="do_you_really_want_to_delete_this_witharg">Do you really want to delete this %s?</string>
+    <plurals name="do_you_really_want_to_delete_this_document">
```
gsantner: please move plurals to arrays.xml and refer both strings by string ID. See others.

mymedia2: This is wrong and causes troubles with translation into languages other than English where rules for varying numerals are different. Please do not do that anymore and fix remaining usage places of plural strings.

gsantner: You didn't get what I mean I guess.

```
    <plurals name="documents" translatable="false">
        <item quantity="one" translatable="false">@string/document</item>
        <item quantity="other" translatable="false">@string/documents</item>
    </plurals>
```

app/src/main/res/values/strings.xml
mymedia2: this indivisible string resource should have the single ID

gsantner: I don't want anything but key=value in strings.xml. Other cases have proofed not to work good with translation tools & services.

app/src/main/res/values/strings.xml
mymedia2: You mark this resource as untranslatable and provide only two options, "one" and "other". That's not enough.

gsantner: Provide as many as you think needed. But use the said architecture. Thats the point.

app/src/main/res/values/strings.xml
mymedia2: Ok, I can put the resource to other file, but `translatable="false"` shouldn't apply to it. Crowdin supports plurals resources.

gsantner: But others don't. I want to depend neither github, crowdin nor both.

gsantner:
I've decided against this. I will reword that to just say along "delete following file(s)", keeping it simple.

PR #834, opened 2020-01-21
Set tooltip text for text actions.

### LINKED ISSUE #829: tooltips for editor tools
#### General information

* **App version:** v2.2.5 (107)
* **System:** Android 9

#### Feature request

For those who use this app for the first time, or those who forget (like me :wink:), it could be useful to have tooltips for the tool icons which are located at the bottom of the editor: 

In my opinion, some icons are not self-evident and would be hard or impossible to improve. And some look like I would rather not try them in a productive file (like the [`delete`](https://material.io/resources/icons/?icon=delete) or the [`open_in_browser`](https://material.io/resources/icons/?icon=open_in_browser) icons).

I suggest to add **[tooltips](https://material.io/components/tooltips/#placement)** to all tool icons. The tooltips would be [displayed through long press](https://storage.googleapis.com/spec-host-backup/mio-components%2Fassets%2F0B8wSqcLwbhFuZVFJeXRLd3hlM0E%2Fbaseline-video-list.mp4):

dhebbeker:
Justification
=========
Tooltips for buttons can help using the application. Especially if the buttons have no label, but only an icon. Tooltips for text actions where not shown until now for non-desktop environments (no hover available).

This pull request fixes #829.

Changes
=======
At the place where the buttons are synthesized (`TextActions.appendTextActionToBar()`), the tooltip text is set via `TooltipCompat.setTooltipText()`. The [TooltipCompat](https://developer.android.com/reference/androidx/appcompat/widget/TooltipCompat.html) from the [AndroidX library](https://developer.android.com/reference#androidx) is used to emulate the behaviour of [`View.setTooltipText(CharSequence)`](https://developer.android.com/reference/android/view/View.html#setTooltipText(java.lang.CharSequence)) prior to API level 26. The API is currently set to minimum of 16.

For the text of the tooltip, the same resource is used as for the button content description.

As the long press listener is set by default, the [return value of the listener](https://developer.android.com/reference/android/view/View.OnLongClickListener.html#onLongClick(android.view.View)) is used by the `View` to determine whether the tooltip can be shown when a long press occured or not. Thus the lambda function has been changed in order to pass the return value to the wrapping listener.

Checklist
=======
- [x] Justify changes in PR
- [x] Describe changes in PR
- [x] Auto-Format
- [ ] Tests in CI: **How?** I could not trigger the [CI](https://app.codacy.com/manual/gsantner/markor/pullRequest?prid=4859688#) to analyse the proceeding commits.
- [x] Test manually
- [x] Request review
- [x] Add myself to [`CONTRIBUTORS.md`](https://github.com/gsantner/markor/blob/master/CONTRIBUTORS.md)

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -40,3 +40,4 @@ Where:
 * **[Alexander Sachse](alexander.sachse@gmail.com)**<br/>~° Added simple word counter
 * **[Kyle Phelps](kyle.phelps757@gmail.com)**<br/>~° Added functionality to chare multiple files from file browser
 * **[Vladislav Glinsky](https://github.com/cl0ne)**<br/>~° Ukrainian/Russian translations
+* **[David Hebbeker](https://david.hebbeker.info/)**<br/>~° Added tooltips for text actions.
--- a/app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
+++ b/app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
@@ -14,6 +14,7 @@ import android.content.Context;
 import android.support.annotation.DrawableRes;
 import android.support.annotation.StringRes;
 import android.support.v4.content.ContextCompat;
+import android.support.v7.widget.TooltipCompat;
 import android.text.Editable;
 import android.text.TextUtils;
 import android.view.View;
@@ -73,6 +74,7 @@ public abstract class TextActions {
         ImageView btn = (ImageView) _activity.getLayoutInflater().inflate(R.layout.quick_keyboard_button, null);
         btn.setImageResource(iconRes);
         btn.setContentDescription(_activity.getString(descRes));
+        TooltipCompat.setTooltipText(btn, _activity.getString(descRes));
         btn.setOnClickListener(v -> {
             try {
                 listener.onClick(v);
--- a/app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
+++ b/app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
@@ -85,11 +85,11 @@ public abstract class TextActions {
         if (longClickListener != null) {
             btn.setOnLongClickListener(v -> {
                 try {
-                    longClickListener.onLongClick(v);
+                    return longClickListener.onLongClick(v);
                 } catch (Exception ex) {
                     ex.printStackTrace();
                 }
-                return true;
+                return false;
             });
         }
         btn.setPadding(_textActionSidePadding, btn.getPaddingTop(), _textActionSidePadding, btn.getPaddingBottom());


dhebbeker:
A problem is, that all calls of `appendTextActionToBar` have the argument `longClickListener != null`. Even though not every object does have a corresponding action for the `onLongClick` method. Thus it seams unnecessary? :roll_eyes:

Maybe this requires a little bit of refactoring? @gsantner, what is your verdict?

gsantner:
Some actions have a additional longpress action (want to keep that, use that everyday :D).
Actually most stuff has no long click action.

Easiest should be to set the tooltip later instead of before adding. and checking if set one is null

dhebbeker:
On Tue, 2020-01-21 at 09:33 -0800, Gregor Santner wrote:
> Some actions have a additional longpress action (want to keep that, use that everyday :D).

Of course! Any idea how provide help to the user for these specific
actions?

> Easiest should be to set the tooltip later instead of before adding. and checking if set one is null

All objects passed to `appendTextActionToBar` are `!=null`. Actually
the same objects are passed for normal and long click arguments. The
information of which action has long press is currently coded in the
respective classes `MarkdownTextActionsImpl`, `PlaintextTextActionImpl`
and `TodoTxtTextActionsImpl`.

dhebbeker:
The difficulty at first glance is, that in the function `TextActions:.appendTextActionToBar()` assigns a `longClickListener` to every button. This is because all calls of this function are with the argument for `longClickListener` being not `null`. Thus, the `longClickListener` will always be run instead of showing the tooltip (in environments, where no "hover" is possible). This results in now tooltips are shown, even if in the run, no action is performed.

Actually I have the following approaches in mind.

### Idea A

Do not assign a `longClickListener` to buttons where not action is performed. Currently the decision on whether an action is performed or not is done at runtime based on the action id or action string by the `longClickListener` object itself. Thus one would have to do the decision beforehand. 

For example the classes for the `longClickListener` objects could be dissolved into functions which do return lambda functions for the appropriate action id. In case no action is designated, `null` would be returned. That object would then be passed to `TextActions:.appendTextActionToBar()`. 

 *  Advantages: The selection of the correct action does not need to be done every time a button is pressed any more.
 * Disadvantages: This would require refactoring parts of about 4 files affecting about 200-300 lines of code.

### Idea B

The `onLongClick` methods already report with their return value if they have *handled* the click. This value is actually evaluated by the button's `performLongClick()` method. The return value just has to be passed further on: From the embedded long click listener result to the wrapper return.

---

As it is less intrusive, I will go with idea B for now.

gsantner:
@dhebbeker 
Thanks, I'm merging this.

Btw, is this some kind of pull request template you use? seems pretty high level^^

dhebbeker:
Thank you!

No, I did not use a template in this particular pull-request. I just value the documentation for the maintainers, others and for myself. So I wrote down what I thought would be necessary to understand the changes.

Have a look at this template of mine: https://github.com/dhebbeker/memorex-android/blob/master/.github/PULL_REQUEST_TEMPLATE.md

gsantner:
Thanks very much, thats like the first time I got such a nice PR :D.

Is the template reusable as in public domain?

dhebbeker:
Yes, please go ahead and use it as you like.

PR #862, opened 2020-02-22
Configurable unordered list character

### LINKED ISSUE #853: Customisable unordered list character
#### General information

* **App version:** v2.2.6 (108)
* **System:** Android 10

#### Description

It would be great to be able to configure the character used for unordered lists, with the options being at least `*` and `-`. Use case: most of my notes predate my use of Markor and I’d like to keep them in a consistent format, and I’m sure many users edit their notes in multiple systems that have different Markdown styles.

If this is something you’d be happy to see in the app, I’d be happy to work on a PR.

#### Log

N/A

markormesher:
Fixes #853.

I tried to keep the blast radius of this change as small as possible and put new strings close to similar-purpose existing strings.


app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextActions.java
```diff
@@ -100,8 +100,8 @@ public void onClick(View view) {
                     runMarkdownRegularPrefixAction("### ");
                     break;
                 }
-                /*case R.string.tmaid_common_unordered_list_hyphen: {
```
markormesher: I updated the code here, even though it's commented, to keep things consistent if this is ever uncommented.

markormesher:
Should this be applied to the insertion of `- [ ]` for [checkbox lists](https://github.com/gsantner/markor/pull/862/files#diff-8d37f70fd1bd3bdf1e3d6a8199b82e5fR308) as well? I'd initially think "yes" but I wouldn't want to break existing lists.

gsantner:
I'm not sure if it works with other implementations. Can only remember to have seen hyphen check

markormesher:
Probably best to leave it as it is for now then

app/src/main/res/values/string-not_translatable.xml
```diff
@@ -54,6 +54,9 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="editor_basic_color_scheme_gruvbox" translatable="false">Gruvbox</string>
     <string name="editor_basic_color_scheme_greenscale" translatable="false">Greenscale</string>
     <string name="editor_basic_color_scheme_sepia" translatable="false">Sepia</string>
+    <string name="unordered_list_char_dash">Dash (-)</string>
+    <string name="unordered_list_char_asterisk">Asterisk (*)</string>
+    <string name="unordered_list_char_plus">Plus (+)</string>
```
gsantner: These 3 are in not_translatable, but they are translatable right? -> move it to normal xml

app/src/main/res/values/strings.xml
```diff
@@ -191,6 +191,8 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="key_tab">Tabulator</string>
     <string name="start_on_bottom">Start on bottom</string>
     <string name="start_on_bottom_when_loading_document">Upon loading a document, position cursor at its end</string>
+    <string name="unordered_list_character">Unordered list character</string>
+    <string name="unordered_list_character_summary">Which character to use to denote unordered lists</string>
```
gsantner:
```
     <string name="character_to_start_items_of_unordered_list_with">Character to start items of unordered lists with</string>
```

app/src/main/res/values/string-not_translatable.xml
markormesher: Thanks - I've move them.

app/src/main/res/values/strings.xml
markormesher: Thanks - renamed.

markormesher:
Thanks for the review @gsantner. I've made the changes you suggested and merged in the latest changes on master.

gsantner:
@markormesher 
Thanks for helping, I'm merging this.

You are very welcome to help on other issues as well, I'm basically working 99% alone on this project - which I'm happy if changes.

Issue #794, opened 2019-12-06
Markdown: Add support for nested check lists


ascii78:
#### General information

* **App version:**  2.2.3
* **System:** android 9

#### Description

Just synced succesfully with my Typora markdown notes. I have some task lists that are indented as
follows:

```
- [ ] main
   - [ ] indented
```

I can tick the main box with the provided tick button, but not the indented one. It does not recognize that it is a task entry and will add another `- [ ]` to the beginning of the line.

harshad1:
This happens with unordered lists too.
```
- Foo
    - Bar
```
> Place cursor on line 'Bar' and press list button
```
- Foo
-      - Bar
```

westurner:
List items nested beyond 3 levels are also being highlighted? Is that a separate bug?

```markdown
- one
  - two
    - three #highlighted
      - four #hl
         - five #hl
```

westurner:
Is it a regex for markdown task lists?
A regex match group for `(\s*)` would likely solve?

gsantner:
As you see in other #s , starting with amount of space will make code highoithting. thats gonna stay.

westurner:
If someone can fix the thus broken markdown implementation, would you be open to such a patch?

https://spec.commonmark.org/0.29/#list-items

https://spec.commonmark.org/0.29/#example-264 :

```markdown
- foo
  - bar
    - baz
      - boo
```

https://spec.commonmark.org/0.29/#example-266 :

> Here we need four, because the list marker is wider:

```markdown
10) foo
    - bar
```

gsantner:
I dont feel like its too much broken, but when you can provide a PR that not negatively impacts performance its welcome.

harshad1:
Here is my attempt at fixing this issue :) https://github.com/gsantner/markor/pull/877

harshad1:
@ascii78 Would you be willing to review / test my attempt at fixing this issue? I have a PR (see comment above).

ascii78:
sorry, not a developer so i don't now how i would a review a change that is not in a release, maybe someone else can help ?

PR #877, opened 2020-03-23
WIP: Ignore leading whitespace for list types

harshad1:
This PR addresses [Markdown: Add support for nested check lists](https://github.com/gsantner/markor/issues/794).

### DIFF
--- a/app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
+++ b/app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
@@ -107,11 +107,21 @@ public abstract class TextActions {
         }
     }
 
-    protected int findLineStart(int cursor, String text) {
+    protected int findLineStart(int cursor, String text, Boolean ignoreLeadingWhitespace) {
         int i = cursor - 1;
+
         for (; i >= 0; i--) {
             if (text.charAt(i) == '\n') {
                 break;
+            }
+        }
+
+        if (ignoreLeadingWhitespace) {
+            for (; i < (cursor - 1); ++i) {
+                char c = text.charAt(i + 1);
+                if (c != ' ' && c != '\t') {
+                    break;
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
```diff
@@ -107,14 +107,24 @@ protected void setBarVisible(ViewGroup barLayout, boolean visible) {
...
         }
 
+        if (ignoreLeadingWhitespace) {
+            for (; i < (cursor - 1); ++i) {
+                char c = text.charAt(i + 1);
+                if (c != ' ' && c != '\t') {
+                    break;
+                }
+            }
+        }
```
harshad1: Linear search forward from newline. Should not cause noticeable performance hit.

harshad1:
@gsantner What would it take to get this reviewed and merged?

gsantner:
How do you practically differenciate between if its whitespace for CodeBlock, OR list item?

westurner:
I haven't reviewed the code in any depth (and haven't written Java in many years), but could you track the state of the previous line with e.g. `int currentListIndent`?

IIUC, codeblocks can be included in nested lists?

```markdown

currentListIndent=0

    # Code block
    import java.util

- 1 (currentListIndent=2)
  - A (currentListIndent=4)

        # Code block
        import java.util
```

But if you want to indicate the code language, it's better to use a fenced code block:

```markdown
- 1 (currentListIndent=2)
  - A (currentListIndent=4)

    ```java
    # Code block
    import java.util
    ````
```

harshad1:
> How do you practically differenciate between if its whitespace for CodeBlock, OR list item?

I do not (and I do not know how to). My approach here is to simply ignore any whitespace between the (previous) newline and the first non-whitespace character. With this change, if you were to tap the list button (say) with the cursor on an indented code block, it would add a leading bullet at the same indent as the code block. To me this is the least-surprising behavior, but I am open to criticism and suggestions.

gsantner:
Please just do some tries and verify it works for your usecase and that one from issue(s).

I will merge it when you tell me it's ready.

harshad1:
> Please just do some tries and verify it works for your usecase and that one from issue(s).
> 
> I will merge it when you tell me it's ready.

Ok. I am doing some more testing and am trying to get the reporter from that issue to do some testing too. 

I will let you know when the PR is ready. This may take a day or two.

It may be useful if you create 'work and progress' and 'ready to merge' labels in github

gsantner:
I guess most of reporting people dont know about code and stuff. Should be fine when you confirm.

PR #923, opened 2020-05-09
Fix readme AES256 typo

tlvince:
Thanks for this feature!


gsantner:
Can you please check readme and strings otherwise too, let me know if you have some other ideas for improvement

PR #926, opened 2020-05-11
Save and remember done file

harshad1:
In this PR I have updated the 'Move done tasks' option so that 

a. The last used done file is actually remembered
b. The last used done file is auto-populated


harshad1:
@gsantner please let me know if you have any opinions about this PR

Issue #942, opened 2020-05-23
Special keys menu: Add list option for "cursor select whole line" 


aisbergde:
#### General information

* **App version:** 
* **System:** 

#### Description

it is hard to mark a whole line (which is a task), for example to move the line from one file to another. Please add a new keyboard action to mark a whole line.

gsantner:
the first button (checkbox) is exactly for this

aisbergde:
no, the first button will add or remove the "x" to mark or unmark the task as comleted.

what I am asking for is a quick way to mark the text of one line as selected text so I can use the context menu on this selected text to copy or cut.

the request behind is to find a way how to move a task (line) between different files, but not only to the done file.

gsantner:
As in keyboard, hardware key-combination? Than <https://github.com/gsantner/markor/issues/154> duplicate

gsantner:
Thanks now I know where you mean. Is quite easy to add, so when somebody reads this: for functionality it's two invocations, plus if+string.

gsantner:
thanks @harshad1

PR #943, opened 2020-05-24
Added special key to select current line

harshad1:
This PR addresses https://github.com/gsantner/markor/issues/942.

I have added a key to select the current line(s). 

All lines which of which the selection is a part will be completely selected.


gsantner:
Hey thanks for fast :D

I've added a new function to hleditor, in the codeblock commit.
`_hlEditor.setSelectionExpandWholeLines()` should be all that is needed :D

gsantner:
I didnt have a look at other textactions when added it these days... I think we can get rid of some code as theres similar code on some others...have to check myself what all it is :D

harshad1:
@gsantner Thanks for pointing out that there is a single function now.

PR #964, opened 2020-05-30
Due and threshold date handling using calendar 

harshad1:
In this PR, I want to add due and threshold date handling using a calendar widget.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
+++ b/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
@@ -11,7 +11,10 @@ package net.gsantner.markor.format.todotxt;
 
 import android.app.Activity;
 import android.app.DatePickerDialog;
+import android.content.Context;
 import android.os.Bundle;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.support.annotation.StringRes;
 import android.support.v4.app.DialogFragment;
 import android.support.v4.app.FragmentActivity;
@@ -19,6 +22,8 @@ import android.text.Editable;
 import android.text.TextUtils;
 import android.view.HapticFeedbackConstants;
 import android.view.View;
+import android.widget.DatePicker;
+import android.widget.TextView;
 
 import net.gsantner.markor.R;
 import net.gsantner.markor.format.general.CommonTextActions;
@@ -170,7 +175,7 @@ public class TodoTxtTextActions extends TextActions {
                     return;
                 }
                 case R.string.tmaid_todotxt_current_date: {
-                    setKeyDate(origTask, "due", 3);
+                    setKeyDate(origTask, "due", 3, getContext().getString(R.string.set_due_date));
                     return;
                 }
                 case R.string.tmaid_common_delete_lines: {
@@ -304,7 +309,7 @@ public class TodoTxtTextActions extends TextActions {
                     return true;
                 }
                 case R.string.tmaid_todotxt_current_date: {
-                    setKeyDate(origTask, "t", 3);
+                    setKeyDate(origTask, "t", 3, getContext().getString(R.string.set_threshold_date));
                     return true;
                 }
             }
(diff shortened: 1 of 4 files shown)


app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -164,7 +170,21 @@ public void onClick(View view) {
...
                 }
                 case R.string.tmaid_todotxt_current_date: {
-                    _hlEditor.getText().insert(origSelectionStart, SttCommander.getToday());
+                    String dueDateString = origTask.getDueDate();
+                    Calendar dueDate = parseDateString(dueDateString, Calendar.getInstance());
+
+                    DatePickerDialog.OnDateSetListener listener = (_view, year, month, day) -> {
+                        Calendar fmtCal = Calendar.getInstance();
+                        fmtCal.set(year, month, day);
+                        origTask.setDueDate(SttCommander.DATEF_YYYY_MM_DD.format(fmtCal.getTime()));
```
harshad1, line 179: @gsantner this doesn't work `sttcmd.regenerateTaskLine(updatedTask)` does not use the new due date

gsantner: As duedate is part of description (i.e. freetext that comes after creation date) it doesnt know where to put the duedate. Basically due is a "parsed only" value.

In that case its neccassary to do strreplace(description, olddue, newdue) && setDue(newdue) && regenerate

harshad1: I see. I am probably going to switch to using the regexReplaceAction then.

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -164,7 +170,7 @@ public void onClick(View view) {
                     return;
                 }
                 case R.string.tmaid_todotxt_current_date: {
-                    _hlEditor.getText().insert(origSelectionStart, SttCommander.getToday());
+                    setKeyDate(origTask, "due", 3, getContext().getString(R.string.set_due_date));
```
harshad1, line 173: Note that if multiple lines are selected, the first one will be used to get the initial due date, but all will be updated.

harshad1:
@gsantner I can't seem to figure out why the CI / tests aren't working.

gsantner:
Hm I cant see it either it just says not up to project quality.

Can you check other PR comments by me and apply them to this one? Like if braces, this/_

harshad1:
@gsantner This is ready for final review and merging

gsantner:
Hey @harshad1 

I gave it a try:
* I cannot spot t/treshold anywhere (neither single/double press). But anyway, I think I'm fine without. Feels like bloat. Is this something you personally do use and **really** would use? To be honest, I don't feel like it makes any sense to add to markor. Instead user really wants it they can always type `t:` and press date button. 
* Single press date button: Just inserts date. Fine!
* Double press, inserts or replaces due date. Also Fine!

So from what I currently get it lookg all good, thanks!  
However I see some code with `"t` and not sure where/how it is used.

If you are **really really** interested and want to use t, we can add `t:` highlighting similar to `due:`.

app/src/main/res/values/strings.xml
```diff
@@ -127,6 +127,8 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="files_and_folders">Files and folders</string>
     <string name="document_browser">Document browser</string>
     <string name="file_browser">File browser</string>
+    <string name="set_due_date">Set due date</string>
+    <string name="set_threshold_date">Set threshold date</string>
```
gsantner, line 131: t

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -298,7 +304,7 @@ public boolean onLongClick(View v) {
                     return true;
                 }
                 case R.string.tmaid_todotxt_current_date: {
-                    _hlEditor.getText().insert(origSelectionStart, " due:" + SttCommander.getDaysFromToday(3));
+                    setKeyDate(origTask, "t", 3, getContext().getString(R.string.set_threshold_date));
```
gsantner, line 307: t?

harshad1: The t is for `t:yyyy-mm-dd` threshold dates. After thinking about this some more, I feel that we do not need threshold dates.

I have updated the pr to:
1. Short press of date inserts date (if a date is selected, the date is replaced)
2. Long press inserts or updates a due date

harshad1:
@gsantner This should be good to go now

gsantner:
@harshad1 
thanks I'm merging this now

PR #970, opened 2020-06-02
WIP: Open correct file or folder from widgets and shortcuts

### LINKED ISSUE #873: creating a document from the markor widget doesn't create it in the widget folder
#### General information

* **App version:** 2.2.7

#### Description
I configured a desktop markor widget to display a chosen folder (a subfolder of the root markor documents folder). The widget properly displays files from the chosen folder.
When I click the "+" button in the widget, it prompts me for creating a new document, but the new document is created in the root folder of markor documents, not in the folder that I configured and which is displayed by the markor widget.

harshad1:
This PR address two related issues:

1. This PR fixes #873.
2. This PR ensures that shortcuts always open the requested document.


harshad1:
@gsantner this PR is ready for review / feedback.

app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
```diff
@@ -304,21 +332,16 @@ public void onBackPressed() {
...
-                    .replace(R.id.document__placeholder_fragment, fragment, fragment.getFragmentTag())
-                    .commit();
-        } else {
-            fragment = currentTop;
-        }
+        // Replace regardless of current state as document or paths may be different
+        _fragManager.beginTransaction()
+                .replace(R.id.document__placeholder_fragment, fragment, fragment.getTag())
+                .commit();
+
```
harshad1, line 340: Need to come up with a way to not do extra work if current document is same as requested document

app/src/main/AndroidManifest.xml
```diff
@@ -132,7 +132,7 @@
         <activity
             android:name=".activity.DocumentActivity"
             android:label="@string/app_name"
-            android:launchMode="standard"
+            android:launchMode="singleTop"
```
gsantner, line 135: This destorys the "Multiple Windows" feature. This way you can only have one file open a time.

app/src/main/res/xml/prefactions_share_into.xml
```diff
@@ -40,7 +40,6 @@
         android:title="@string/calendar" />
 
     <PreferenceCategory
-        android:key="@string/pref_key__favourite_files"
```
gsantner, line 43: intentional?

harshad1: Yes. Android studio was complaining about there being a duplicate "perf_key__favourite_files". I don't think removing this second one hurts anything.

app/src/main/AndroidManifest.xml
harshad1: Ah. Will fix

harshad1: Actually, could you confirm that making `DocumentActivity` `singleTop` actually prevents multiple windows. I am able to open multiple windows just fine with singleTop enabled.

gsantner:
@harshad1 
I tried it. It's impossible to have two documents open at same time with these changes. Particularly due some crash AND singleTop.

On master branch I can have two files open at same time. Works i.e. that way:

* Open app
* Split-screen
* Open a file (opens on bottom/right-half)
* close split screen
* open another file
* Split-screen .. now you can have two files

video_2020-06-11_15-50-39

app/thirdparty/java/other/writeily/widget/WrMarkorWidgetProvider.java
```diff
@@ -29,6 +29,11 @@
 
 public class WrMarkorWidgetProvider extends AppWidgetProvider {
     public static final String WIDGET_PATH = "WIDGET_PATH";
+    private int requestCode = 0;
+
+    private int nextCode() {
```
gsantner, line 34: how about random code instead? (Random.nextint()). Should be more likely to happen 0 twice (due value forgotten by rotation) than hitting random twice

gsantner:
Checked out branch, and tried same like GIF from last time. That one works now!

Do you have any doubts it makes problems / makes issues for users, or think it's stable enough for use?

harshad1:
> Checked out branch, and tried same like GIF from last time. Fixed!

Yes, the bug was actually unrelated to `singleTop`. It was a null de-reference which was not handled correctly. This is fixed now.

> Do you have any doubts it makes problems / makes issues for users, or think it's stable enough for use?

The thing I am trying to solve is that, when Multi Window is enabled, and a document is opened from from MainActivity, a new application window is opened. However when I launch a document from the widget or the shortcuts, it always opens in the last window. PendingIntent and regular intent behavior doesn't match.

If I cannot solve this today, I am going to clean up the PR a little and mark it ready

gsantner:
Leave it up to you, as long no crashes its OK from my side.

harshad1:
Made a new, cleaned up PR: https://github.com/gsantner/markor/pull/977

Closing this one

PR #971, opened 2020-06-03
PATTERN_PROJECTS PATTERN_CONTEXT regex modified

### LINKED ISSUE #957: todotxt: Allow to insert tags/categories with multiple ++/@@
#### General information

* **App version:** 
* **System:** 

#### Description

I use some context / list entries starting with double @, like `@@WartenAuf`, `@@Irgendwann`, `@@Test`

The main reason is: these are some meta-lists, they should be ordered and marked separately and maybe there are some other reasons. 

This is conform to the todotxt syntax. But Markor has some issues with these entries. The list of list entries is strange and they can't be assigned from the list list to an entry if they are not right in the list. The list startes

- "
- Irgendwann
- Test
- WartenAuf
- abc"

Could you handle @@ (and maybe @@@, who knows)?

aisbergde:
fixes #957 

current definition:

    public static final Pattern PATTERN_PROJECTS = Pattern.compile("\\B(?:\\++)(\\S+)");
    public static final Pattern PATTERN_CONTEXTS = Pattern.compile("\\B(?:\\@+)(\\S+)");

I am not really sure what the `Patter.compile` does, but from my understanding there is some tranformation, maybe the final regex is `\B(?:\@+)(\S+)`

I testerd here: [Online regex tester and debugger: PHP, PCRE, Python, Golang and JavaScript](https://regex101.com/) or here [RegExr: Learn, Build, &amp; Test RegEx](https://regexr.com/)

Test cases:

```
@aaa
qwert @bbb
asd @@ccc
ghi @@@ddd
jkl a@bbb
mno c@@ddd
```

`\B(?:@+)(\S+)`

current definition, has issues

`\B\@(\@*\S+)`

issues with

> mno c@@ddd

`\s\@(\@*\S+)`

works best

### DIFF
--- a/app/src/main/java/net/gsantner/opoc/format/todotxt/SttCommander.java
+++ b/app/src/main/java/net/gsantner/opoc/format/todotxt/SttCommander.java
@@ -46,8 +46,8 @@ public class SttCommander {
             " )?" + // End of prefix
             "((a|.)*)" // Take whats left
     );
-    public static final Pattern PATTERN_PROJECTS = Pattern.compile("\\B(?:\\++)(\\S+)");
-    public static final Pattern PATTERN_CONTEXTS = Pattern.compile("\\B(?:\\@+)(\\S+)");
+    public static final Pattern PATTERN_PROJECTS = Pattern.compile("\\s\\+(\\+*\\S+)");
+    public static final Pattern PATTERN_CONTEXTS = Pattern.compile("\\s\\@(\\@*\\S+)");
     public static final Pattern PATTERN_DONE = Pattern.compile("(?m)(^[Xx]) (.*)$");
     public static final Pattern PATTERN_DATE = Pattern.compile("(?:^|\\s|:)(" + PT_DATE + ")(?:$|\\s)");
     public static final Pattern PATTERN_KEY_VALUE_PAIRS__TAG_ONLY = Pattern.compile("(?i)([a-z]+):([a-z0-9_-]+)");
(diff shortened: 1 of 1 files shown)


gsantner:
`c@@ddd` should not be marked, only when starts with @

aisbergde:
You are right, it doesn't work when on the very first line :-(

I don't know how this all works in the program, can't compile and test. I tested only on these websites.
`\B` has some issues with double @@ inside a word

`\B\@(\@*\S+)`

>@aaa
>aaa@bbb
>ccc@@ddd

will be parsed as

>@aaa
>@ddd

but maybe this is OK in most cases?

This could do the job, but only if you parse per line, not if you parse a multi line text
`\s|^`

`\s|^\@(\@*\S+)`

But If I see how you parse begin of line in other patterns, this should work?:
    public static final Pattern PATTERN_PRIORITY_A = Pattern.compile("(?:^|\\n)\\(([Aa])\\)\\s");

Should I change or will you change wile merging?
Then this could be OK?

aisbergde:
public static final Pattern PATTERN_PROJECTS = Pattern.compile("(?:^|\\n|\\s)\\+(\\+*\\S+)");
    public static final Pattern PATTERN_CONTEXTS = Pattern.compile("(?:^|\\n|\\s)\\@(\\@*\\S+)");

aisbergde:
Sorry, I never really used all this regex staff, only for very simple expressions.

harshad1:
This is the website i have been using to test my regexes https://www.freeformatter.com/java-regex-tester.html.

gsantner:
I just temporary lock that one, to have discussion/comments at #957 (else it gets hard to follow)

gsantner:
Closing as invalid.

As noted in issue, the regex is fine. Its only about inserting/loading into tag/project in dialog, where we strip +/@ and just add one.

PR #975, opened 2020-06-10
Smooth horizontal scrolling (fixes #973)

### LINKED ISSUE #973: Improve edit mode horizontal scrolling
In edit mode, if line wrap is disabled, horizontal scrolling is jerky and interrupted. Scrolling up and down is very smooth. This affects all file types.

Expected: I can scroll smoothly left and right as I can up and down.
Observed: If I drag left or right to scroll, the view scrolls a few characters and then stops. I cannot fling to scroll fast either.

harshad1:
This PR addresses https://github.com/gsantner/markor/issues/973. In this PR I add a horizontal scroll view to the editor in order to enable smooth left-right scrolling.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
@@ -202,7 +202,7 @@ public class DocumentActivity extends AppActivityBase {
 
     @Override
     public boolean dispatchTouchEvent(MotionEvent event) {
-        if (_appSettings.isSwipeToChangeMode() && getCurrentVisibleFragment() instanceof DocumentEditFragment) {
+        if (_appSettings.isSwipeToChangeMode() && _appSettings.isEditorLineBreakingEnabled() && getCurrentVisibleFragment() instanceof DocumentEditFragment) {
             try {
                 Rect activityVisibleSize = new Rect();
                 getWindow().getDecorView().getWindowVisibleDisplayFrame(activityVisibleSize);
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
@@ -542,8 +542,9 @@ public class DocumentEditFragment extends GsFragmentBase implements TextFormat.T
             _primaryScrollView.removeAllViews();
             if (enable) {
                 HorizontalScrollView hsView = new HorizontalScrollView(context);
-                _primaryScrollView.addView(hsView);
+                hsView.setFillViewport(true);
                 hsView.addView(_hlEditor);
+                _primaryScrollView.addView(hsView);
             } else {
                 _primaryScrollView.addView(_hlEditor);
             }
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
@@ -33,6 +33,7 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.webkit.WebSettings;
 import android.webkit.WebView;
+import android.widget.HorizontalScrollView;
 import android.widget.SearchView;
 import android.widget.TextView;
 import android.widget.Toast;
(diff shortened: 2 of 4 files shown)


app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -522,13 +527,29 @@ private void setupAppearancePreferences(View fragmentView) {
         AppSettings as = AppSettings.get();
         _hlEditor.setTextSize(TypedValue.COMPLEX_UNIT_SP, as.getFontSize());
         _hlEditor.setTypeface(FontPreferenceCompat.typeface(getContext(), as.getFontFamily(), Typeface.NORMAL));
-        _hlEditor.setHorizontallyScrolling(!as.isEditorLineBreakingEnabled() && !isInMainActivity);
+        setHorizontalScrollMode(!as.isEditorLineBreakingEnabled() && !isInMainActivity);
```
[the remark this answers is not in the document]
harshad1: If we force-disable hscroll in mainactivity, we will have an inconsistent behavior between opening (say) quicknote through the shortcut and tapping on the icon in the main app.

I did not mean a popup, I meant to change the option message to: 
```xml
<string name="soft_break_if_not_fit_to_width_else_enable_horizontal_scrolling">Soft break if text doesn\'t fit to screen width. Else enable horizontal scrolling. Disabling this option will block swiping to change modes or views.</string>
```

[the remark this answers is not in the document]
gsantner: > If we force-disable hscroll in mainactivity, we will have an inconsistent behavior between opening (say) quicknote through the shortcut and tapping on the icon in the main app.

Yes I am aware of that, and totally fine with that. Please nothing with hscroll at MainActivity.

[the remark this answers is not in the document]
harshad1: got it :+1:

app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
```diff
@@ -202,7 +202,7 @@ protected void onCreate(Bundle savedInstanceState) {
 
     @Override
     public boolean dispatchTouchEvent(MotionEvent event) {
-        if (_appSettings.isSwipeToChangeMode() && getCurrentVisibleFragment() instanceof DocumentEditFragment) {
+        if (_appSettings.isSwipeToChangeMode() && _appSettings.isEditorLineBreakingEnabled() && getCurrentVisibleFragment() instanceof DocumentEditFragment) {
```
harshad1, line 205: @gsantner I have also added this test to only enable swipe to change mode if line breaking is enabled.

Would you prefer if I flip this logic? i.e. only enable horizontal scrolling if swipe to change mode is disabled?

gsantner: experimental swipe-to-change doesn't work well..whenever hscroll is enabled it's fine that hscroll is enabled and swipe ignored.

gsantner: we can basically also remove swipe and think of other solution...it makes more problems than it tries to resolve at this point

harshad1: In the future, we could probably enable multi-finger gestures. 1 fingers = scroll. 2 fingers = swipe or something like that. The button is always available in the tool bar any way, so swipe is an extra convenience.

harshad1:
@gsantner This should be good to go.

harshad1:
@gsantner Does it make sense to move the line break / wrap option to the editor menu? i.e. along with save, share etc. That way you can quickly switch between wrap and horizontal scrolling, which is useful imo.

gsantner:
@harshad1 
Basically everything about editor options might be of user interest to have options directly inside toolbar. I suggest however to not make exception at wrapping now to handle it differently.

I.e. at some future some/most things might be placed in editor itself at one go. But I don't suggest to happen soon..there are so many configurable things and users are overwhelmed if thers like additional 20 buttons :smile: .

harshad1:
:+1:

gsantner:
@harshad1 
I checked out the branch, please see:

video_2020-06-13_14-38-07

Basically from what I see:
* We definitely need MATCH_PARENT. At clearest you can see stuff doesn't expand..at todotxt (width of dividers). Maybe you don't see that on your device that much because it is smaller?
* Theres issue with adding/removing/swapping the editor component, and seemingly others as well. Most things are duplicated in UI.
* And btw, yes the "UI glitch" on the right side at the end of video is really looking like this in front of the device. (It's not a recording artifacts)

harshad1:
@gsantner I think this should be fixed now

gsantner:
@harshad1 
Thanks much, I've tried it and haven't noticed noted issue anymore :D.

Merging :tada:  :+1: 

self note: checked wrap option = no horiz scroll, unchecked = horiz scroll

PR #977, opened 2020-06-16
Simplified open correct file

harshad1:
This PR supersedes https://github.com/gsantner/markor/pull/970.

In this PR I address 3 issues:
1. Always open the correct file from shortcuts and the widget
2. Correctly handle multi windows
3. Widget new file is added to correct folder (#873)

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentOpener.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentOpener.java
@@ -1,0 +1,31 @@
+package net.gsantner.markor.activity;
+
+import android.app.Activity;
+import android.content.Intent;
+import android.os.Bundle;
+
+/**
+ * This Activity exists solely to launch DocumentActivity with the correct intent
+ * it is necessary as widget and shortcut intents do not respect MultipleTask etc
+ */
+public class DocumentOpener extends Activity {
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        launchDocumentActivityAndFinish(getIntent());
+    }
+
+    @Override
+    protected void onNewIntent(Intent intent) {
+        super.onNewIntent(intent);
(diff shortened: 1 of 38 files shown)


app/src/main/res/xml/prefactions_share_into.xml
```diff
@@ -40,7 +40,6 @@
         android:title="@string/calendar" />
 
     <PreferenceCategory
-        android:key="@string/pref_key__favourite_files"
```
harshad1, line 43: Removed as android studio complained about this key being a duplicate. Don't think there is any detriment to removing it.

harshad1:
@gsantner I will be testing this on another device later in the evening. If you could test it, too that would be great. Otherwise it is good to go

app/src/main/java/net/gsantner/markor/activity/DocumentOpener.java
```diff
@@ -0,0 +1,31 @@
...
+
+import android.app.Activity;
+import android.content.Intent;
+import android.os.Bundle;
+
+/**
+ * This Activity exists solely to launch DocumentActivity with the correct intent
+ * it is necessary as widget and shortcut intents do not respect MultipleTask etc
+ */
+public class DocumentOpener extends Activity {
```
gsantner, line 11: Suggestion postfix with Activity

app/src/main/res/xml/prefactions_share_into.xml
gsantner: I see, seems like copy paste error

app/src/main/java/net/gsantner/markor/activity/DocumentOpener.java
gsantner: renamed pushed

gsantner:
Just merged in master and renamed activity, please pull before changing code :D

gsantner:
thinks did some minor changes and merged!

Issue #979, opened 2020-06-17
The markdown list actions should correctly change list types


harshad1:
The list type should _change_ rather than being _preprended_

For example, the unordered list action should convert `1. foo` to `- foo` instead of `- 1. foo`, and vice-versa.

PR #983, opened 2020-06-21
Properly change list prefix types

harshad1:
This PR addresses https://github.com/gsantner/markor/issues/979.

In this PR, I have added functionality to properly transition between prefix types.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/general/CommonTextActions.java
+++ b/app/src/main/java/net/gsantner/markor/format/general/CommonTextActions.java
@@ -31,8 +31,6 @@ import net.gsantner.opoc.util.Callback;
 import net.gsantner.opoc.util.ContextUtils;
 import net.gsantner.opoc.util.StringUtils;
 
-import java.util.Arrays;
-
 @SuppressWarnings("WeakerAccess")
 public class CommonTextActions {
     public static final int ACTION_SPECIAL_KEY__ICON = R.drawable.ic_keyboard_black_24dp;
@@ -234,9 +232,7 @@ public class CommonTextActions {
 
         int lineStart = StringUtils.getLineStart(text, selectionStart);
 
-        char[] tabChars = new char[_tabWidth];
-        Arrays.fill(tabChars, ' ');
-        String tabString = new String(tabChars);
+        String tabString = StringUtils.repeatChars(' ', _tabWidth);
 
         while (lineStart <= selectionEnd) {
 
--- a/app/src/main/java/net/gsantner/markor/format/general/CommonTextActions.java
+++ b/app/src/main/java/net/gsantner/markor/format/general/CommonTextActions.java
@@ -245,7 +245,7 @@ public class CommonTextActions {
                 int spaceCount = textStart - lineStart;
                 int delCount = Math.min(_tabWidth, spaceCount);
                 int delEnd = lineStart + delCount;
-                if (delCount > 0 && delEnd < text.length()) {
+                if (delCount > 0 && delEnd <= text.length()) {
                     text.delete(lineStart, delEnd);
                     selectionEnd -= delCount;
                 }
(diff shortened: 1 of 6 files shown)


app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextActions.java
```diff
@@ -24,12 +24,34 @@
...
+            PREFIX_ORDERED_LIST,
+            PREFIX_ATX_HEADING,
+            PREFIX_QUOTE,
+            PREFIX_CHECKED_LIST,
+            PREFIX_UNCHECKED_LIST,
+            // Unordered has to be after checked list
+            PREFIX_UNORDERED_LIST,
+            PREFIX_LEADING_SPACE,
+    };
+
```
harshad1, line 54: should these patterns go somewhere else?

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextActions.java
```diff
@@ -114,10 +136,23 @@ private boolean onClickImpl(final View view) {
...
+                    final String uncheck = "$1" + listChar + " [ ] ";
+                    final String check = "$1" + listChar + " [x] ";
+                    runPrefixReplaceAction(PREFIX_UNCHECKED_LIST, uncheck, check);
                     return true;
-                }*/
+                }
+                case R.string.tmaid_common_ordered_list_number: {
+                    runPrefixReplaceAction(PREFIX_ORDERED_LIST, "$11. ", "$1");
+                    return true;
+                }
```
harshad1, line 155: Copied here from Textactions. As these changes are markdown-specific, I have added them here. Plain text will still default to the previous `runRegularPrefixAction` as found in `TextActions`.

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextActions.java
gsantner: do we have similar in plaintext format as well?

harshad1: We do not have similar patterns in plain text. I have made changes only to markdown behavior.

gsantner: hm then it was something different :smile:

app/src/main/java/net/gsantner/markor/format/general/CommonTextActions.java
```diff
@@ -245,7 +241,7 @@ protected void runIndentLines(Boolean deIndent) {
                 int spaceCount = textStart - lineStart;
                 int delCount = Math.min(_tabWidth, spaceCount);
                 int delEnd = lineStart + delCount;
-                if (delCount > 0 && delEnd < text.length()) {
+                if (delCount > 0 && delEnd <= text.length()) {
                     text.delete(lineStart, delEnd);
```
gsantner, line 245: potential out of index exception, can you confirm works fine on last line/ last character?

harshad1: I made this change because de-indent was not working on the last line. WIll test and confirm

app/src/main/java/net/gsantner/opoc/util/StringUtils.java
```diff
@@ -77,4 +79,10 @@ public static int getNextNonWhitespace(CharSequence s, int start, int maxRange)
         return selection;
     }
 
+    public static String repeatChars(Character character, int count) {
```
gsantner, line 82: suggest to use char, and convert to Character inside if required - it's impossible to get null exception this way

app/src/main/java/net/gsantner/markor/format/general/CommonTextActions.java
harshad1: Looks like it works. I will be testing this for a day or so before I give the :+1:

harshad1: Good to go :+1:

Issue #978, opened 2020-06-17
markdown: Auto update succeeding numbered list items when adding/updating in the middle


harshad1:
When editing ordered lists, the list numbering often goes out of order. While this does not affect the markdown rendering, it makes lists a little harder to read in edit mode.

In order to address this, Markor should have an option to reorder the current list.

harshad1:
Probably the best way to activate this is by long pressing the ordered list action button. When long pressed, the list under the cursor will be walked and re-numbered.

PR #987, opened 2020-06-25
Renumber ordered list

harshad1:
This PR addresses https://github.com/gsantner/markor/issues/978

In this PR, I have added functionality to reorder the current ordered list. Only the current list level is selected. The first number in the current list level is used as the starting number.

Todo:
- [x] Investigate whether we want to update the entire list (all levels)
- [x] Investigate whether we want to add functionality to automatically call this routine as the user types

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/TextFormat.java
+++ b/app/src/main/java/net/gsantner/markor/format/TextFormat.java
@@ -90,8 +90,8 @@ public class TextFormat {
             default:
             case FORMAT_MARKDOWN: {
                 format.setConverter(CONVERTER_MARKDOWN);
-                format.setTextActions(new MarkdownTextActions(activity, document));
                 format.setHighlighter(new MarkdownHighlighter(hlEditor, document));
+                format.setTextActions(new MarkdownTextActions(activity, document));
                 break;
             }
         }
--- a/app/src/main/java/net/gsantner/markor/format/TextFormat.java
+++ b/app/src/main/java/net/gsantner/markor/format/TextFormat.java
@@ -90,8 +90,9 @@ public class TextFormat {
             default:
             case FORMAT_MARKDOWN: {
                 format.setConverter(CONVERTER_MARKDOWN);
-                format.setHighlighter(new MarkdownHighlighter(hlEditor, document));
-                format.setTextActions(new MarkdownTextActions(activity, document));
+                MarkdownTextActions mta = new MarkdownTextActions(activity, document);
+                format.setTextActions(mta);
+                format.setHighlighter(new MarkdownHighlighter(hlEditor, document, mta));
                 break;
             }
         }
--- a/app/src/main/java/net/gsantner/markor/format/TextFormat.java
+++ b/app/src/main/java/net/gsantner/markor/format/TextFormat.java
@@ -90,9 +90,8 @@ public class TextFormat {
             default:
             case FORMAT_MARKDOWN: {
                 format.setConverter(CONVERTER_MARKDOWN);
-                MarkdownTextActions mta = new MarkdownTextActions(activity, document);
-                format.setTextActions(mta);
-                format.setHighlighter(new MarkdownHighlighter(hlEditor, document, mta));
+                format.setTextActions(new MarkdownTextActions(activity, document));
+                format.setHighlighter(new MarkdownHighlighter(hlEditor, document));
                 break;
             }
         }
--- a/app/src/main/java/net/gsantner/markor/format/markdown/ListHandler.java
+++ b/app/src/main/java/net/gsantner/markor/format/markdown/ListHandler.java
@@ -19,19 +19,21 @@ import net.gsantner.opoc.util.StringUtils;
 import java.util.regex.Matcher;
 
 public class ListHandler implements TextWatcher {
-    private boolean _runReorder = false;
-    private int reorderPosition = -1;
+    private final boolean _reorderEnabled;
+    private int reorderPosition;
+    private boolean triggerReorder = false;
 
 
-    public ListHandler(final boolean runReorder) {
+
+    public ListHandler(final boolean reorderEnabled) {
         super();
-        _runReorder = runReorder;
+        _reorderEnabled = reorderEnabled;
     }
 
     @Override
     public void onTextChanged(CharSequence s, int start, int before, int count) {
 
-        reorderPosition = -1;
+        triggerReorder |= containsNewline(s, start, count);
 
         // Detects if enter pressed on empty list (correctly handles indent) and marks line for deletion.
         if (count > 0 && start > -1 && start < s.length() && s.charAt(start) == '\n') {
@@ -66,14 +68,25 @@ public class ListHandler implements TextWatcher {
                 e.delete(e.getSpanStart(span), e.getSpanEnd(span));
             }
         }
-        if (_runReorder && reorderPosition > 0 && reorderPosition < e.length()) {
+        if (_reorderEnabled && triggerReorder && reorderPosition > 0 && reorderPosition < e.length()) {
             MarkdownAutoFormat.renumberOrderedList(e, reorderPosition);
         }
     }
 
     @Override
     public void beforeTextChanged(CharSequence s, int start, int count, int after) {
-        // Not used
+         triggerReorder = containsNewline(s, start, count);
+         reorderPosition = start;
+    }
+
+    private boolean containsNewline(CharSequence s, int start, int count) {
+        final int end = start + count;
+        for (int i = start; i < end; i ++) {
+            if (s.charAt(i) == '\n') {
+                return true;
+            }
+        }
+        return false;
     }
 }
 
--- a/app/src/main/java/net/gsantner/markor/format/markdown/ListHandler.java
+++ b/app/src/main/java/net/gsantner/markor/format/markdown/ListHandler.java
@@ -24,7 +24,6 @@ public class ListHandler implements TextWatcher {
     private boolean triggerReorder = false;
 
 
-
     public ListHandler(final boolean reorderEnabled) {
         super();
         _reorderEnabled = reorderEnabled;
@@ -75,13 +74,13 @@ public class ListHandler implements TextWatcher {
 
     @Override
     public void beforeTextChanged(CharSequence s, int start, int count, int after) {
-         triggerReorder = containsNewline(s, start, count);
-         reorderPosition = start;
+        triggerReorder = containsNewline(s, start, count);
+        reorderPosition = start;
     }
 
     private boolean containsNewline(CharSequence s, int start, int count) {
         final int end = start + count;
-        for (int i = start; i < end; i ++) {
+        for (int i = start; i < end; i++) {
             if (s.charAt(i) == '\n') {
                 return true;
             }
--- a/app/src/main/java/net/gsantner/markor/format/markdown/ListHandler.java
+++ b/app/src/main/java/net/gsantner/markor/format/markdown/ListHandler.java
@@ -32,7 +32,7 @@ public class ListHandler implements TextWatcher {
     @Override
     public void onTextChanged(CharSequence s, int start, int before, int count) {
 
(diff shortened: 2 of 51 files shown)


app/src/main/java/net/gsantner/markor/format/markdown/MarkdownHighlighter.java
```diff
@@ -36,7 +35,7 @@
     private static final int MD_COLOR_QUOTE = 0xff88b04c;
     private static final int MD_COLOR_CODEBLOCK = 0xff8c8c8c;
 
-    public MarkdownHighlighter(HighlightingEditor hlEditor, Document document) {
+    public MarkdownHighlighter(HighlightingEditor hlEditor, Document document, MarkdownTextActions actions) {
```
harshad1, line 38: Passing MarkdownTextActions in feels inelegant. What do you think @gsantner.

I may remove this functionality.

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownHighlighterPattern.java
```diff
@@ -18,7 +18,7 @@
     HEADING_SIMPLE(Pattern.compile("(?m)^(#{1,6}\\s.*$)")),
     LINK(Pattern.compile("\\[([^\\[]+)\\]\\(([^\\)]+)\\)")),
     LIST_UNORDERED(Pattern.compile("(\\n|^)\\s{0,16}([*+-])( \\[[ xX]\\])?(?= )")),
-    LIST_ORDERED(Pattern.compile("(?m)^\\s{0,16}\\d+\\.\\s")),
+    LIST_ORDERED(Pattern.compile("(?m)^\\s{0,16}(\\d+)(:?\\.|\\))\\s")),
```
harshad1, line 21: Handle ordered lists demarcated with ')' per commonmarkd

app/src/main/java/net/gsantner/markor/format/TextFormat.java
```diff
@@ -90,8 +90,9 @@ public static TextFormat getFormat(int formatId, Activity activity, Document doc
             default:
             case FORMAT_MARKDOWN: {
                 format.setConverter(CONVERTER_MARKDOWN);
-                format.setHighlighter(new MarkdownHighlighter(hlEditor, document));
-                format.setTextActions(new MarkdownTextActions(activity, document));
+                MarkdownTextActions mta = new MarkdownTextActions(activity, document);
+                format.setTextActions(mta);
+                format.setHighlighter(new MarkdownHighlighter(hlEditor, document, mta));
```
harshad1, line 95: See comment below. I feel this is not very clean. Not sure if 'update as you type' is needed at all.

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownHighlighter.java
harshad1: Solved.

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextActions.java
```diff
@@ -305,4 +313,183 @@ private void runPrefixReplaceAction(final Pattern actionPattern, final String ac
 
         runRegexReplaceAction(patterns);
     }
+
+
+    private static class ListLine {
```
harshad1, line 318: All of this should probably move to a separate file. But for now, its here

harshad1:
@gsantner This is ready for review / testing / merging

app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -170,6 +170,10 @@ public boolean isMarkdownDisableCodeBlockHighlight() {
         return getBool(R.string.pref_key__markdown__disable_code_block_highlight, false);
     }
 
+    public boolean isMarkdownAutoUpdateList() {
+        return getBool(R.string.pref_key__markdown__auto_renumber_ordered_list, false);
```
gsantner, line 174: I'm fine with it being easy to enable/disable in code, but we don't need a settings option for it. It's fine to have this appsettings method and just return true. And have no settings.xml option.

app/src/main/res/xml/preferences_master.xml
```diff
@@ -403,6 +403,12 @@
                     android:icon="@drawable/ic_code_black_24dp"
                     android:key="@string/pref_key__markdown__disable_code_block_highlight"
                     android:title="@string/disable_code_block_highlight" />
+                <CheckBoxPreference
```
gsantner, line 406: I'm fine with it being easy to enable/disable in code, but we don't need a settings option for it. It's fine to have this appsettings method and just return true. And have no settings.xml option.

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextActions.java
gsantner: I'm fine with having it in one of the existing markdown classes. However, I feel the better fit is `MarkdownAutoFormat.java`, which has exactly the use of "do something when user enters text" .... auto-heading etc.

---> ListLine,  OrderedListLine, UnOrderedListLine

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownHighlighterPattern.java
gsantner: If you think it improves something it's ok.

app/src/main/res/xml/preferences_master.xml
harshad1: I think we should either
1. Have a settings option for it
2. Do not enable auto-update at all (manual update is still possible with long-press)

If we permanently enable this, then it will be impossible to edit lists without changing the numbering.

For example, I sometimes like to leave my lists as:
```
1. foo
1. bar
1. baz
```
Cant do this if the auto-renumber is always true as entering a new item will renumber the entire list.

harshad1: Another issue is there may be performance issues on long lists and older phones.

harshad1:
Changing to WIP - will be adding logic for multi-level list renumbering and some other fixes

harshad1:
@gsantner This is now ready

app/src/main/java/net/gsantner/markor/format/markdown/ListHandler.java
```diff
@@ -7,23 +7,33 @@
...
+
+    public ListHandler(final boolean reorderEnabled) {
+        super();
+        _reorderEnabled = reorderEnabled;
+    }
 
     @Override
     public void onTextChanged(CharSequence s, int start, int before, int count) {
 
+        triggerReorder |= containsNewline(s, start, count);
```
gsantner, line 35: I know it does same, but please lets ahve `x = x || y()`

app/src/main/java/net/gsantner/markor/format/markdown/ListHandler.java
```diff
@@ -38,8 +48,12 @@ public void onTextChanged(CharSequence s, int start, int before, int count) {
                 sSpan.setSpan(this, iStart, start + 1, Spanned.SPAN_COMPOSING);
             } else {
                 Matcher oMatch = MarkdownHighlighterPattern.LIST_ORDERED.pattern.matcher(previousLine);
-                if (oMatch.find() && previousLine.equals(oMatch.group(1) + ". ")) {
-                    sSpan.setSpan(this, iStart, start + 1, Spanned.SPAN_COMPOSING);
+                if (oMatch.find()) {
+                    if (previousLine.equals(oMatch.group(1) + ". ")) {
```
gsantner, line 52: Do I see right, this is the only relevant point for ordered list on each line (ie  1. text\n2. text\3. text)? If so, please make `". "` a `final String ul_sep = ". ";` in the methods beginning. This makes it easier to see next time when coming here.... where the most important point is.

app/src/main/res/values/strings.xml
```diff
@@ -174,6 +174,8 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="highlight_line_ending_if_two_or_more_spaces">Highlight spaces at the end of lines containing two or more spaces</string>
     <string name="use_monospace_for_code">Use monospace font for code</string>
     <string name="disable_code_block_highlight">Disable highlighting for code blocks</string>
+    <string name="auto_renumber_ordered_list">Auto-renumber ordered lists</string>
+    <string name="auto_renumber_ordered_list_slowdown">This option may impact app performance.</string>
```
gsantner, line 178: Have you really spotted a decline in performance? To be honest, the only point where it would be slow is when pressing enter right? I feel like there is no need to confuse users and just have in description what it does. We can add a performance notice later when we see it's really required.

app/src/main/res/values/strings.xml
```diff
@@ -349,7 +351,7 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="example_of_a_markdown_table">Example of a Markdown table</string>
     <string name="delete_lines">Delete lines</string>
     <string name="unordered_list">Unordered list</string>
-    <string name="ordered_list">Ordered list</string>
```
gsantner, line 352: This is used for description at UI Ordererd list button too as far I know. Please have a new renumber_ordered_list key

gsantner:
@harshad1 
Thanks, I looked over it on non-developer PC so only looked into code but not install try.

From code I only see some quite minor stuff.  
I also feel we might be able to move some stuff to generic classes later on, but thats something really "all at once later on" and not required on this PR.

app/src/main/java/net/gsantner/markor/format/markdown/ListHandler.java
harshad1: I'll be fixing this in the next PR :) where I update the way Listhandler detects empty lines among other things.

app/src/main/res/values/strings.xml
harshad1: Long press of the ordered list action triggers a renumber ordered list for the current list. This is quite useful. Changing the above string changes the long-press message.

app/src/main/res/values/strings.xml
harshad1: I have not spotted a decline in performance but I use a pixel 3, which is a pretty fast phone. I suppose a user could switch to plaintext if they wanted to disable auto-renumbering. 

I will disable this for now.

harshad1:
@gsantner

app/src/main/res/xml/preferences_master.xml
```diff
@@ -403,6 +403,12 @@
                     android:icon="@drawable/ic_code_black_24dp"
                     android:key="@string/pref_key__markdown__disable_code_block_highlight"
                     android:title="@string/disable_code_block_highlight" />
+                <!-- CheckBoxPreference
+                    android:defaultValue="false"
+                    android:icon="@drawable/ic_format_list_numbered_black_24dp"
+                    android:key="@string/pref_key__markdown__auto_renumber_ordered_list"
+                    android:title="@string/auto_renumber_ordered_list"
+                    android:summary="@string/auto_renumber_ordered_list_slowdown"/ -->
```
harshad1, line 411: Left in the option code but commented it out

app/src/main/res/values/strings.xml
gsantner: @harshad1 still please use new res renumber_ordered_list if you change the string, OR revert that line of change.

Reason is that translation services/Crowdin act bad on string change, and if the rate is matching may reuse the old one in a given language automatically

gsantner:
thanks, merged

gsantner:
seems like theres issue with github/api today.
I see you created two new PRs (mail) but I dont see them in browser.

harshad1:
@gsantner It's actually just 1 PR. I closed and re-created it as new pushed commits were not showing up in the PR. Let's wait a day or so and see. If you still can't see the PR, I will re-create them again.

Issue #994, opened 2020-07-09
Question regarding newfile + overwrite


subjekcintasemuacg:
In newest update, making file/title with same name as existing file automatically DELETED existing file contents! Isn't it supposed to "direct" you to your existing file???????????

Uh this is my first time making issue I don't know what to write.....

#### General information

* **App version:** 
* **System:**

gsantner:
I can reproduce on current git version that it's overwriting.

However not sure if it's better to 

* "remove no overwrite notice" 
* OR "fix no overwrite on existing file".

What you think?

subjekcintasemuacg:
Uh I don't quite understand. And I'm not really sure about the git thing.
Is it possible to just do it like last time??

I think it's really awesome before.
If there's a "notice" I think it's too cluttering. And "redirect" to the existing file is just perfect.

gsantner:
OK so option 2 is more the one that is wanted.

subjekcintasemuacg:
Ok. Thank you.
Cannot wait for next update.

harshad1:
@gsantner My understanding is that the desired behavior is 'if the file already exists, then don't re-create it, instead just show it'

gsantner:
Jup that is right

Russe:
Uhh, just lost a file with this behaviour.
I agree, a note is not a good idea, because it can be missed or misinterpreted.
So your decision to show the already existing file instead of recreating it is fine for me.

gsantner:
reopened

@harshad1 
I found in latest release that it's overwriting still/again

harshad1:
I'm not able to reproduce this. Could you give me the exact steps to reproduce?

gsantner:
Create file a.md, go back to file browser, create new file again with a.md (empty file template).And then whatever was in a.md got emptied out, empty file shows up

harshad1:
I followed those steps with a fresh install from F-Droid and the file was not overwritten.

What software do you use to record your phone, and how do you upload the recordings here?

gsantner:
I use  normal screenrecord from OS and ffmpeg to GIF. Theres also "Video transcoder" on fdroid which can do the same.


![Screenshot_20200803-053700_Markor](https://user-images.githubusercontent.com/6735650/89143608-7b881200-d54b-11ea-9139-1e6b7d1d241b.png)


[...]

PR #998, opened 2020-07-14
Don't nuke existing file

harshad1:
This PR addresses https://github.com/gsantner/markor/issues/994.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
+++ b/app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
@@ -157,7 +157,7 @@ public class NewFileDialog extends DialogFragment {
                     final byte[] templateContents = getTemplateContent(templateSpinner, basedir, encryptCheckbox.isChecked());
                     shareUtil.writeFile(f, false, (arg_ok, arg_fos) -> {
                         try {
-                            if (f.exists() && f.length() < 5 && templateContents != null) {
+                            if (f.exists() && f.length() == 0 && templateContents != null) {
                                 arg_fos.write(templateContents);
                             }
                         } catch (Exception ignored) {
--- a/app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
+++ b/app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
@@ -157,7 +157,7 @@ public class NewFileDialog extends DialogFragment {
                     final byte[] templateContents = getTemplateContent(templateSpinner, basedir, encryptCheckbox.isChecked());
                     shareUtil.writeFile(f, false, (arg_ok, arg_fos) -> {
                         try {
-                            if (f.exists() && f.length() == 0 && templateContents != null) {
+                            if (f.exists() && f.length() < 5 && templateContents != null) {
                                 arg_fos.write(templateContents);
                             }
                         } catch (Exception ignored) {
--- a/app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
+++ b/app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
@@ -157,7 +157,7 @@ public class NewFileDialog extends DialogFragment {
                     final byte[] templateContents = getTemplateContent(templateSpinner, basedir, encryptCheckbox.isChecked());
                     shareUtil.writeFile(f, false, (arg_ok, arg_fos) -> {
                         try {
-                            if (f.exists() && f.length() < 5 && templateContents != null) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
```diff
@@ -157,7 +157,7 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
                     final byte[] templateContents = getTemplateContent(templateSpinner, basedir, encryptCheckbox.isChecked());
                     shareUtil.writeFile(f, false, (arg_ok, arg_fos) -> {
                         try {
-                            if (f.exists() && f.length() < 5 && templateContents != null) {
```
harshad1, line 160: Why do we have < 5? Because it is ok to append a template to a nearly empty file but not to a file with more than a few characters?

I can revert this, if we want to preserve the < 5 logic

app/src/main/java/net/gsantner/markor/util/DocumentIO.java
```diff
@@ -45,7 +45,7 @@
     public static final String EXTRA_PATH_IS_FOLDER = "EXTRA_PATH_IS_FOLDER"; // boolean
 
     public static final int MAX_TITLE_EXTRACTION_LENGTH = 25;
-    public static boolean SAVE_IGNORE_EMTPY_NEXT_TIME = false;
+    public static boolean SAVE_IGNORE_EMTPY_NEXT_TIME = true;
```
harshad1, line 48: @gsantner Why do we prevent saving empty files? I can revert this too but I made the change so I could ask.

app/src/main/java/net/gsantner/opoc/util/ShareUtil.java
```diff
@@ -1136,7 +1136,7 @@ public void writeFile(final File file, final boolean isDirectory, final Callback
                 if (isDirectory) {
                     file.mkdirs();
                 } else {
-                    fileOutputStream = new FileOutputStream(file);
+                    fileOutputStream = new FileOutputStream(file, true);
```
harshad1, line 1139: Adding append here seems to solve the overwrite issue

harshad1: Needs more testing

app/src/main/java/net/gsantner/markor/util/DocumentIO.java
gsantner, line 48: Please revert yes. Not save empty files.

app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
gsantner: It's mere for me and others doing testfiles. Like create file to see how / if stuff works, and then overwrite it again with something else.

app/src/main/java/net/gsantner/markor/util/DocumentIO.java
gsantner: Please for sure revert.

For reason, some devices have either slow disk i/o OR slow cpu filling edittext slow. In either combination it may (and has happened) to result in empty files being written.

app/src/main/java/net/gsantner/opoc/util/ShareUtil.java
gsantner: Description of java append is: `    append - if true, then bytes will be written to the end of the file rather than the beginning`. I would say this is not the right way.

app/src/main/java/net/gsantner/opoc/util/ShareUtil.java
```diff
@@ -1132,7 +1134,9 @@ public void writeFile(final File file, final boolean isDirectory, final Callback
         try {
             FileOutputStream fileOutputStream = null;
             ParcelFileDescriptor pfd = null;
-            if (file.canWrite() || (!file.exists() && file.getParentFile().canWrite())) {
+            final boolean existingEmptyFile = file.canWrite() && file.length() < MIN_OVERWRITE_LENGTH;
+            final boolean nonExistingCreatableFile = !file.exists() && file.getParentFile().canWrite();
+            if (existingEmptyFile || nonExistingCreatableFile) {
```
harshad1, line 1139: More explicit logic. Still needs more testing.

harshad1: @gsantner Please test this for all the scenarios you can think of. It seems to work for all that I have tried.

Issue #990, opened 2020-06-28
markdown: Enter press auto-insert empty checkbox list


harshad1:
Currently, in markdown, if I press 'Enter' and the current line is a checked checklist item a new completed checklist item is inserted. This is rarely useful and it would be better to always insert an unchecked checkbox.

```
- [x] foo <- CR here
```
should give 

```
- [x] foo
- [ ]
```

PR #1002, opened 2020-07-15
Improved auto insert list prefix

harshad1:
In this PR, I address #990. 

The new ListLine constructs are use to improve the way list prefixes are auto-inserted / auto removed in markdown.

Note: re-created PR as some commits were not showing up on the first PR (?!)

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextActions.java
+++ b/app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextActions.java
@@ -191,7 +191,11 @@ public class MarkdownTextActions extends TextActions {
                     });
                     return true;
                 }
-                case R.string.tmaid_common_indent:
+                case R.string.tmaid_common_indent: {
+                    if (_appSettings.isMarkdownAutoUpdateList()) {
+                        runPrefixReplaceAction(PREFIX_ORDERED_LIST, "$0", "$11$4 ");
+                    }
+                }
                 case R.string.tmaid_common_deindent: {
                     runCommonTextAction(_context.getString(_action));
                     runRenumberOrderedListIfRequired();
--- a/app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextActions.java
+++ b/app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextActions.java
@@ -191,11 +191,7 @@ public class MarkdownTextActions extends TextActions {
                     });
                     return true;
                 }
-                case R.string.tmaid_common_indent: {
-                    if (_appSettings.isMarkdownAutoUpdateList()) {
-                        runPrefixReplaceAction(PREFIX_ORDERED_LIST, "$0", "$11$4 ");
-                    }
-                }
+                case R.string.tmaid_common_indent:
                 case R.string.tmaid_common_deindent: {
                     runCommonTextAction(_context.getString(_action));
                     runRenumberOrderedListIfRequired();
--- a/app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
+++ b/app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
@@ -339,12 +339,19 @@ public abstract class TextActions {
                 Matcher matcher = pattern.searchPattern.matcher(line);
                 if (matcher.find()) {
 
-                    String newLine;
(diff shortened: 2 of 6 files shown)


app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextActions.java
```diff
@@ -191,7 +191,11 @@ private boolean onClickImpl(final View view) {
                     });
                     return true;
                 }
-                case R.string.tmaid_common_indent:
+                case R.string.tmaid_common_indent: {
+                    if (_appSettings.isMarkdownAutoUpdateList()) {
+                        runPrefixReplaceAction(PREFIX_ORDERED_LIST, "$0", "$11$4 ");
```
harshad1, line 196: Not strictly related to insert list. This makes sure ordered list numbering is restarted on indent.

app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
```diff
@@ -350,6 +352,10 @@ protected void runRegexReplaceAction(final List<ReplacePattern> patterns, final
 
             lineStart = StringUtils.getLineEnd(text, lineStart, selEnd) + 1;
         }
+
+        _hlEditor.setSelection(
+                StringUtils.getIndexFromLineOffset(text, lStart),
+                StringUtils.getIndexFromLineOffset(text, lEnd));
```
harshad1, line 358: This logic + the code in `StringUtils` preserves selection when action is pressed

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextActions.java
gsantner: this looks very cryptic, very sure don't know what it this line this does later on. I don't understand fully either even now. Can you please add a comment there?

gsantner:
other than comment, is this ready or WIP? By code I don't fully see what it does, but as long as you say it works it's OK.

harshad1:
It was a pretty cryptic way of doing things. I am removing it and replacing it was much more readable code built into the renumber logic itself. Here is an explanation for your curiosity:

```
// If auto-update is enabled, this replaces any ordered list numbering to '1. '
// The auto-update will then fix the numbering for the whole list
if (_appSettings.isMarkdownAutoUpdateList()) {
    // Replace ordered list prefix, with leading space ($1), the number 1,
    // The delimiter ($4) and a trailing space
    // Any other prefixes are replaced with ($0) i.e. no change
    runPrefixReplaceAction(PREFIX_ORDERED_LIST, "$0", "$11$4 ");
}
```

harshad1:
I will be testing my new code for a little while. I will let you know when it is ready (~1 day)

gsantner:
When performance is better with the first one, it's preffered one. What do you think is better? That would be enough documentation for that line at least.

harshad1:
> When performance is better with the original one, it's the preffered one. What do you think is better? That would be enough documentation for that line at least.

Performance should be better for my new version. It is doing much less work and is cleaner. I just didn't think of doing it this way until now :)

harshad1:
This PR is ready to merge

gsantner:
@harshad1 
Thanks, I've merged all open PRs.

Also released Markor update v2.3 (as planned for a long time but didn't happen yet :D).

I heard the post might be of your interest too :wink: . <https://gsantner.net/blog/2020/07/25/markor-v2.3-outline-action-custom-order-todotxt-markdown-plaintext.html>

gsantner:
@harshad1 
I'm getting lots of reports about crashes now since the release:

grafik

grafik

harshad1:
@gsantner looking at it

gsantner:
@harshad1 
Just pushed on master, can you please verify that it crashed before, and it's fixed by the change? Pushed this as quick hotfix

harshad1:
I am not able to reproduce the crash using the version from F-droid (2.3.1).

I see your changes. Looking into cases where things may break

gsantner:
let's maybe use such a safety guard in general in the future, costs not too much performance

harshad1:
> let's maybe use such a safety guard in general in the future, costs not too much performance

I think we can add range safety checks outside the loop and avoid all performance issues. I am writing this now.

I still haven't been able to reproduce the issue and I have been using master on my phone for a while now.

PR #1015, opened 2020-07-31
Bugfix: Delete empty list correctly

harshad1:
Currently, if the cursor is between the list prefix and the start of the line and enter is pressed, the list prefix is incorrectly deleted.

---
i.e. (`|` used to show cursor)
```
1. |foo
```
becomes
```
2. foo
```
---
Desired behavior is 
```
1. |foo
```
becomes
```
1. 
2. foo
```
---

This change fixes this issue. Deletion of truly empty lines is still handled correctly
```
1. foo
2. |
```
becomes
```
1. foo
|
```
This now works correctly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### Recent changes
+- Markdown: Correctly insert or remove list item on press enter at empty list item
+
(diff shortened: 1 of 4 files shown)


harshad1:
@gsantner This is good to go. I have tested it over the last couple days

gsantner:
Would you mind to add a line to changelog? I.e. Markdown: Make line empty when press enter on empy list element

harshad1:
Updated changelog

PR #1032, opened 2020-08-18
Rework todo.txt support

harshad1:
This PR is meant to facilitate discussion re simplified todo.txt processing. 

I have been using this version of TodoTxtActions and Stt* for a while now. It offers several advantages. The biggest 2 are:
1. The todotxt parsing code etc is much simpler (one file)
2. I can update multiple tasks at the same time (e.g. Highlight 5 tasks and add a context to all of them)

I originally did not intend to submit them as a PR, but the changes have been sufficiently useful to me to add them here.


app/src/main/java/net/gsantner/opoc/format/todotxt/SttTask.java
```diff
@@ -10,137 +10,343 @@
 #########################################################*/
 package net.gsantner.opoc.format.todotxt;
 
-import net.gsantner.opoc.preference.MapPropertyBackend;
+import android.widget.TextView;
```
harshad1, line 13: All Stt* functionality has been simplified and moved into `SttTask`. `SttCommander` etc are no longer used. The only purpose of the new SttTask is to parse todo.txt lines and extract projects, contexts etc. All modifications are done directly with `RegexReplace` in `TodoTxtTextActions.java`

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
```diff
@@ -394,7 +396,7 @@ public static void showPriorityDialog(Activity activity, char selectedPriority,
             availableData.add(Character.toString((char) i));
         }
         highlightedData.add(none);
-        if (selectedPriority != SttTask.PRIORITY_NONE) {
+        if (selectedPriority != 0) {
```
harshad1, line 399: 0 is the null character

app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
```diff
@@ -324,7 +324,7 @@ public Boolean onPreferenceClicked(Preference preference, String key, int keyId)
                     if (permc.doIfExtStoragePermissionGranted()) {
                         String sep = "\n";
                         if (appSettings.isTodoStartTasksWithTodaysDateEnabled()) {
-                            sep = SttCommander.getToday() + " ";
+                            sep = SttTask.getToday() + " ";
```
harshad1, line 327: Lots of SttCommander* functionality is moved to SttTask

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -62,10 +59,13 @@ protected ActionCallback getActionCallback(@StringRes int keyId) {
     @Override
     public List<ActionItem> getActiveActionList() {
 
+
+        final int projectIcon = _appSettings.isTodoTxtAlternativeNaming() ? R.drawable.ic_local_offer_black_24dp : R.drawable.ic_baseline_add_24;
```
harshad1, line 63: Chose icon based on naming

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -96,110 +96,72 @@ int getFormatActionsKey() {
         @Override
         public void onClick(View view) {
             view.performHapticFeedback(HapticFeedbackConstants.KEYBOARD_TAP);
-            final SttCommander sttcmd = SttCommander.get();
             final String origText = _hlEditor.getText().toString();
-            final int origSelectionStart = _hlEditor.getSelectionStart();
-            final SttTaskWithParserInfo origTask = sttcmd.parseTask(origText, origSelectionStart);
             final CommonTextActions commonTextActions = new CommonTextActions(_activity, _hlEditor);
-
-            final Callback.a1<SttTaskWithParserInfo> cbUpdateOrigTask = (updatedTask) -> {
```
harshad1, line 105: All of this callback is now unnecessary

app/build.gradle
```diff
@@ -41,16 +41,23 @@ android {
...
     flavorDimensions "default"
     productFlavors {
+
         flavorAtest {
             applicationId "net.gsantner.markor_test"
             versionCode = Integer.parseInt(new Date().format('yyMMdd'))
             versionName = new Date().format('yyMMdd')
             buildConfigField "boolean", "IS_TEST_BUILD", "true"
+
+            vectorDrawables.useSupportLibrary = true
```
gsantner, line 51: We should be able to strike this one, or at least I would prefer to remove it. I made bad experience with `vectorDrawables.useSupportLibrary = true` in the past.

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
gsantner: I would say one icon (the existing one) is fine, anyway the 99% will have the project&tag one.

gsantner:
@harshad1 
Thanks! I won't make any test/code look itself before weekend.

Can you please: 

* Revert to state of master branch for all Stt classes (net.gsantner.opoc)
* Move/Rename everything you done to net.gsantner.markor.format  namespace. You can use then some TodoTxtClass prefix instead of Stt for these.

Reason is, I want to keep the old classes as is. I did them in the past with a potential desktop client in mind, and don't want to couple it with any Android stuff.

gsantner:
mostly ready to take a look at? or something major left?

harshad1:
Should be ready to look at.

gsantner:
@harshad1 

Question, how would you add the change from PR1043 to the new implementation?

I'm just feeling like while it might be better, for me it's harder to understand what is happening exactly. Would it be enough to use just `x ` then in the first line at switch-case?

grafik

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -96,112 +94,78 @@ int getFormatActionsKey() {
...
             switch (_action) {
                 case R.string.tmaid_todotxt_toggle_done: {
-                    origTask.setDone(!origTask.isDone());
+                    String replaceDone = "x ";
                     if (AppSettings.get().isTodoAddCompletionDateEnabled()) {
-                        origTask.setCompletionDate(SttCommander.getToday());
+                        replaceDone += TodoTxtTask.getToday() + " ";
                     }
-                    cbUpdateOrigTask.callback(origTask);
+                    runRegexReplaceAction(
```
harshad1, line 107: @gsantner this is how we can implement #1043 here.

Note that this has an issue:

if I mark a task as done and then unmark it, the _creation date_ is removed as well.

i.e.

`(C) 2020-01-01 foo` -> mark done `x 2020-01-01 foo` -> unmark done -> `x 2020-01-01 foo`. 

This seems to be the behavior of master as well so I have left it as is here.

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -96,112 +94,78 @@ int getFormatActionsKey() {
...
+                            // If task starts with a priority, replace priority with 'x ...'
+                            new ReplacePattern(TodoTxtTask.PATTERN_PRIORITY_ANY, replaceDone),
+                            // else if task stars with 'x + (completion date)?' replace with "" (i.e. remove done)
+                            new ReplacePattern(TodoTxtTask.PATTERN_COMPLETION_DATE, ""),
+                            // else replace task start with 'x ...'
+                            new ReplacePattern("^", replaceDone)
```
harshad1, line 113: I have also added a comment to show how RegexReplaceAction is basically a sequence of `if` - `else if` - `else if`

gsantner:
@harshad1 
I've tried it, feedback:

Remove priority not works
screenrecord

Otherwise haven't found something obvious

gsantner:
I'm not 100% sure if everything works and no regression, but I'm sure we will be able to fix if something comes up. Overall I think it's for sure a big improvement.

Thanks for help, merging now.

PR #1034, opened 2020-08-20
Qualify shortcut by extension

harshad1:
If you have multiple files with the same name but different extensions, for example `project.md`, `project.log` etc, then it is not possible to create shortcuts to both. Any shortcut created will open one of them, seemingly at random.

Including the extension fixes this problem.


gsantner:
thx

PR #1059, opened 2020-09-12
Fix due date handling

harshad1:
My changes to TodoTxt apparently broke due date handling a little.
1. +3 days was broken
2. the pattern was wrong

This PR:
1. Fixed entry of due date entry and pattern
2. Removes TodoTxtHighlighterPattern.java (redundant)
3. Adds a clear due date option to the due date dialog


harshad1:
I am still testing this. If I do not find any issues in the next 2 days I will give the :+1 to merge.

Additionally @gsantner. Currently we have

short press -> insert date under cursor
long press -> add or update due date

I find that 99% of the time I am entering or changing a due date. I think we should therefore flip the short and long press i.e. we should change to:

short press -> add or update due date
long press -> insert date under cursor

what do you think?

gsantner:
Hm, I'm not using the date button that much, neither having due date anywhere. Unless it would be much much helpful to you I don't see the need of making the swap - beside users coming then to bugtracker/play-reports asking why it was swapped^^

gsantner:
hey @harshad1 

ready to merge? something related to last comment?

gsantner:
btw, I want to do "minor" release cycles faster. So release this with 2.4 these days

harshad1:
This is good to go. Didn't have internet access for a few days there...

gsantner:
ok then thx, merging

Issue #1064, opened 2020-09-24
Apply highlighting for todo.txt in search


harshad1:
The current highlighting for todo.txt is very useful for understanding a task at a glance. It would be very helpful if this highlighting were applied in the search UI as well.

PR #1065, opened 2020-09-24
highlighting in TodoTxt search dialog

harshad1:
This PR addresses #1064.


app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtHighlighter.java
```diff
@@ -38,80 +38,107 @@ public TodoTxtHighlighter(HighlightingEditor hlEditor, Document document) {
...
+                clearSpans(spannable);
+            }
+
+            if (spannable.length() == 0) {
+                return spannable;
+            }
+
+            final boolean isDarkBg = _appSettings.isDarkThemeEnabled();
 
             _profiler.restart("Context");
```
harshad1, line 91: Need to fix profiler so it works when separated

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -47,6 +47,23 @@ public TodoTxtTextActions(Activity activity, Document document) {
 
     @Override
     public boolean runAction(String action, boolean modLongClick, String anotherArg) {
+        if (action == CommonTextActions.ACTION_SEARCH ) {
+            final String origText = _hlEditor.getText().toString();
+
+            SearchOrCustomTextDialogCreator.showSearchDialog(_activity, origText,
```
harshad1, line 53: Should test with `instanceof` and consider adding a settings switch if performance is a concern.

app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
```diff
@@ -38,6 +40,7 @@
 import android.widget.ListView;
 import android.widget.TextView;
 
+import net.gsantner.markor.ui.hleditor.Highlighter;
```
gsantner, line 43: No Markor references in opoc. Especially this one as it is used for all my other apps as well.

gsantner:
I have to check it, not sure if really want full highlighting there at the dialog.

Maybe just key elements are enough

app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
harshad1: Fixed it will a callback instead

harshad1:
@gsantner Could you take a look now. I think it is good to go if you feel so.

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtHighlighter.java
```diff
@@ -38,80 +39,114 @@ public TodoTxtHighlighter(HighlightingEditor hlEditor, Document document) {
...
+            _profiler = profiler;
+        }
+        public void restart(final String ... optionalText) {
+            if (_profiler != null) {
+                _profiler.restart(optionalText);
+            }
+        }
+    }
+
+    public static Spannable basicTodoTxtHighlights(
```
harshad1, line 92: Broke todotxt highlighting into 2 functions. We can include only the highlighting we want in this second function.

Right now I have included everything except the padding and line separator/

app/src/main/java/net/gsantner/markor/ui/hleditor/Highlighter.java
```diff
@@ -15,7 +15,7 @@
 import android.graphics.Typeface;
 import android.support.annotation.ColorInt;
 import android.support.annotation.NonNull;
-import android.text.Editable;
+import android.text.Spannable;
```
harshad1, line 18: Most of these functions can be static and only require a Spannable (not an Editable). This change makes the Highlighter class more flexible.

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -47,6 +47,28 @@ public TodoTxtTextActions(Activity activity, Document document) {
...
+                    (spannable) -> {
+                        TodoTxtHighlighter.basicTodoTxtHighlights(
+                                spannable,
+                                true,
+                                new TodoTxtHighlighterColors(),
+                                _appSettings.isDarkThemeEnabled(),
+                                null
+                        );
```
harshad1, line 61: Using a callback instead of passing Markor into Opoc

app/src/main/java/net/gsantner/markor/format/keyvalue/KeyValueHighlighter.java
```diff
@@ -11,7 +11,7 @@
 
 import android.content.Context;
 import android.graphics.Typeface;
-import android.text.Editable;
```
harshad1, line 14: The only change in `MarkdownHighlighter.java`, `KeyValueHighlighter.java`, `PlainTextHighlighter.java` is Editable -> Spannable

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtHighlighter.java
```diff
@@ -38,80 +39,114 @@ public TodoTxtHighlighter(HighlightingEditor hlEditor, Document document) {
...
+        try {
+            if (clear) {
+                clearSpans(spannable);
+            }
+
+            if (spannable.length() == 0) {
+                return spannable;
+            }
+
+            ProfileCaller _profiler = new ProfileCaller(profiler);
```
gsantner, line 109: Please get rid of this profilecaller stuff. Just pass instance from texteditor.

And for search create instance there and just pass the same always. Profiler anyway not matters there.

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtHighlighter.java
```diff
@@ -38,80 +41,103 @@ public TodoTxtHighlighter(HighlightingEditor hlEditor, Document document) {
...
+            if (clear) {
+                clearSpans(spannable);
+            }
+
+            if (spannable.length() == 0) {
+                return spannable;
+            }
+
+            if (profiler == null) {
+                profiler = new NanoProfiler().setEnabled(BuildConfig.IS_TEST_BUILD || MainActivity.IS_DEBUG_ENABLED);
```
harshad1, line 99: Just create a NanoProfiler if null passed in

harshad1:
@gsantner Are there any other changes you would like to see in this PR?

Issue #524, opened 2019-03-24
File browser: Customization of Filename and description/date format


ubergeek77:
#### General information

* **App version:** 1.7.2 F-Droid
* **System:** Android 9.0

#### Description

I've searched around for this first, but I don't think this is has been requested before.

If you're a heathen like me, you're more used to the non-European date format, where the month comes before the Day (insane, right?) This is just what I'm accustomed to, so for me, this:

03/24/2019

Is easier for me to read at first glance than this:

24/03/2019

Markor doesn't appear to have any settings to change the displayed date of a note on the main screen, and it defaults to the day-month-year display format.

Could an option to switch to the other date format be considered for a future release?

Related feature request/suggestion (please let me know if I should file this as a separate issue):

If a note has a display date that is up to two or so days older than the current date, perhaps the display could show things such as "2 days ago," "Yesterday," or "Today." If it is older than two days old, it would just show the date as normal. I feel like this would be easier to read at a glance, especially if you have a good handful of notes that are not very old.

HoWilgh:
2019/03/24 would be easier for sorting. This is also the ISO version.... please see https://www.iso.org/iso-8601-date-and-time-format.html.

gsantner:
The filesystem/notebook view will be released with fully new layout.
While it won't allow to set format, it will always take  the default for date and time according to the current locale.

E.g. german:

![screenshot-2019-03-28__16-56-36](https://user-images.githubusercontent.com/6735650/55172816-91750e00-5172-11e9-95a1-87be0474e024.png)

gsantner:
The goal on this has changed sligthly. It's now to be able to configure the description line (where currently the date shows up.
With this extended way, the users should be able to configure both, custom date format as well e.g. show filesize. (`%YYYY-%MM-%DD %FS` for example)

gsantner:
Thought on the "blinking filename" to show full filename by horizontal animating filename: Makes sense to do that on-hover or on-selected. However by default no such scrolling, as otherwise everything blinky blinky the whole time and you cannot focus.

uheinema:
Thumbs up.
Or just show it on longpress? Maybe easier.
And Configurations are nice, but tend to implie GUi changes later, and a source of  problems by settimg them wrong.

gsantner:
no setting for that part.

PR #1078, opened 2020-10-06
Add file description customization for file browser entries (#524)

xuvatilavv:
Suggested in #524.

Added a settings entry to allow a customizable file description. If no format string is configured it uses the locale default date format like before. New format string uses the built-in formatting rules of SimpleDateFormat, just filters for a new entry FS and replaces it with a quoted string including the file size, then passes the format string to SimpleDateFormat directly.

I think it'd be nice if we could default the configured string to locale date format, but I didn't see a way to set that up without manually creating strings for all the formats for each locale. For now, default configuration is an empty string, which shows "Default" on the settings screen.

(Sorry for the cancelled PR, it's been a long time since I've done one and I'm remembering how it works 😅 )

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 ### Recent changes
+- 
+
+### v2.4 [Blog Post](https://gsantner.net/blog/2020/xx/yy/markor-v2.4-something.html?source=changelog)
 - Markdown: Correctly insert or remove list item on press enter at empty list item
+- Remove title from todo.txt date dialog, better usable on small devices
+- Fix search sometimes not working when chaging from view to edit mode
+- Rework Indent & Move lines Actions
+- Add settings option to control todo.txt completition date auto insert
+- All new todo.txt support in Markor
+- Add many languages to view-mode code highlighting
+- Add xlf format (plaintext)
+- Remove colored highlighting from changelog dialog
+- Add search to Markdown edit-mode outline/TOC dialog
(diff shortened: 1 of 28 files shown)


gsantner:
Hello thanks, I will look into it today or tomorrow after work.

Would you mind to 

* Remove AppSettings and pref_key__file_description_format reference from FilesystemViewerAdapter.java (only that file). I want to keep Markor specifics minimal there. Instead use directly SharedPreferences.getString("pref_key__file_description_format", "default")
* Remove the `*bytes_abbreviated` string ressources. No way they will be translated. Just have `String.format(Locale.getDefault(), "fmt", nr, "B");`
* Move the getHumanReadableByteCountSI function to FileUtils class. --> `public static String getHumanReadableByteCountSI(final long bytes)`

xuvatilavv:
Can do.

For the file size translation, I have found that various languages may use different abbreviations, particularly I've found that French uses "20Ko" instead of "20KB". The only major places I've been able to find that uses these translations are Apple devices and Xamarin, both via an NSByteCountFormatter function, so I'm not sure whether it's standard or expected. Would you still like to remove the possibility of translating them?

gsantner:
Yeah remove please, B/KB/MB/GB is fine.

xuvatilavv:
Alright, I've made the changes you requested.

gsantner:
Did various improvements, merging this now.

Thanks for your help! And congrats for your first-time merged contribution!

PR #1085, opened 2020-10-11
Apply app theme to launcher widget, closes #727

### LINKED ISSUE #727: Apply app theme to launcher widget
IMG_20191008_044002

el416:
Fixed the issue ( #727 ) with widget not changing font color properly when theme in the app was changed.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/SettingsActivity.java
+++ b/app/src/main/java/net/gsantner/markor/activity/SettingsActivity.java
@@ -229,13 +229,13 @@ public class SettingsActivity extends AppActivityBase {
                 activityRetVal = RESULT.RESTART_REQ;
                 _as.setRecreateMainRequired(true);
             } else if (eq(key, R.string.pref_key__app_theme)) {
-                restartActivity();
-                _as.setRecreateMainRequired(true);
                 // Handling widget color scheme
                 WrMarkorWidgetProvider.handleWidgetScheme(
                         getContext(),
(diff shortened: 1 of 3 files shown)


gsantner (commented):
Can you please keep Markor installed and widgets on desktop & try this for some days?  
Report back then if all is fine.

el416:
Will do. So far I found a bug where the widget would initially be displayed with white background when the app was set to a dark theme. 

I also found a problem, where after changing the app theme and using home button to go to home screen, widget wouldn't display correct text color until you clicked on it. I fixed both problems, but I had to use a finishactivity method instead of restartactivity in settingsactivity after changing app theme, since it was causing problems. 

I tested a few more scenarios and didn't find any more problems, so I think you can merge, but if you want, you can wait a few more days, though I doubt I will find anything else.

gsantner:
Past has shown that at widgets, issues mostly appears some days after

el416:
Okay then, I will use it for a few days and report the results to you afterwards.

el416:
Hey @gsantner , I've been using the app for ~4 days now, I kept the widget on desktop, and I didn't encounter any bugs. I tried a few different scenarios but it seems to be working fine.

el416:
Well I tried to also change the colors of the horizontal lines, but didn't manage to do it without needing to remove and re-add the widget to desktop, so I didn't include this change in the commit. Unfortunately, I can't work on any more improvements right now as I won't have much time to work on this, so I think you can merge the changes. If in the future I would want to work again on some of the widget issues, I will let you know.

PR #1087, opened 2020-10-12
Settings toggles to show/hide text actions

harshad1:
This PR adds support for enabling / disabling text actions. Actions are disabled by means of a switch in the reorder actions activity.

This is necessary as it would be good to add actions without cluttering up the actions bar.

image

TODO:
- [x] Determine if we want to remove the 'Enable' label
- [x] Add spacing between the switch and the drag icon

Future tasks:
- [ ] Disable text actions bar if all actions are disabled
- [ ] Add enable all and disable all options

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
@@ -515,13 +515,6 @@ public class DocumentEditFragment extends GsFragmentBase implements TextFormat.T
         _textFormat.getTextActions()
                 .setHighlightingEditor(_hlEditor)
                 .appendTextActionsToBar(_textActionsBar);
-        if (_textActionsBar.getChildCount() == 0) {
-            _textActionsBar.setVisibility(View.GONE);
-            ((HorizontalScrollView) _textActionsBar.getParent()).setVisibility(View.GONE);
-        } else {
-            _textActionsBar.setVisibility(View.VISIBLE);
-            ((HorizontalScrollView) _textActionsBar.getParent()).setVisibility(View.VISIBLE);
-        }
     }
 
     private void setupAppearancePreferences(View fragmentView) {
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
@@ -517,8 +517,10 @@ public class DocumentEditFragment extends GsFragmentBase implements TextFormat.T
                 .appendTextActionsToBar(_textActionsBar);
         if (_textActionsBar.getChildCount() == 0) {
             _textActionsBar.setVisibility(View.GONE);
+            ((HorizontalScrollView) _textActionsBar.getParent()).setVisibility(View.GONE);
         } else {
             _textActionsBar.setVisibility(View.VISIBLE);
+            ((HorizontalScrollView) _textActionsBar.getParent()).setVisibility(View.VISIBLE);
         }
     }
 
--- a/app/src/main/java/net/gsantner/markor/activity/ActionOrderActivity.java
+++ b/app/src/main/java/net/gsantner/markor/activity/ActionOrderActivity.java
@@ -9,6 +9,7 @@
 #########################################################*/
 package net.gsantner.markor.activity;
 
+import android.content.res.Resources;
 import android.graphics.Color;
 import android.os.Build;
 import android.os.Bundle;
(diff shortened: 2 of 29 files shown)


gsantner:
I would go for visible/invisible and its icon instead. For the reason, enable could also mean "still visible, but not clickable".

BeayemX:
Maybe use the terms Show/Hide?

gsantner:
I already started typing something along "OK".  
But as I thought more about it, I got into mind that like I remembered that likley the bad thing will happen that things overlap or are cut-off. As a person reading primary english you won't experience that, but on german for example you get "Anzeigen/Verstecken" instead of short "Show/Hide".  
...So, when making it Show/Hide it will be a issue when gets localized.

Unbenannt

Thus, use icons only, no text. If required, with long-press tooltip set like "toggle visiblity".

gsantner (commented):
I already started typing something along "OK".  
But as I thought more about it, I got into mind that likley the bad thing will happen with overlap and cut-offs. As a person reading primary english you won't experience that, but on german for example you get "Anzeigen/Verstecken" instead of short "Show/Hide".  
...So, when making it Show/Hide it will be a issue when gets localized.

Unbenannt

Thus, use icons only, no text. If required, with long-press tooltip set like "toggle visiblity".

harshad1:
@gsantner Good point.

I have removed the label and added a tooltip.

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -517,8 +517,10 @@ public void applyTextFormat(int textFormatId) {
                 .appendTextActionsToBar(_textActionsBar);
         if (_textActionsBar.getChildCount() == 0) {
             _textActionsBar.setVisibility(View.GONE);
+            ((HorizontalScrollView) _textActionsBar.getParent()).setVisibility(View.GONE);
         } else {
             _textActionsBar.setVisibility(View.VISIBLE);
+            ((HorizontalScrollView) _textActionsBar.getParent()).setVisibility(View.VISIBLE);
```
harshad1, line 523: I am not able to get the textactions bar to hide. There is always a band visible. Any suggestions here would be appreciated.

gsantner:
>  Disable text actions bar if all actions are disabled

Please don't do that. We already had that in the past and it happend to work bad on various "cheap" devices. Also textactions is a major feature of Markor..:D

harshad1:
> Please don't do that. We already had that in the past and it happend to work bad on various "cheap" devices. Also textactions is a major feature of Markor..:D
> 
> When all is unchecked and zero items to show, just have a empty bar.

:+1

harshad1:
This is ready for testing and merging

gsantner (changes requested):
@harshad1 

we have quite some duplication and unused things there:

```
private final Set<String> _disabled;



    private void saveNewOrder() {
        final ArrayList<String> disabledKeys = new ArrayList<>();
        _textActions.saveDisabledActions(new ArrayList<>(_adapter._disabled));


        public boolean getEnabled() {
            return _enabled.isChecked();
        }
```


grafik
grafik

Can you please remove duplicates / clean that?

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -515,11 +515,6 @@ public void applyTextFormat(int textFormatId) {
         _textFormat.getTextActions()
                 .setHighlightingEditor(_hlEditor)
                 .appendTextActionsToBar(_textActionsBar);
-        if (_textActionsBar.getChildCount() == 0) {
-            _textActionsBar.setVisibility(View.GONE);
-        } else {
-            _textActionsBar.setVisibility(View.VISIBLE);
-        }
```
harshad1, line 522: @gsantner Please note that I removed this existing code as it didn't seem to do anything useful.

gsantner:
does it also work when changing the format when being in the editor? I.e. do the correct actions show up in correct format, for new selected format?

harshad1:
Yes. Works in both the editor when changing formats and when switching tabs in the main screen.

Oct. 15, 2020 09:13:20 Gregor Santner <notifications@github.com>:

gsantner:
thx mergeed

gsantner:
@harshad1 
due new device, new markor installation.

I found that the Markdown actions order is different than was previously (random?). Can you check that?

harshad1:
@gsantner Hmm I think I see the issue

It is here
https://github.com/gsantner/markor/blob/master/app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java#L211-L212

This code has been like this for a while now (since the initial reorder actions change)

To fix this we just need to change the HashSets to ordered sets. I can do this in my search / replace pr or I can make a new PR.

gsantner:
however you prefer

PR #1095, opened 2020-10-18
Insert Date/Time text action: Recent time formats, closes #1092

### LINKED ISSUE #1092: Remember more than 1 recent time format
The time / date dialog is pretty useful for entering dates and times. Currently, only the most recent time format is remembered (in addition to the list of default formats). It would be helpful if the 'n' most recent formats were remembered.

harshad1:
This PR addresses #1092.

The most recent 5 time formats are remembered. All default formats are also available (but not duplicated with recent formats).

### DIFF
--- a/app/src/main/java/net/gsantner/markor/util/AppSettings.java
+++ b/app/src/main/java/net/gsantner/markor/util/AppSettings.java
@@ -722,8 +722,4 @@ public class AppSettings extends SharedPreferencesPropertyBackend {
     public File getFileBrowserLastBrowsedFolder() {
         return new File(getString(R.string.pref_key__file_browser_last_browsed_folder, getNotebookDirectoryAsStr()));
     }
-
-    public int getDateDialogRecentCount() {
-        return 5;
-    }
 }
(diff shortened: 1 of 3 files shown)


app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -722,4 +722,8 @@ public void setFileBrowserLastBrowsedFolder(File f) {
     public File getFileBrowserLastBrowsedFolder() {
         return new File(getString(R.string.pref_key__file_browser_last_browsed_folder, getNotebookDirectoryAsStr()));
     }
+
+    public int getDateDialogRecentCount() {
+        return 5;
```
harshad1, line 727: @gsantner Do you want to make this user-configurable? I didn't think so...

gsantner:
So this adds recent 5 to the existing dropdown spinner?

app/src/main/java/net/gsantner/markor/util/AppSettings.java
gsantner: No, just have the 5 where it's required / date, no appsetting method required

harshad1:
> So this adds recent 5 to the existing dropdown spinner, and keep the "most recent button"?

Yes. Note that entries will not be duplicated. If, for example, one of the default options is the last used format, it will be present in the top of the list but will not be repeated below. I.e. the list is an ordered set.

gsantner:
Ready to merge then?

harshad1:
99%. I'd like to test this for a few more hours though.

Oct. 18, 2020 11:38:54 Gregor Santner <notifications@github.com>:

harshad1:
@gsantner This is ready to merge

PR #1096, opened 2020-10-18
Add Actions: Move line up & down, start new line

harshad1:
Now that actions can be disabled, this PR adds 3 small actions which I have found to be very useful every day:

a. Actions to move one line up and down. This is already in the keyboard menu, but making it an action is useful for rearranging lists etc
b. Actions to insert newline after current line


app/src/main/res/values/strings.xml
```diff
@@ -188,6 +188,7 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="key_page_down">Page Down (Page Down)</string>
     <string name="move_text_one_line_up">Move text one line up</string>
     <string name="move_text_one_line_down">Move text one line down</string>
+    <string name="insert_next_line">Insert line below</string>
```
gsantner, line 191: to my understanding, it goes to end of the next line, and inserts nothing

harshad1: It goes to the end of the current line and inserts a new line. This is useful when editing lists or todo.txt. I can insert a line below with one click instead of having to find the line end and clicking enter. In markdown, the list will be automatically updated too.

If you have used vim, this is similar to the `o` command.

gsantner: vim-o Okay - I would name it "Start new line below"

harshad1:
This is good to go.

app/src/main/java/net/gsantner/markor/format/general/CommonTextActions.java
```diff
@@ -134,6 +138,20 @@ public boolean runAction(String action) {
...
+                moveLineBy1(true);
+                return true;
+            }
+            case ACTION_MOVE_DOWN: {
+                moveLineBy1(false);
+                return true;
+            }
+            case ACTION_NEW_LINE_BELOW: {
+                // Go to end of line, works with wrapped lines too
+                _hlEditor.setSelection(StringUtils.getLineEnd(_hlEditor.getText(), StringUtils.getSelection(_hlEditor)[1]));
```
gsantner, line 151: safe regarding Index out of bounds?

app/thirdparty/res/drawable/ic_baseline_arrow_downward_24.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
gsantner, line 1: add all android ic to main, not thirdparty

app/src/main/java/net/gsantner/markor/format/general/CommonTextActions.java
harshad1: Both `getLineEnd` and `setSelection` have defensive checks. I've been using this in my own branch for a long time now. Should be ok.

app/thirdparty/res/drawable/ic_baseline_arrow_downward_24.xml
harshad1: Moved

gsantner: grafik

@harshad1 
Are you sure you have commited that? Still shows up thirdparty here

app/src/main/res/values/string-not_translatable.xml
```diff
@@ -222,9 +222,9 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="tmaid_common_accordion" translatable="false">tmaid_common_accordion</string>
     <string name="tmaid_common_indent" translatable="false">tmaid_common_indent</string>
     <string name="tmaid_common_deindent" translatable="false">tmaid_common_deindent</string>
-    <string name="tmaid_common_toolbar_title_clicked_edit_action">tmaid_common_toolbar_title_clicked_edit_action</string>
-    <string name="tmaid_common_move_text_one_line_up">tmaid_common_move_text_one_line_up</string>
-    <string name="tmaid_common_move_text_one_line_down">tmaid_common_move_text_one_line_down</string>
+    <string name="tmaid_common_toolbar_title_clicked_edit_action" translatable="false">tmaid_common_toolbar_title_clicked_edit_action</string>
+    <string name="tmaid_common_move_text_one_line_up" translatable="false">tmaid_common_move_text_one_line_up</string>
+    <string name="tmaid_common_move_text_one_line_down" translatable="false">tmaid_common_move_text_one_line_down</string>
```
harshad1, line 227: Ah, thanks for catching this.

app/thirdparty/res/drawable/ic_baseline_arrow_downward_24.xml
harshad1: I think this may have been a problem with github's display. Should show correctly now

gsantner:
@harshad1 
Are you able to reproduce the issue?

Screenshot_20210305-123620

received this by email today

harshad1:
@gsantner

This is the behavior we've had since the beginning and it's expected.

Renumber ordered list does not change the very first, top level, number. This is because markdown starts numbering from this value so we don't want to change it.

```
1. Foo
2. Bar <--- move up
```
When #2 here is moved to the top the list starts with 2 and renumber ordered list will not change the value resulting in

```
2. Bar
3. Foo
```

It may be possible to fix this with lots of additional logic (there are several edge cases). But I've found it easiest to just tap the ordered list action 2x on the top level item in order to fix the numbering, after I'm done manipulating the list.

Mar. 5, 2021 03:37:39 Gregor Santner <notifications@github.com>:

PR #1088, opened 2020-10-12
(duplicate) Zim-Wiki support nr2 fredericjacob

gsantner:
from https://github.com/fredericjacob/markor/tree/zim-highlighting

other impl #1084 , issue #393

PR #1084, opened 2020-10-09
(duplicate) Zim-Wiki support nr1 pschwede

pschwede:
ref  Format: Zim Wiki #393 

---------

Hi. Sorry, my PR is premature. I asked @gsantner some question concerning getting my `ZimWikiTextConverter` on Matrix. In order answer it he needed some code to see and suggested to post a PR.

#### Issues
- [x] Links to other files like `[[PageLink]]` and `[[:NotebookLink]]` cause line to be completely invisible.
- [x] Syntax highlighting doesn't work for `== headings ==`.
- [x] Syntax highlighting doesn't work for `[[Links]]` (other than `[[web|links]]`).
- [ ] support `{{images.jpg}}`
- [ ] `__marked__` has to get a yellow background color (or similar according to theme)
- [ ] support tables
- [ ] support check boxes `[>]` and `[*]`
- [ ] support `'''Verbatim block'''`
- [x] support `_{Subscript}`
- [x] support `^{Superscript}`
- [x] ~~:question: make check boxes interactive in HTML view~~
- [ ] Turn ZimWikiTextActions into useful ones
   - [ ] Insert Zim Wiki Header action
- [ ] Take a look into ZimWikiAutoFormat
- [ ] (Optional) support `$${TeX}$$`

Issue #393, opened 2018-10-14
Format: Zim Wiki 


ghost:
#### General information

* **App version:** 1.3.0
* **System:** Lineageos microg

#### Description
I was wondering if it would be possible to add support for the zim wiki format. Zim wiki is a notetaking app on desktop. It would be really cool if this app could be the android counterpart to it, using services like syncthing to sync files between them. Here's the site for zim: http://zim-wiki.org

gsantner:
I'm not using/knowing the format myself, but the project welcomes additonal formats (and supports adding them)! Open for contributions!

Designboy:
I would appretiate this feature too. Would be very nice to edit my zim Notebooks. More about the Syntax can be found here. http://zim-wiki.org/manual/Help/Wiki_Syntax.html

gsantner:
@Designboy @ManchurianSector  @sojusnik @mherrmann3

Having some news about this ;) !
While there are no actions or other implementation, Zim Wiki (`.zim`) is now a new option in the new `New file dialog`. Also .zim files will now show up in the `Notebook`.

As of implementation, free to take. It's best if you guys start on it as you use the format, and know best whats expected and how! Waiting for merge requests :smile: 

![screenshot_20181208-042553](https://user-images.githubusercontent.com/6735650/49681414-843bf280-faa1-11e8-85d4-cddec16f2cb9.png)

trevor87:
First of all: Great job on markor! Its a really useful app!

Zim support would indeed be great as one could have ones wiki on the phone!

Just my two cents on this: I just saw the Zim Wiki option when creating a new note and tried it out and it turned out that it does not work yet. Wouldn't it be better then to take the option out then as long as it does nothing? For me it was quite confusing..

Greetings

gsantner:
@trevor87 
While youre right that there is no highlighting or other special features for it, you can create,edit and view zim files everywhere in Markor.

This currently uses normal plaintext editing functioanlities as you would work from any device.
More format specific funtionalities are planned, but is something that should be done by a interested zim user. (Or somebody that wants to be;).

trevor87:
Thanks for the quick answer!

Just a note that came up during my research: I checked my zim installation and it seems that ".zim" files are only project files that contain information about the notebook, but no content. Zim stores the content in ".txt" files.

silmaril42:
I think this might work very similar to the markdown format.
The following is a simple Zim document, in which I tried to gather all the possible formats.

```
[...]

PR #1098, opened 2020-10-18
Add ZimWiki Format, closes#393

fredericjacob:
UPDATE Oct 10th: This PR adds ZimWiki support (see #393). It includes highlighting of ZimWiki syntax elements in the editor, view mode and basic text actions.

~~This PR belongs to #393. It is currently based on #1084 and #1088, but could be rebased if needed. The PR adds some basic text actions for the most important formattings in ZimWiki files:~~

### General
- [x] Format added
- [x] File detection (.txt files with a zim header)
- [x] Zim format showcase notebook (demonstrates syntax elements and folder and link structure)
- [x] create new zim text files with automatically inserted header, title and creation date
- [x] more -> just add to this list...

### Syntax Highlighting (Editor)
- [x] General supported
- [x] coloring of headings
- [x] adjusted size of headings according to level (if activated in settings)
- [x] bold
- [x] italics
- [x] marked text (highlighted with a yellow background)
- [x] striked-through text
- [x] preformatted (monospaced) text inline
- [x] preformatted text multiple lines
- [x] unordered lists
- [x] ordered lists
- [x] links
- [x] superscript
- [x] subscript
- [x] checklists (different colors according to the check state)
- [x] more -> just add to this list...

### Parser / HTML Converter (View Mode)
- [x] General supported
- [x] Headlines
- [x] bold
- [x] italics
- [x] marked text (highlighted with a yellow background)
- [x] striked-through text
- [x] preformatted (monospaced) text inline
- [x] preformatted text multiple lines
- [x] unordered lists
- [x] ordered lists (except for letters)
- [x] links
- [x] superscript
- [x] subscript
- [x] checklists (only show checked items as checked, for now crossed and arrowed items are displayed as unchecked)
- [x] line breaks (line breaks in zim text files should be treated as such)
- [x] render locally stored images
- [ ] (optional) tables
- [x] more -> just add to this list...

### Actions
- [x] General supported
- [x] headings with level 1-5
- [x] bold
- [x] italic
- [x] code inline
- [x] strikethrough
- [x] unordered list
- [x] ordered list
- [x] checkboxes (toggle between unchecked, checked, crossed, arrow) 
- [x] indent/deindent with tabs instead of spaces
- [ ] (optional) insert link
- [ ] (optional) insert image
- [ ] (optional) show table of content
- [x] more -> just add to this list...

### AutoFormat
- [x] error-free auto-formatting (reuse markdown auto format where possible)
- [x] support automatic numbering of letters (lowercase and capital) in addition to numbers
- [ ] (optional) alternate between numbers and letters (like in zim desktop)
- [x] (optional) more to come...

## ToDos
Unordered priority list

- [x] Links to other files like `[[PageLink]]` and `[[:NotebookLink]]` cause line to be completely invisible.
- [x] Syntax highlighting doesn't work for `== headings ==`.
- [x] Syntax highlighting doesn't work for `[[Links]]` (other than `[[web|links]]`).
- [x] support `{{images.jpg}}`
- [x] support `_{Subscript}`
- [x] support `^{Superscript}`
- [x] ~~:question: make check boxes interactive in HTML view~~
- [x] Turn ZimWikiTextActions into useful ones
- [x] Insert Zim Wiki Header action
- [ ] (optional) support tables
- [ ] (Optional) support `$${TeX}$$`


gsantner:
@fredericjacob Can you please grant @pschwede permission to write/commit to your repository.  
Then you two please work together on this PR branch `fredericjacob:text-actions`.  OK? 👍🏻   

Rebase is not required, I will squash-merge the whole PR so history doesn't matter.

fredericjacob:
@gsantner That's good, so now we basically have our zim-wiki "master" branch (even if it's named "text-actions"). Should make collaborating easier.
@pschwede I invited you to my fork repository. :)

fredericjacob:
@gsantner Oh, I have a question to you regarding icons in markor. For the zim text actions I need not only H1-H3 headings, but also H4 and H5, for which there are no icons yet in the resources. Are the icons used from some particular (free) icon website where I could just get the two missing icons? Or were they designed specifically for markor?

gsantner:
The h1-h5 are from android material design, straigt from Android Studio -> new vector icon

gsantner:
I added the list of supported things, todos and status to the PR on top. Makes it easier to get some overview whats already there and what not

fredericjacob:
@gsantner Thanks, found the icons. For some reason, there were no heading icons in the list directly in Android Studio (so it makes sense that the h1-h3 are located in the "thirdparty" folder). But I found them on http://materialdesignicons.com/ (under Apache license, seem to be exactly the same icons as used in the markor repo).
Personally, I use quite a few h4 headings in my notebook pages and even some h5 ones, so I'd prefer to include these actions. But since the recent merge of #1087 with the possibility of hiding actions, a bloated bar shouldn't be a big issue, right? (I'm still going to add a setting for rearranging the zim icons, like for markdown.)

fredericjacob:
All of the basic text actions are implemented now! :)

@gsantner There are additional (somewhat more advanced) features/elements from zim wiki to be implemented, like image and inter-wiki-links, extensions like code boxes and tables, etc. But I'd rather stay with the basic support in this PR as the amount of changes is already quite big (which also increases the risk of merge conflicts over time...). After the merge, I (or someone else) can continue to work on more advanced features in separate PRs.
So next I'd like to focus on finishing and fixing the most important stuff like the view mode and to some extent auto-formatting. After this, I'd polish the code a bit and mark the PR as ready to be merged.
Are you okay with this plan?

@pschwede Are you still working on the view mode? If not, I can continue to work on the converter.

gsantner (commented):
Hey, here is my initial review of the current state.

Some things are easy fix (like code format), some likely will stay open for later review

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiTextActions.java
```diff
@@ -0,0 +1,239 @@
...
+                new ActionItem(R.string.tmaid_common_unordered_list_char, R.drawable.ic_list_black_24dp, R.string.unordered_list),
+                new ActionItem(R.string.tmaid_zimwiki_bold, R.drawable.ic_format_bold_black_24dp, R.string.bold),
+                new ActionItem(R.string.tmaid_zimwiki_strikeout, R.drawable.ic_format_strikethrough_black_24dp, R.string.strikeout),
+                new ActionItem(R.string.tmaid_zimwiki_italic, R.drawable.ic_format_italic_black_24dp, R.string.italic),
+                new ActionItem(R.string.tmaid_zimwiki_mark, R.drawable.ic_format_underlined_black_24dp, R.string.marked),
+                new ActionItem(R.string.tmaid_zimwiki_code_inline, R.drawable.ic_code_black_24dp, R.string.inline_code),
+                new ActionItem(R.string.tmaid_zimwiki_h1, R.drawable.format_header_1, R.string.heading_1),
+                new ActionItem(R.string.tmaid_zimwiki_h2, R.drawable.format_header_2, R.string.heading_2),
+                new ActionItem(R.string.tmaid_zimwiki_h3, R.drawable.format_header_3, R.string.heading_3),
+                new ActionItem(R.string.tmaid_zimwiki_h4, R.drawable.format_header_4, R.string.heading_4),
```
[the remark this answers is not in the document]
harshad1: It may be possible to default-disable them. But this will be more work.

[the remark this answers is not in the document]
gsantner: thought too, but "default at the end" is fine and easier

app/src/main/java/net/gsantner/markor/ui/hleditor/ReplacePatternGenerator.java
```diff
@@ -0,0 +1,33 @@
...
+public abstract class ReplacePatternGenerator {
+    protected List<TextActions.ReplacePattern> replaceNonSelectedPatternsWithSelectedOrReplaceWithAlternative(
+            final Pattern[] allSearchPatterns,
+            final Pattern selectedSearchPattern, final String selectedReplacement,
+            final String alternativeReplacement
+    ) {
+        List<TextActions.ReplacePattern> replacePatterns = new ArrayList<>();
+        for (final Pattern searchPattern : allSearchPatterns) {
+            String replacement;
+            if (searchPattern != selectedSearchPattern) {
```
harshad1, line 16: Note that the comparison here was for object identity. This code worked because the patterns passed into it were all  'const' `static final`. If the pattern is constructed this will not match correctly.

app/src/main/java/net/gsantner/markor/ui/hleditor/ReplacePatternGenerator.java
```diff
@@ -0,0 +1,33 @@
...
+            replacePatterns.add(new TextActions.ReplacePattern(searchPattern, replacement));
+        }
+        return replacePatterns;
+    }
+
+    protected List<TextActions.ReplacePattern> replaceOtherPrefixWithSelectedOrRemovePrefix(
+            final Pattern[] allPrefixPatterns,
+            final Pattern selectedPrefixPattern, final String selectedPrefixReplacement) {
+        String removePrefixReplacement = "$1";  // only keep whitespaces before prefix
+        return replaceNonSelectedPatternsWithSelectedOrReplaceWithAlternative(allPrefixPatterns,
```
harshad1, line 30: Shorter function names in this class may be better :). This is pretty unwieldy.

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownReplacePatternGenerator.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public List<TextActions.ReplacePattern> replaceWithOrderedListPrefixOrRemovePrefix() {
+        return replaceOtherPrefixWithSelectedOrRemovePrefix(PREFIX_PATTERNS, PREFIX_ORDERED_LIST, ORDERED_LIST_REPLACEMENT);
+    }
+
+    public List<TextActions.ReplacePattern> toogleQuote() {
+        return replaceNonSelectedPatternsWithSelectedOrReplaceWithAlternative(PREFIX_PATTERNS, PREFIX_QUOTE, ">$1 ", "");
+    }
+}
```
harshad1, line 85: Is a new base class and inheritance tree the best way here? If necessary to prevent duplication of `setHeadingAction` and `runPrefixReplaceAction` (both are pretty small functions) across zim and markdown, it may be better to just abstract out these helpers into static functions which are called appropriately.

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiReplacePatternGenerator.java
```diff
@@ -0,0 +1,121 @@
...
+        return replaceOtherPrefixWithSelectedOrRemovePrefix(PREFIX_PATTERNS, PREFIX_ORDERED_LIST, orderedListReplacement);
+    }
+
+    public List<TextActions.ReplacePattern> deindentOneTab() {
+        final Pattern leadingTabIndentPattern = Pattern.compile("^\t");
+        return Collections.singletonList(new TextActions.ReplacePattern(leadingTabIndentPattern, ""));
+    }
+
+    public List<TextActions.ReplacePattern> indentOneTab() {
+        final Pattern lineStart = Pattern.compile("^");
```
harshad1, line 118: Should consider making these class level `static final` declarations.

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiReplacePatternGenerator.java
```diff
@@ -0,0 +1,121 @@
...
+        return Collections.singletonList(new TextActions.ReplacePattern(lineStart, "\t"));
```
harshad1, line 119: If making lineStart etc static final at the class level is not possible, then `TextActions.ReplacePattern` has a constructor which will take 2 strings and call pattern.compile internally. No need to do this in 2 steps. Also, `regexReplaceAction` has an overload [which should do all this for you](https://github.com/gsantner/markor/blob/master/app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java#L356-L358).

fredericjacob:
Thank you @gsantner  and @harshad1 for the reviews, I will soon go through your comments!

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextActions.java
```diff
@@ -121,36 +103,33 @@ private boolean onClickImpl(final View view) {
...
                     final String listChar = _appSettings.getUnorderedListCharacter();
-                    final String uncheck = "$1" + listChar + " [ ] ";
-                    final String check = "$1" + listChar + " [x] ";
-                    runPrefixReplaceAction(PREFIX_UNCHECKED_LIST, uncheck, check);
+                    runRegexReplaceAction(replacePatternGenerator.toggleToCheckedOrUncheckedListPrefix(listChar));
                     return true;
                 }
                 case R.string.tmaid_common_ordered_list_number: {
-                    runPrefixReplaceAction(PREFIX_ORDERED_LIST, "$11. ", "$1");
+                    runRegexReplaceAction(replacePatternGenerator.replaceWithOrderedListPrefixOrRemovePrefix());
```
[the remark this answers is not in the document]
fredericjacob: At least for this refactoring, I tried out the markdown actions right after the change and didn't notice any behavior change. For regex replacements, I didn't change anything in the other formats, only for markdown.
A good thing about extracting this replacement logic to own classes is that now unit tests can be written for the action behaviors (which I already did for zim while the implementation) to avoid breaking code here in future revisions.

But yeah, it's a good idea to go through the (changed) formats once again at the end to verify that they still work as expected, I will do this.

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiHighlighter.java
```diff
@@ -0,0 +1,133 @@
...
+import net.gsantner.markor.model.Document;
+import net.gsantner.markor.ui.hleditor.Highlighter;
+import net.gsantner.markor.ui.hleditor.HighlightingEditor;
+import net.gsantner.markor.util.AppSettings;
+
+import java.util.regex.Pattern;
+
+import other.writeily.format.zimwiki.WrZimWikiHeaderSpanCreator;
+
+public class ZimWikiHighlighter extends Highlighter {
```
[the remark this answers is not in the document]
fredericjacob: Oh, I wasn't sure about the writeily directory, it is not copy-pasted or unused... I made some refactoring in `writeily` to `WrMarkdownHeaderSpanCreator`, extracted logic to a new class `WrProportionalHeaderSpanCreator` and reused this for a new `WrZimHeaderSpanCreator` (needed for adjusting heading sizes for zim wiki syntax).
So what are the best practices there? Should the writeily directory stay untouched, should I move files to the markor directory instead? Or something else?

app/src/main/java/net/gsantner/markor/format/zimwiki/ListHandler.java
```diff
@@ -0,0 +1,85 @@
...
+package net.gsantner.markor.format.zimwiki;
+
+import android.text.Editable;
+import android.text.Spannable;
+import android.text.Spanned;
+import android.text.TextWatcher;
+
+import net.gsantner.opoc.util.StringUtils;
+
+public class ListHandler implements TextWatcher {
```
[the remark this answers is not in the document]
fredericjacob: Yes, it's a leftover from the initial copy-pastes, will still be removed/changed/fixed.

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiHighlighterPattern.java
```diff
@@ -0,0 +1,39 @@
...
+    CHECKLIST_UNCHECKED(Pattern.compile("(?<=(\\n|^))\t*(\\[)( )(])(?= )")),
+    CHECKLIST_CHECKED(Pattern.compile("(?<=(\\n|^))\t*(\\[)(\\*)(])(?= )")),
+    CHECKLIST_CROSSED(Pattern.compile("(?<=(\\n|^))\t*(\\[)(x)(])(?= )")),
+    CHECKLIST_ARROW(Pattern.compile("(?<=(\\n|^))\t*(\\[)(>)(])(?= )")),
+    SUBSCRIPT(Pattern.compile("(_\\{(?!~)(.+?)\\})")),
+    SUPERSCRIPT(Pattern.compile("(\\^\\{(?!~)(.+?)\\})")),
+    ZIMHEADER(Pattern.compile("^Content-Type: text/x-zim-wiki(\r\n|\r|\n)" +
+            "Wiki-Format: zim \\d+\\.\\d+(\r\n|\r|\n)" +
+            "Creation-Date: \\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}[.+:\\d]+"));
+
```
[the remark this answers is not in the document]
fredericjacob: No, it was introduced to highlight (gray out) the header lines of zim wiki text files.

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiReplacePatternGenerator.java
```diff
@@ -0,0 +1,121 @@
...
+import net.gsantner.markor.ui.hleditor.TextActions;
+import net.gsantner.opoc.util.StringUtils;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.regex.Pattern;
+
+public class ZimWikiReplacePatternGenerator extends ReplacePatternGenerator {
```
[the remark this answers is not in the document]
fredericjacob: No, this is where the replacement logic for Zim-specific text actions goes on (different from markdown).

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiTextConverter.java
```diff
@@ -0,0 +1,233 @@
...
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Wrapper class around MarkdownTextConverter
+ */
+@SuppressWarnings("WeakerAccess")
+public class ZimWikiTextConverter extends net.gsantner.markor.format.TextConverter {
```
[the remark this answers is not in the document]
fredericjacob: This started from copy-paste initially, but was already to some extent adjusted to zim (and needs a zim-specific implementation for rendering the view mode). I will look at it and continue working on it when I'm done with auto-formatting.

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiTextConverter.java
```diff
@@ -0,0 +1,233 @@
...
+        // TODO Table
+
+        private Pattern pattern;
+
+        ZimWikiPatterns(Pattern pattern) {
+            this.pattern = pattern;
+        }
+    }
+
+    public ZimWikiTextConverter(MarkdownTextConverter converter) {
```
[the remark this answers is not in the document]
fredericjacob: Same like for the general comment on the file, I will fix this when I start working on the converter and view mode.

app/src/main/java/net/gsantner/markor/util/DocumentIO.java
```diff
@@ -127,8 +130,8 @@ public static synchronized Document loadDocument(Context context, Bundle argumen
                 document.setFormat(TextFormat.FORMAT_KEYVALUE);
             } else if (TextFormat.CONVERTER_MARKDOWN.isFileOutOfThisFormat(fnlower)) {
                 document.setFormat(TextFormat.FORMAT_MARKDOWN);
-            } else if (fnlower.endsWith(".txt") || fnlower.endsWith(".zim")) {
-                document.setFormat(TextFormat.FORMAT_PLAIN);
+            } else if (isZimWikiFile(fnlower, document)) {
```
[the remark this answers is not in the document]
fredericjacob: This was a not-so-easy design decision I made:
The problem is that for zim wiki text files, it's not enough to look at the filename. Instead, the content has to be examined to find out if the zim header is present there or not.
`TextFormat.CONVERTER_ZIMWIKI.isFileOutOfThisFormat` only gets the filepath as an argument. For all other formats, checking the ending is enough, but for zim not. So in order to use the method, the file has to be opened and checked. However, I found out that the passed name is not necessarily the full file path, then causing an error when attempting to open it. 
So instead of digging deeper in the base code and making potentially harmful changes in the existing code, I decided to just introduce the `isZimWikiFile` method. Advantage here: the content is readily available in the `Document` (also better performance-wise) and can easily be checked for the header.

app/src/main/java/net/gsantner/markor/util/DocumentIO.java
```diff
@@ -148,6 +151,16 @@ public static synchronized Document loadDocument(Context context, Bundle argumen
         return document;
     }
 
+    private static boolean isZimWikiFile(String filename, Document document) {
```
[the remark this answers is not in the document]
fredericjacob: See comment above.

app/src/main/res/xml/preferences_master.xml
```diff
@@ -498,7 +498,28 @@
...
         </PreferenceScreen>
 
-        <!-- To-Do Screen -->
+        <!-- ZimWiki screen -->
+        <PreferenceScreen
+            android:icon="@drawable/fountain_pen"
+            android:title="@string/zim_wiki">
+
+            <PreferenceCategory android:title="@string/syntax_highlighting">
+                <CheckBoxPreference
```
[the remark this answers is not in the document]
fredericjacob: Yes, it is implemented, see `WrZimWikiHeaderSpanCreator`.

app/thirdparty/java/other/writeily/format/markdown/WrMarkdownHeaderSpanCreator.java
```diff
@@ -9,55 +9,35 @@
...
 
-import net.gsantner.markor.format.markdown.MarkdownHighlighter;
+import net.gsantner.markor.ui.hleditor.Highlighter;
 import net.gsantner.markor.ui.hleditor.SpanCreator;
 
 import java.util.regex.Matcher;
 
+import other.writeily.format.WrProportionalHeaderSpanCreator;
+
 public class WrMarkdownHeaderSpanCreator implements SpanCreator.ParcelableSpanCreator {
```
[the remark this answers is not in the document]
fredericjacob: See above my answer above to the other writeily import in `ZimWikiHighlighter`. I wasn't sure about how to "properly" do this refactoring so as to keep a consistent project structure.

app/thirdparty/java/other/writeily/format/zimwiki/WrZimWikiHeaderSpanCreator.java
```diff
@@ -0,0 +1,49 @@
...
+import android.text.Spannable;
+
+import net.gsantner.markor.format.zimwiki.ZimWikiHighlighter;
+import net.gsantner.markor.ui.hleditor.SpanCreator;
+
+import java.util.regex.Matcher;
+
+import other.writeily.format.WrProportionalHeaderSpanCreator;
+
+public class WrZimWikiHeaderSpanCreator implements SpanCreator.ParcelableSpanCreator {
```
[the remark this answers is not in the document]
fredericjacob: It is needed for the bigger headings, see above my answers to the writeily refactoring.

app/src/main/java/net/gsantner/markor/ui/hleditor/ReplacePatternGenerator.java
```diff
@@ -0,0 +1,33 @@
...
+            } else {
+                replacement = alternativeReplacement;
+            }
+            replacePatterns.add(new TextActions.ReplacePattern(searchPattern, replacement));
+        }
+        return replacePatterns;
+    }
+
+    protected List<TextActions.ReplacePattern> replaceOtherPrefixWithSelectedOrRemovePrefix(
+            final Pattern[] allPrefixPatterns,
```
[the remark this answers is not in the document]
fredericjacob: I changed this, but I would argue that it would be easier to read on multiple lines... Or is this against some convention?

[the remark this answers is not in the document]
gsantner: I'm at least not aware of any multi-line method deceleration in the project

app/src/main/res/values/strings.xml
```diff
@@ -362,8 +362,11 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="heading_1">Heading 1</string>
     <string name="heading_2">Heading 2</string>
     <string name="heading_3">Heading 3</string>
+    <string name="heading_4">Heading 4</string>
+    <string name="heading_5">Heading 5</string>
     <string name="bold">Bold</string>
     <string name="italic">Italic</string>
+    <string name="marked">Marked</string>
```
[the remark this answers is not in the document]
fredericjacob: Changed this to "highlighted", and I agree that this makes the meaning more clear and is less confusing. However, I'd like to point out that in Zim Desktop the corresponding button is indeed named "mark", so zim users might be more familiar with it.

app/src/main/java/net/gsantner/markor/ui/hleditor/ReplacePatternGenerator.java
fredericjacob: Yes its pretty long, I tried hard to find a name which describes the intent of the method well, but couldn't come up with a good short one. Do you have an idea/suggestion?

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiReplacePatternGenerator.java
fredericjacob: Changed this to use the two-string constructor.

The other thing is probably a matter of taste, but in some cases (like here) I prefer longer but more obvious code over shorter one. Especially when it comes to meaningful method and variable names. So instead of just using the overload and calling `regexReplaceAction("^\t", "")` directly in `ZimWikiTextActions`, the method `deindentOneTab` states much more clearly what the intended effect of the action is.

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownReplacePatternGenerator.java
fredericjacob: Changed the methods to be static, this makes sense - there is not even any data hold to be manipulated.
Using a `ReplacePatternGeneratorHelper` with helping functions now.

app/src/main/java/net/gsantner/markor/ui/hleditor/ReplacePatternGenerator.java
fredericjacob: Good point, it still worked because the passed patterns stayed unique, but this makes the function error-prone. Fixed it.

fredericjacob:
@gsantner @harshad1 I went through your reviews, changed many things you mentioned (including all formatting-related stuff) and answered your questions. Apart from the comments to things which will be fixed/worked on later, I left some open where I could need some more help/clarification. :)

app/src/main/java/net/gsantner/markor/format/zimwiki/ListHandler.java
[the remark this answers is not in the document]
fredericjacob: With the recent refactoring, there's only one ListHandler class used for both zim and markdown.

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiTextConverter.java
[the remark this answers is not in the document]
fredericjacob: Changed, however MarkdownTextConverter is still used as an intermediate converter due to a native ZimWiki-HTML-Renderer. So zim syntax is converted to markdown first and then rendered.

fredericjacob:
@gsantner The PR is complete now and ready for the final review (and being merged then)! :)

The bigger things I worked on since the last reviews are (see also the feature checklist in the PR description):
* auto-formatting:
   * refactored `MarkdownAutoFormat` (extracted most code to a class `AutoFormatter`) to reuse the logic for both markdown and zim, a single (parametrized) `ListHandler` for both formats
   * implemented zim-specific auto-formatting
* creation of new zim text files:
   * page title is constructed from the file name (for the filename, underscores are used instead of spaces, like in the zim desktop application)
   * automatic insertion of the zim header and creation date
* `ZimWikiTextConverter`/ view mode:
   * I continued with @pschwedes initial approach to translate zim syntax elements into markdown first and then reuse the `MarkdownTextConverter` (which makes use of the flexmark-java parser) for creating the html view. I didn't find a java/javascript library for direct rendering of zim files to html. So I think for now this is the best and easiest solution, which works quite well, especially because certain html elements can be inserted manually if needed.
   * support for all basic elements, including local images
   * some current restrictions: letter ordering is replaced with numbers, crossed and arrowed checkboxes are displayed as unchecked
* added a demo zim notebook to `samples` which includes a few subpages, an image, the constructs from the previous zim sample file and some interwiki links
   * can be used as a reference for future development

Some current restrictions which can be overcome with future PRs:
* interwiki links don't work yet, could also need corresponding text actions to insert them
* some advanced syntax elements (like tables, advanced code blocks, ...) are not yet supported

Here are some screenshots to demonstrate the current state (using the sample notebook):

## Editor
<div style="display:flex;">




</div>

## View Mode
<div style="display:flex;">




</div>

## New Files
<div style="display:flex;">


</div>

gsantner:
@fredericjacob 
Great, thank you very much! Sorry I missed this notification somehow.

Has something come up in the meantime already or still "ready"?

gsantner:
@fredericjacob 
Great, thank you very much! Sorry I missed this notification somehow.

Has something come up in the meantime already or still "ready"?

Would also kindly ask @harshad1  and @pschwede to join reviewing

harshad1:
I will be able to review this over the weekend

gsantner:
@fredericjacob 
Is the document from the screenshots this one <https://github.com/gsantner/markor/blob/master/samples/zim-wiki.txt> or have you worked that out? If it's improved/updated/newer, would you mind to share that as well?  You can commit over that file

gsantner:
screenshot-2020-11-14__13-00-56

* feedback - gs1

markor dark theme: Color for highlighting doesn't work in this case. I suggest to use a darkish-yellow/brown in that case. (Instead of keeping yellow for dark mode, and adjusting text foreground color, because that costs performance twice)

* 

* ✔️  feedback - gs2

color for links not much readable. We already have a specific blue color for links that used in ALL formats (todo/plaintext/markdown), so just gonna swap it to it (taking care of it)

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiTextConverter.java
```diff
@@ -0,0 +1,215 @@
...
+            markdownContent.append(markdownEquivalentLine);
+            markdownContent.append("  "); // line breaks must be made explicit in markdown by two spaces
+            markdownContent.append(String.format("%n"));
+        }
+
+        return _markdownConverter.convertMarkup(markdownContent.toString(), context, isExportInLightMode, file);
+    }
+
+    private String getMarkdownEquivalentLine(String zimWikiLine, final boolean isExportInLightMode) {
+        _currentLine = zimWikiLine;
```
gsantner, line 74: Is it required to cache the line here? Should be un-needed duplication that blows up memory size (depending on amount of text)

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiHighlighter.java
```diff
@@ -0,0 +1,145 @@
...
+    private final boolean _highlightBiggerHeadings;
+    public final String _fontType;
+    public final Integer _fontSize;
+
+    private static final int COLOR_HEADING = 0xff4e9a06;
+    private static final int HIGHLIGHT_BACKGROUND_COLOR_LIGHT_MODE = 0xffffff00;
+    private static final int HIGHLIGHT_BACKGROUND_COLOR_DARK_MODE = 0xffFFA062;
+    private static final int UNORDERED_LIST_BULLET_COLOR = 0xffdaa521;
+    private static final int ORDERED_LIST_NUMBER_COLOR = 0xffdaa521;
+    private static final int LINK_COLOR = 0xff1ea3fd;// 0xff0000ff;
```
gsantner, line 37: * ✔️  feedback - gs1

color for links not much readable. We already have a specific blue color for links that used in ALL formats (todo/plaintext/markdown), so just gonna swap it to it (taking care of it)

c

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiHighlighter.java
```diff
@@ -0,0 +1,145 @@
...
+
+import other.writeily.format.zimwiki.WrZimWikiHeaderSpanCreator;
+
+public class ZimWikiHighlighter extends Highlighter {
+    private final boolean _highlightBiggerHeadings;
+    public final String _fontType;
+    public final Integer _fontSize;
+
+    private static final int COLOR_HEADING = 0xff4e9a06;
+    private static final int HIGHLIGHT_BACKGROUND_COLOR_LIGHT_MODE = 0xffffff00;
```
gsantner, line 33: * feedback - gs2

markor dark theme: Color for highlighting doesn't work in this case. I suggest to use a darkish-yellow/brown in that case. (Instead of keeping yellow for dark mode, and adjusting text foreground color, because that costs performance twice)

Thought: `0xff_FFA062`, kinda yellow-orang'ish-brown'ish

screenshot-2020-11-14__13-19-52

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiHighlighter.java
gsantner: I took care of it, just screenshot for reference

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownReplacePatternGenerator.java
```diff
@@ -0,0 +1,94 @@
...
+
+    public static List<TextActions.ReplacePattern> replaceWithUnorderedListPrefixOrRemovePrefix(String listChar) {
+        final String unorderedListReplacement = "$1" + listChar + " ";
+        return ReplacePatternGeneratorHelper.replaceOtherPrefixWithSelectedOrRemovePrefix(PREFIX_PATTERNS, PREFIX_UNORDERED_LIST, unorderedListReplacement);
+    }
+
+    public static List<TextActions.ReplacePattern> toggleToCheckedOrUncheckedListPrefix(String listChar) {
+        final String unchecked = "$1" + listChar + " [ ] ";
+        final String checked = "$1" + listChar + " [x] ";
+        return ReplacePatternGeneratorHelper.replaceNonSelectedPatternsWithSelectedOrReplaceWithAlternative(PREFIX_PATTERNS, PREFIX_UNCHECKED_LIST, unchecked, checked);
```
harshad1, line 84: Please use shorter function and class names. This is quite hard to read.

fredericjacob:
Thank you for the reviews, I will go over them when I find some time.

@gsantner I used the extended app a lot in the last weeks and I didn't notice any real bugs so far. But there are one or two small things that could be improved a bit, probably I'll fix them too when I work on the review stuff.
And regarding the document from the screenshots: it is part of the sample notebook I created (see my last post). It's already checked in, you can find the corresponding folder in the "samples" directory.

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiHighlighter.java
fredericjacob: The orange color is fine, changed it to use it exclusively for both light and dark mode.

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiTextConverter.java
fredericjacob: Yes, introducing the field variable makes it easier to manipulate the content of the line in the other methods.
Actually, it does not duplicate the content of the string and blow up memory (except for an additional pointer) because Strings are passed by reference in Java like any other objects, not by value like primitive types - so the two variables refer to the same content in memory.

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownReplacePatternGenerator.java
fredericjacob: Changed, using shorter names now. Also added some JavaDoc to make the methods easier to understand.

app/src/main/java/net/gsantner/markor/ui/hleditor/ReplacePatternGenerator.java
fredericjacob: Changed.

fredericjacob:
I went through the reviews and changed the code accordingly or commented back.

@gsantner There are still some of your comments open from the previous review where I commented back, could you look at these and give me additional feedback or resolve them?

I also fixed few problems that I encountered when using the app. I'm currently still working on two other small issues:
- the cursor is not placed between the equal signs left and right after creating/toggling a heading
- ordered lists with indented levels (e.g. 1., 2., a., b., 3.) are reordered wrongly when auto-formatting kicks in, due to the way indentation works for markdown - letters are overriden with numbers and automatic enumeration doesn't stop before the first indented item

gsantner:
I would not mind a/b/i/ii/iii for ordered lists too much,   1/2/3/4 is fine

app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
```diff
@@ -155,12 +156,24 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
...
-                            if (f.exists() && f.length() < ShareUtil.MIN_OVERWRITE_LENGTH && templateContents != null) {
+                            boolean isAllowedToWrite = f.exists() && f.length() < ShareUtil.MIN_OVERWRITE_LENGTH;
+                            if (isAllowedToWrite && templateContents != null) {
                                 arg_fos.write(templateContents);
+                            } else {
+                                if (isAllowedToWrite && isNewZimWikiFile) {
+                                    String headerAndTitleContents = ZimWikiFileContentsCreator.createZimWikiHeaderAndTitleContents(fileNameWithoutExtension, new Date(), getResources().getString(R.string.created));
+                                    arg_fos.write(getBytesFromStringContents(encryptCheckbox.isChecked(), headerAndTitleContents));
+                                }
+
```
gsantner, line 176: @fredericjacob 
I'm not really fond of these changes. We already have a method that is soley there for handling templates.  
What I suggest is: 

* getTemplateContent()::::: Add case 8 (Empty Zim file). Return whatever you want there in `case 8: {`, but don't write any file. Thats job of this overall, not of the `ZimWikiFileContentsCreator`.
* typeSpinner.setOnItemSelectedListener():::::: When Zim selected, select template Empty Zim file (once). User may select a different template afterwards, thats up to the user. Like the advanced Zim template.

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiTextConverter.java
gsantner: I handled this.

app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
gsantner: I handled this.

gsantner:
to reduce review time, I've ahead merged the sample text files from the samples directory. also merged master into this one and fixed merge conflicts.

gsantner:
also reduced amount of other unnecessary changes, and drastically reduced amount of "new files" by merging the hl unit tests into subclasses of one file

fredericjacob:
Hi @gsantner , thanks for reviewing! Unfortunately, I have very little time available currently, also had a full weekend. But I'll try to go over your comments tomorrow evening

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiTextActions.java
```diff
@@ -0,0 +1,257 @@
...
+                    _hlEditor.getText().insert(_hlEditor.getSelectionEnd(), "\n'''\n");
+                    _hlEditor.setSelection(c + "\n'''\n".length());
+                    _hlEditor.enableHighlighterAutoFormat();
+                    Toast.makeText(_activity, R.string.code_block, Toast.LENGTH_SHORT).show();
+                    return true;
+                }
+                case R.string.tmaid_common_ordered_list_number: {
+                    // TODO: adapt to zim wiki
+                    AutoFormatter.renumberOrderedList(_hlEditor.getText(), StringUtils.getSelection(_hlEditor)[0], ZimWikiAutoFormat.getPrefixPatterns());
+                }
```
harshad1, line 225: Please see the code recently added here: https://github.com/gsantner/markor/pull/1129/files#diff-bbc685651c09c955da0b077fe835a688483119842448adedfa7296e0a8920b68R256-R262.

You will want to include this here

app/src/main/java/net/gsantner/markor/format/AutoFormatter.java
```diff
@@ -0,0 +1,368 @@
...
+
+        private final PrefixPatterns prefixPatterns;
+
+        public final boolean isOrderedList;
+        public final char delimiter;
+        public final int numStart, numEnd;
+        public final int groupStart, groupEnd;
+        public final String value;
+
+        public OrderedListLine(CharSequence text, int position, PrefixPatterns prefixPatterns) {
```
harshad1, line 148: Why are we passing the search patterns around? Is it because zim and markdown have different Ordered list etc formats and you want to select behavior by passing in the pattern-set.

If so, I would suggest use 'format' or 'type' in the name. `class FormatPatterns` or something like that to show that this is format specific.

gsantner:
Sorry for little missusing this PR to also make use new CI system (GitHub actions).  
But as you can see, it already reports useful stuff to us.  
Specifically, the unit test with zim file generation is failing. For me locally it succeedes, probably also for @fredericjacob. Very likeley this id due we are in same time zone? 😉 . So that one needs a fix.

grafik

fredericjacob:
> Sorry for little missusing this PR to also make use new CI system (GitHub actions). It comes handy due bigger changes, and new test cases wink .
> But as you can see, it already reports useful stuff to us.
> Specifically, the unit test with zim file generation is failing, so that one needs a fix.
> 
> grafik

Good to have the CI and to see that there is an issue... On my local machine, the test runs fine.
However the fix in 09655d6 just ignores the desired date-time-format completely and thus hides the issue - the time string should include the timezone and have the format "18:00:30+01:00", not "17:00:30Z" which should be achieved with the SimpleDateFormat string ending on "XXX"...
@gsantner Does the test fail on your local machine too (before revision 09655d6)?

gsantner:
Yes it was failing on my computers and on CI.

fredericjacob:
Hmm, that's strange, I cannot reproduce this, even on the previous revision the test passes for me.
I also wonder why the CI doesn't complain on commit 97688ec, but on the next master merge with 3dde5af it does as there were no changes to the zim wiki code.
Was there some change introduced in the master recently which might have changed the way how date formats are handled? Or some changes to the used Android API?
In the class `ZimWikiTextActions` I use the line `SimpleDateFormat headerDateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssXXX", Locale.ROOT);` for specifying how the Date-Time-String should look like. According to https://developer.android.com/reference/java/text/SimpleDateFormat#date-and-time-patterns, "XXX" should produce the desired result (and in fact, it does for me).
So any ideas why it might fail on the CI (or other local development environments) are very welcome...

gsantner:
it also fails local for me as said, not only on CI. So it may have todo with your computer TZ being the exact working one? :D

gsantner:
I want to like to include Zim in this years last update. Would be cool if we can finish up by next week :D

fredericjacob:
> it also fails local for me as said, not only on CI. So it may have todo with your computer TZ being the exact working one? :D
> 
> I added fail detection and error handling step by step to CI, the test part was just not finished at that point.

Hm okay, then I'll have to take a closer look at this in the future. For now, I'll just adjusted the test to ignore the time part.
I also checked in the app if the timestamp is created there correctly if I create a zim wiki file - which is the case.
Could you create a new zim wiki file in the app and check the header there? If the creation date there has the form "...18:00:30+01:00" and not something like "...17:00:30Z", then everything is correct and the bug only seems to occur in the test code.

> I want to like to include Zim in this years last update. Would be cool if we can finish up by next week :D

I will do my best. I still went over most of your changes and comments and collected things to work on, I will gradually fix/comment on them. My holiday starts on thursday next week and I'll have some free time then to work on Markor. So we should be able to be done with the important stuff at the end of the week.

gsantner:
please keep .zim.txt, I set that intentionally. It makes clear the file is about zim.

app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
fredericjacob: @gsantner I understand your point here. However, these changes also remove the feature I introduced: that typed file names for new pages are automatically adjusted so that the files are always compatible with Zim Desktop Wiki.
In Zim Desktop, the created text file for a new page with the title "My Page" automatically gets the file name "My_Page.txt". That's why I implemented the logic to replace spaces with underscores for the file name, so that the end user does not have to care about the internals of file naming in Zim. If a user types a name that doesn't conform this naming rule, compatibility with the desktop application will break.

If your okay with this, I will introduce a method that determines the file name to use (for zim wiki) instead of the plain calls of `fileNameEdit.getText()`. I think this will prevent users from quite some confusion why their notebook pages created with Markor are not shown in the desktop application.

fredericjacob:
> please keep .zim.txt, I set that intentionally. It makes clear the file is about zim.

This will break compatibility with Zim Desktop at least to some extent. Zim pages follow certain naming rules so that they're displayed correctly in the wiki tree. The text file corresponding to a page with the name "My Page" will always be named "My_Page.txt", and the corresponding folder which contains subpages (if there are any) "My_Page". If a file is named "My_Page.zim.txt", the page is still visible but with the name "My Page.zim". Maybe there are other hidden bugs which would be introduced by such a naming.
In any case, this would be annoying to all users who want to cross-use their zim notebooks on desktop and mobile, so I'd strongly suggest to keep the pure .txt endings.

gsantner:
@fredericjacob @harshad1 
I want to merge this today. Please if you are aware of something important, please note/fix it. This will get into the next update that goes to google play, fdroid & github.  

Otherwise we have to fix later.

harshad1:
@gsantner @fredericjacob The code for setting indent with long press is missing in this PR.

as mentioned in earlier comment: [this](https://github.com/gsantner/markor/pull/1129/files#diff-bbc685651c09c955da0b077fe835a688483119842448adedfa7296e0a8920b68R256-R262) has to be added to the long click text actions for zim

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiTextActions.java
```diff
@@ -0,0 +1,271 @@
...
+                    // TODO: adapt to zim wiki
+                    final String origText = _hlEditor.getText().toString();
+                    SearchOrCustomTextDialogCreator.showMarkdownHeadlineDialog(_activity, origText, (text, lineNr) -> {
+                        _hlEditor.setSelection(StringUtils.getIndexFromLineOffset(origText, lineNr, 0));
+                    });
+                    return true;
+                }
+                case R.string.tmaid_common_indent:
+                    runRegexReplaceAction(ZimWikiReplacePatternGenerator.indentOneTab());
+                    return true;
```
harshad1, line 179: @fredericjacob missing `renumberOrderedListIfRequired()`?

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiTextActions.java
harshad1: @fredericjacob ^^

fredericjacob:
@gsantner Oh, that comes fast. I fixed the zim filename creation (see my answer to your review comments above on NewFileDialog) as I think this is a critical issue for zim compatibility if the branch is going to be merged today.
When do you want to deploy and publish the master? Is there still time for additional PRs for fixing things?

@harshad1 I added the lines you mentioned to the text actions and tried it out. I could select a number for indentation, but saw no effect...? (But if this feature is missing for now and the PR has to be merged now, it's at least not the most critical thing IMO. Still I'll try to find out what is going wrong.)

harshad1:
> @harshad1 I added the lines you mentioned to the text actions and tried it out. I could select a number for indentation, but saw no effect...? (But if this feature is missing for now and the PR has to be merged now, it's at least not the most critical thing IMO. Still I'll try to find out what is going wrong.)

I see the reason, the mistake was mine. Zim uses tabs and not spaces for indentation. The logic I pointed at is only necessary for indenting with spaces and therefore does not apply to Zim and will have no effect.

gsantner:
@fredericjacob 
> Oh, that comes fast.

Well it's open since quite some time now with many people involved and it's somewhat getting stalled.  
I already did some comments that next update 2.5 will get out this year still, and for sure I want to include this too - like everything else.  

The problem is, it gets bigger and bigger & hard to review over time. I don't want this to go keep growing - till a point where I MUST say, NO I cannot merge this. It requires look over the mass of changes regularly and that for everybody involved.

Please do the fixes and polishing where needed, don't add new stuff in this PR more. We can add stuff again later where required.

Thanks everybody for working on this 😄 👍🏻

fredericjacob:
@harshad1 Okay. This tab/spaces thing is something that could need a closer look in the future. The current extra logic for zim indentation is not ideal, so it might make sense to add a global setting if tabs or spaces should be used (and set defaults for the particular formats).

> @fredericjacob missing `renumberOrderedListIfRequired()`?

This is also one method which relies on spaces where the tabs lead to bugs, so I left it out here. Needs some fix in the future. :D

gsantner:
@fredericjacob 
> Oh, that comes fast.

Well it's open since quite some time now with many people involved and it's somewhat getting stalled. I.e. I don't see much activity towards getting it done and ready for merge.   
I already did some comments that next update 2.5 will get out this year still, and for sure I want to include this too - like everything else.  

The problem is, it gets bigger and bigger & hard to review over time. I don't want this to keep growing - till a point where I MUST say, NO I cannot merge this. It requires look over the mass of changes regularly and that for everybody involved.

Please do the fixes and polishing where needed, don't add new stuff in this PR more. We can add stuff again later where required.

Thanks everybody for working on this 😄 👍🏻

gsantner:
notice: "tab" = X space characters atm at Markor. Not aware of something else for indentation. (Except: "Special Key" Tab button, but thats not relevant to this)

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiTextActions.java
fredericjacob: See comment below.

fredericjacob:
@gsantner Alright, I was only a bit afraid that this could suddenly go live, leaving behind some angry Zim-Markor-users wondering why the format is not working like expected. ^^ I'm definitely also looking forward for this giant (and more and more getting-overwhelming) PR being merged! :+1:
When I started developing on this, I had some weeks with plenty of free time available, but the situation changed, so yeah - progress has gotten slower now and the PR gets harder to maintain. :smile: 

From my side, I don't see any actual bugs or issues anymore in the latest state, so I'm okay with this being merged. There are improvements to be done of course, but these could be gradually made with additional PRs which should make reviewing easier.

fredericjacob:
> notice: "tab" = X space characters atm at Markor. Not aware of something else for indentation. (Except: "Special Key" Tab button, but thats not relevant to this)

Yes, that's why the zim text actions currently use a separate indentation logic (for tabs) - which is not ideal, but at least it works.

app/src/main/java/net/gsantner/markor/format/zimwiki/ZimWikiTextActions.java
fredericjacob: Does not work yet currently because of zim using tabs instead of spaces, but I added this to my TODO backlog. :+1:

app/src/main/java/net/gsantner/markor/format/AutoFormatter.java
fredericjacob: Yes exactly, it's a small helper wrapper class I introduced which is used exclusively to distinguish the prefixes between Markdown and Zim, to avoid long argument lists.
I like to keep this name for now because the wrapped patterns only refer to prefixes, but I'd change this if other format-specific patterns should be included in the future.

gsantner:
Ok as far I see @harshad1 feedback is also OK/resolved, and from your both sides nothing important anymore. right?

Then I keep the PR open till tomorrow and when nothing comes up I merge it tomorrow around this time.

gsantner (approved):

gsantner:
@fredericjacob @pschwede @harshad1 
Thanks for working so hard on this. It is now merged 👍🏻 🎈

fredericjacob:
Great, nice to see this! :)

gsantner:
Superscript and subscript look broken to me:

Screenshot_20201220-213456_Trebuchet

gsantner:
@fredericjacob 

Feedback for improvements:   

## 1 - Crossed list items
For crossed list items, we might use `[x] ~~striked text~~` for the zim-to-markdown conversion, and striketrough syntax highlighting at editor. What do you think?

fredericjacob:
> Superscript and subscript look broken to me:

Hm strange, I tried it with version 2.5 on my phone and with an emulator, for me it looks fine:
<div style="display: flex">


</div>

> For crossed list items, we might use `[x] ~~striked text~~` for the zim-to-markdown conversion. What do you think?

Yes, that's a good idea as "cross replacement" in view mode. :+1:

gsantner:
super/sub script issue seems not to happen always, not sure how to reproduce

PR #1100, opened 2020-10-20
fix crashing file search dialog caused by null object

fredericjacob:
Commit 05c92f074e7b6af2a8b02e344e6fdedb143ba4a1 introduced a bug which made Markor crash when pressing the file search button in the explorer. The reason is that `dopt.data` is not initialized and remains null, but it is queried at several places in `SearchOrCustomTextDialog`, leading to `NullPointerException`s or `TargetInvocationException`s.

This PR fixes the problem.

(@harshad1 I am not sure about the other `DialogOptions` attributes which are not initialized any more since the commit, maybe they might cause hidden bugs at other places, too, but maybe not.)

### DIFF
--- a/app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
+++ b/app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
@@ -65,7 +65,7 @@ public class SearchOrCustomTextDialog {
     public static class DialogOptions {
         public Callback.a1<String> callback;
         public Callback.a2<String, Integer> withPositionCallback;
-        public List<? extends CharSequence> data = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


harshad1:
It may be better to add defensive checks around uses of `dopt.data` and other parameters instead. I can work on this if you don't want to.

harshad1:
Could you also post how to reproduce this bug in practice?

fredericjacob:
@harshad1 Yes, you know your code better, so of course you can also work on a cleaner and more complete solution if you want to.

To reproduce it, just open the file explorer section in Markor. On top, there is the search icon for recursive file search. When tapping it, the app just crashes (tried it on my phone Oneplus 3 and in the emulator). With the previous commit before the mentioned one, the search dialog opens up and everything works fine.

PR #1101, opened 2020-10-21
Null checks around dopt.data

harshad1:
This addressed the issue raised in #1100 by @fredericjacob.

A null check has been added around dopt.data and another unchecked + unused line has been removed.


harshad1:
I have tested this against everything I could think of.

gsantner:
@fredericjacob
can you please try this and confirm it's ok where you found the issue? If so, as you said probably this fix is better than the simple fix.

fredericjacob:
@gsantner @harshad1 Tried it on my device, the file search dialog now works again (no crash anymore). :+1:

PR #1105, opened 2020-10-24
Handle trailing punctuations

harshad1:
Small improvement to handle opening and trailing punctuation in markdown bold and italics highlighting. There is still a lot of room for improvement, but this handles a couple of common use cases.


gsantner:
Can you spot noticable difference on NanoProfiler log?

harshad1:
@gsantner I could not see any difference.

Do you have any plans for moving to faster highlighting? A damage model or something?

gsantner:
When we want to move on to something other than Regex - the best would be to use something that others developed already, not brew our self. However this results in that we cannot add formats as we wish.. like Zim & todo.txt

PR #1109, opened 2020-10-26
Long press sort to sort by apply most recent sort

harshad1:
This PR adds the ability to apply the most recently applied sort in todo.txt files by long pressing the sort action.

Really useful when updating priorities / due dates and trying to see how things stack up.


app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
```diff
@@ -35,15 +35,12 @@
 import net.gsantner.markor.util.AppSettings;
 import net.gsantner.opoc.util.StringUtils;
 
-import java.text.SimpleDateFormat;
```
harshad1, line 38: This is just cleanup of unused imports

PR #1112, opened 2020-11-02
Editor Search & replace, closes #155

### LINKED ISSUE #155: Search (+replace?)
#### General information

* **App version:** 0.3.0
* **System:** 7.0.1

#### Description

Add search feature to document editor. This will allow to enter text, jump between found results. Maybe display matches and current jumped index too. (e.g. 5/10).

If time available, too go for simple search and replace. Add an "Replace all" action too in this case.

harshad1:
This PR adds search and replace functionality.

The search / replace dialog is accessed by pressing the search dialog and then clicking Search / Replace.

Todo:
- [x] Saving of recent search - replace pairs
- [ ] Testing

Future / would be nice / maybe:
~~- [ ] Add come common useful actions to the dropdown~~
~~- [ ] Add a replace preview window~~

image


app/src/main/java/net/gsantner/markor/ui/SearchReplaceDialog.java
```diff
@@ -0,0 +1,316 @@
...
+
+        // Popup window for ComboBox
+        popupWindow.setAdapter(new ArrayAdapter<ReplaceGroup>(activity, android.R.layout.simple_list_item_1, recentReplaces) {
+            @NonNull
+            @Override
+            public View getView(int pos, @Nullable View view, @NonNull ViewGroup parent) {
+                final TextView textView = (TextView) super.getView(pos, view, parent);
+
+                final ReplaceGroup rg = getItem(pos);
+                textView.setText(res.getString(R.string.search_replace_recent_format, rg._search, rg._replace, rg._isRegex, rg._isMultiline));
```
harshad1, line 110: This is ugliest part of this dialog. Each line of the recents pattern is simply:

```
Search for: <search pattern>
Replace with: <replace pattern>
Use regex: <true/false>, Multiline regex <true/false>

```
(note the empty line). It works ok, but is pretty ugly. Any suggestions are most welcome

app/src/main/res/values/strings.xml
```diff
@@ -398,4 +398,14 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
...
+    <string name="search_and_replace">Search / Replace</string>
+    <string name="use_regex">Use regex</string>
+    <string name="multiline_regex">Multiline regex</string>
+    <string name="text_to_search">Search for</string>
+    <string name="text_to_replace">Replace with</string>
+    <string name="replace_first">Replace First</string>
+    <string name="replace_all">Replace All</string>
+    <string name="search_replace_count_message">Found %d matches</string>
+    <string name="search_replace_pattern_error_message">Search or replace pattern error.</string>
+    <string name="search_replace_recent_format">Search pattern: %s\nReplace plattern: %s\nUse regex: %b, Multiline regex: %b\n</string>
```
harshad1, line 410: See comment above about uglyness

harshad1:
@gsantner This PR is ready for review. I am pretty bad at UX design :), so please let me know what changes might make this better.

There is a lot more which can be done, but this is a decent minimum viable product IMO

app/src/main/res/values/strings.xml
gsantner: I would do:
```
     <string name="search_replace_count_message">Found matches: </string>   ..... + number
string("Search pattern:") + number + "\n"
string("Replace pattern:) + number....
```

makes it easier to translate. happens often that translators don't add `%` properly.

gsantner:
Can you share a screenshot how it looks for rought view?

gsantner:
Can you share a screenshot how it looks roughly?

harshad1:
image

gsantner:
Oh so like 5 items, plus a dropdown with predefined templates?

harshad1:
Dropdown only has history. I don't have any predefined templates.

Current history count = 10.

Nov. 3, 2020 10:19:11 Gregor Santner <notifications@github.com>:

harshad1:
This is ready to be merged imo

app/src/main/java/net/gsantner/opoc/util/StringUtils.java
```diff
@@ -195,4 +195,50 @@ public static boolean isNewLine(CharSequence source, int start, int end) {
...
+     * A\tB -> A    B
+     *
+     * A\nB -> A
+     *         B
+     *
+     * @param input Input string
+     * @return String with escaped sequences converted
+     * @throws IllegalArgumentException if an invalid escape sequence is found in the string
+     */
+    public static String unescapeString(final String input) throws IllegalArgumentException {
```
harshad1, line 213: This code allows a user to use special characters such as newline `\n` and tab `\t` in the replace text box

harshad1:
This is now ready. 

The only thing missing is documentation to go in the wiki

Adding some notes here:
1. This will search / replace in selected text. If no text is selected (start == end), then the whole document will be searched / replaced
2. No regex means the search pattern is compiled with the java `Pattern.LITERAL` modifier
3. Multiline regex means that the search pattern is compiled with the `Pattern.MULTILINE` modifier
4. Some sequences in the 'replace' string will be unescaped. i.e.
    1. `\t` will insert a tab
    2. `\b` will insert a backspace
    3. `\r` will insert a carriage return
    4. `\n` will insert a new line
    5. `\f` will insert a form feed

The unescape behavior is roughly similar in behavior to (though simplified) version of the Apache [unescapeJava](http://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringEscapeUtils.html#unescapeJava(java.lang.String)) routine.

gsantner:
> The only thing missing is documentation to go in the wiki

If you want to add something, please prepare it here on this PR as a (markdown) comment.  
(I'm moving that to markor's jekyll website then, GitHub Wiki&Readme is not much on focus anymore)

harshad1:
@gsantner Order bug fix has been pushed to this PR

gsantner:
@harshad1 
Feedback:

* Would it be hard to make the "replace first" do replace first match on or after current cursor position? Do you think that makes sense?
* You shown a fancy list in screenshot. What do you think of providing that as default history at first start? So when user clicks the dropdown button it actually does something. (Here at first click it just did nothing). Might be helpful for some people.
* Changing `Found matches` to just `Matches: X` - in german likeley first goes into two lines.
* Same for Use Regex -> just regex

pushing the renaming to your branch

harshad1:
>     * Would it be hard to make the "replace first" do replace first match on or after current cursor position? And if nothing matches till end, do from index 0. Do you think that makes sense? Thats at least how it works usually in desktop editors / notepads

This makes sense. I will look into implementing it.

>     * You shown a fancy list in screenshot. What do you think of providing that as default history at first start? So when user clicks the dropdown button it actually does something. (Here at first click it just did nothing). Might be helpful for some people. Alternative: setVisibilty(invisible)

I've added 4 default useful patterns. I am looking into adding labels or something to make them easier to understand.

gsantner:
I would say no labels required, no need

harshad1:
Will be testing this for another day because of the changes.

harshad1:
@gsantner This is good to go

gsantner:
I see some textview parameters..would Spannable also work in that case?

harshad1:
> I see some textview parameters..would Spannable also work in that case?

Would need an editable as we need to get selection and set text

gsantner:
```
     private SearchReplaceDialog(final Activity activity, final TextView text) {
```
What about replacing textview with `Editable editable, int cursor`?

harshad1:
@gsantner switched

harshad1:
@gsantner are there any changes that you would like here?

app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
```diff
@@ -80,6 +80,9 @@
         public Callback.a1<Spannable> highlighter;
         public String extraFilter = null;
 
+        public String neutralButtonText = null;
```
gsantner, line 83: can be int then as well like the others, pass **R.string.search_and_replace**.

I will push that to branch self

gsantner:
thx merged

Issue #1004, opened 2020-07-20
Setting to control document open cursor pos (top/bottom), seperate last pos option


DS-109:
#### General information

* **App version:** 2.2.10
* **System:** Android

#### Description
The default is to enter a document at the last cursor position, which was implemented in 2018:
https://github.com/gsantner/markor/issues/205
As an alternative, the settings have an option "start at bottom" (?, in German "Unten beginnen").
I would be happy if this option could be extended to a third alternative "start at top"!

When I open documents after a longer time, I usually read it from the beginning before I decide where to write something ( a chapter, a thing in a sorted list, etc.)

FrantisekRadacovsky:
Hey. I would like to start working on it, is it still up for grab? If yes could you please assign it to me? Thanks.

PR #1113, opened 2020-11-03
Setting to control document open cursor pos (top/bottom), seperate la…

FrantisekRadacovsky:
- Fixed bug, when setting cursor on the bottom doesn't work.
- Issue "Setting to control document open cursor pos (top/bottom), seperate last pos option #1004" -> done.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
@@ -24,6 +24,7 @@ import android.support.annotation.NonNull;
 import android.support.v4.content.ContextCompat;
 import android.support.v7.app.AppCompatActivity;
 import android.support.v7.widget.Toolbar;
+import android.util.Log;
 import android.util.TypedValue;
 import android.view.Gravity;
 import android.view.Menu;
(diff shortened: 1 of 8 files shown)


gsantner:
Hello thanks

This is still work in progress right?

I found some debug code and commented code.

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -656,17 +657,28 @@ public void onFragmentFirstTimeVisible() {
         AppSettings as = new AppSettings(getContext());
         if (_savedInstanceState == null || !_savedInstanceState.containsKey(SAVESTATE_CURSOR_POS) && _hlEditor.length() > 0) {
             int lastPos;
-            if (_document != null && _document.getFile() != null && (lastPos = as.getLastEditPositionChar(_document.getFile())) >= 0 && lastPos <= _hlEditor.length()) {
-                if (!as.isPreviewFirst()) {
-                    _hlEditor.requestFocus();
+            if (as.getCursorPosition() != null) {
+                if (as.getCursorPosition().equals("End of document")) {
```
harshad1, line 661: This will break when translations are applied. You should do something like

```java
final Resources res = activity.getResources();
if (as.getCursorPosition().equals(res.getString(R.string.cursor_end)) {
```

harshad1: It would be even better, imo, to separate the string description from the key

gsantner: Yes that one was meant especially with "work in progress" :D

harshad1: Ah sorry, I am so used to looking for `WIP:` that I did not see the draft tag :)

gsantner:
@FrantisekRadacovsky ?

FrantisekRadacovsky:
Hi. I'll look into it as soon as possible. Sorry for the inconvenience.

gsantner:
thanks

no problem, just let me know when you have something to look at.

gsantner:
closing for now, can always be reopened again when it is ready.

PR #1129, opened 2020-11-19
Per file settings

harshad1:
This PR adds the following toggles in the document menu

1. Wrap state
2. Enable / Disable highlighting

Additionally, this PR adds the ability to set indent per file by long pressing the indent actions, and the ability to remember the last mode (view vs edit)

Also included is some code to reduce the errors caused by AccessibilityEvents in longer files. Accessibility is disabled in files > 10k characters. AFAIK this is an android limitation. Above 10k characters, accessibility can cause crashes and data loss.


harshad1:
@gsantner As discussed, please let me know which (if any) other options should have toggles in the document menu

app/src/main/res/menu/document__edit__menu.xml
```diff
@@ -79,22 +80,39 @@
...
         android:title="@string/search"
         android:icon="@drawable/ic_search_black_24dp"
         app:showAsAction="always"
         app:actionViewClass="android.widget.SearchView" />
+
     <item
         android:id="@+id/action_save"
         android:icon="@drawable/ic_save_black_24dp"
         android:title="@string/save"
         app:showAsAction="never" />
```
gsantner, line 101: as first step, please wrap the new 2 into a common submenu

gsantner:
I'm not totally sure if it's the right path with checkbox menu options (as we are limited to exactly that with mrnu then)... we will see

app/src/main/java/net/gsantner/markor/ui/hleditor/HighlightingEditor.java
```diff
@@ -150,22 +145,44 @@ private void cancelUpdate() {
...
+        highlightWithoutChange();
+    }
+
+    // This allows us to disable sending of accessibility events when highlighting is being
+    // performed. Sending of many accessibility events can cause performance issues and crashes.
+    // Blocking accessibility during _highlighting_ should have no impact on users.
+    @Override
+    public void sendAccessibilityEventUnchecked(AccessibilityEvent event) {
+        if (_accessibilityEnabled && getText().length() < 10000) {
+            super.sendAccessibilityEventUnchecked(event);
```
harshad1, line 157: @gsantner Added the fix for the accessibility / crash issue to this PR as there were some conflicting changes. I will close the other PR. This fix is a little better, but we still need to determine the correct max length and what the impact is on accessibility users.

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -120,6 +128,8 @@ public static DocumentEditFragment newInstance(File path, boolean pathIsFolder,
     private boolean _firstFileLoad = true;
 
     public DocumentEditFragment() {
+        super();
+        _appSettings = AppSettings.get();
```
gsantner, line 132: preferably, please use one of the init / onCreate method overrides to init with fragment context .. avoid AppSettings.get() if possible

app/src/main/java/net/gsantner/markor/ui/hleditor/HighlightingEditor.java
gsantner: as far I can see accessbility is only disabled during highlighting runs, right? if so, should be fine

harshad1: Am still trying to figure out the minimal impact fix here. New version does 2 thing:
1. Block accessibility during highlighting
2. If the text in the accessibility message is too large, then cut it down

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
harshad1: Fixed

app/src/main/java/net/gsantner/markor/ui/hleditor/HighlightingEditor.java
harshad1: This seems to be working so far. If we can do this properly, then there should be no impact on accessibility users while solving the crash issue

harshad1:
> @gsantner As discussed, please let me know which (if any) other options should have toggles in the document menu

harshad1:
@gsantner Added the long press indent button to change setting here (seemed logical)

Accessibility is currently disabled above 10k characters. Should we add an experimental setting to allow this to be overwritten 

-[ ] Enable accessibility for large files
Warning: may cause crashes and data loss

harshad1:
This is ready to go.

If you have suggestions of what else we want to add to File options menu I can add them here.

Or we can always make a follow up or.

harshad1:
@gsantner I have had absolutely no crashes or performance issues since implementing the 10000 char limit for accessibility here.

harshad1:
Do you think adding a per file 'open in view mode' makes sense?

gsantner:
Storing last state for current file, and apply again when open next time is a better option a would say

harshad1:
> Storing last state for current file, and apply again when open next time is a better option

Implemented and pushed. I have removed the global setting as it doesn't make sense any more.

app/src/main/java/net/gsantner/opoc/util/StringUtils.java
```diff
@@ -245,4 +248,14 @@ public static String unescapeString(final String input) {
 
         return builder.toString();
     }
+
+    public static String getPath(final Document document) {
```
gsantner, line 252: please don't add markor.Document to StringUtils if not required, which is not here.  
For example, have it in DocumentIO or at the other getPath() place

gsantner:
Those people that ask specifically for "how to always open directly in view mode" are those who usually check the "Always view mode" settings option. Thus just always return true when it is checked in settings. For false case we return the (NEW) last state.

pushed that small change

harshad1:
> pushed that small change

Nice approach,

I have moved getPath to Document (still as a static function)

gsantner:
`it's ok to be static` :D.

I spent like 1h-2h now to go through all the PRs and issues..did quite some things. Let me know if you have some feedback somewhere.

What do you think, still something todo in this PR, something you want to test?

harshad1:
> I spent like 1h-2h now to go through all the PRs and issues..did quite some things. Let me know if you have some feedback somewhere.

I see all the emails :D.

> What do you think, still something todo in this PR, something you want to test?

Want to spend a day testing the more recent changes. I don't want to add any more to this PR.

harshad1:
@gsantner this is ready to go

gsantner:
thanks much, merged! 

If you have time, can you please help review/test test the Zim PR? The most important point is that it not break Markdown and other formats (in regards of the ListHandler, AutoFormat, Highlighter, TextActions etc).  
Due you refactored a lot there recently you know that best atm, so if you can help to check that would be cool.  
However, don't look at Zim format, focus is on "not break something else except Zim". Zim is the first/initial version and it's ok when it's not 100% done right away.

harshad1:
> If you have time, can you please help review/test test the Zim PR? The most important point is that it not break Markdown and other formats (in regards of the ListHandler, AutoFormat, Highlighter, TextActions etc).

Will do.

Issue #1128, opened 2020-11-18
Repeated crashes when editing a large markdown file


harshad1:
I am seeing repeated crashes when editing a relatively large markdown file (~1600 lines, ~12000 words, ~60000 characters.).

I see the following trace. But I am unsure what / where the underlying cause is

```
E/View: hasTransientState decremented below 0: unmatched pair of setHasTransientState calls
E/JavaBinder: !!! FAILED BINDER TRANSACTION !!!  (parcel size = 93208)
E/AccessibilityManager: Error during sending EventType: TYPE_VIEW_TEXT_SELECTION_CHANGED; EventTime: 434232397; PackageName: net.gsantner.markor_test; MovementGranularity: 0; Action: 0; ContentChangeTypes: []; WindowChangeTypes: [] [ ClassName: android.widget.EditText; Text: 
< ENTIRE CONTENTS OF THE DOCUMENT>
ContentDescription: null; ItemCount: 61944; CurrentItemIndex: -1; Enabled: true; Password: false; Checked: false; FullScreen: false; Scrollable: false; BeforeText: null; FromIndex: 61835; ToIndex: 61835; ScrollX: 0; ScrollY: 0; MaxScrollX: 0; MaxScrollY: 0; ScrollDeltaX: -1; ScrollDeltaY: -1; AddedCount: -1; RemovedCount: -1; ParcelableData: null ]; recordCount: 0 
    android.os.DeadObjectException: Transaction failed on small parcel; remote process probably died
        at android.os.BinderProxy.transactNative(Native Method)
        at android.os.BinderProxy.transact(BinderProxy.java:540)
        at android.view.accessibility.IAccessibilityManager$Stub$Proxy.sendAccessibilityEvent(IAccessibilityManager.java:580)
        at android.view.accessibility.AccessibilityManager.sendAccessibilityEvent(AccessibilityManager.java:606)
        at android.view.ViewRootImpl.requestSendAccessibilityEvent(ViewRootImpl.java:8618)
        at android.view.ViewGroup.requestSendAccessibilityEvent(ViewGroup.java:1094)
        at android.view.ViewGroup.requestSendAccessibilityEvent(ViewGroup.java:1094)
        at android.view.ViewGroup.requestSendAccessibilityEvent(ViewGroup.java:1094)
        at android.view.ViewGroup.requestSendAccessibilityEvent(ViewGroup.java:1094)
[...]

PR #1130, opened 2020-11-19
Prevent onSelectionChanged from triggering a bug with larger files

harshad1:
The fundamental issue in https://github.com/gsantner/markor/issues/1128 appears to be that the `onSelectionChange` code in the `TextView` base class calls some accessibility service. The _whole text of the document_ is passed to this service through an intent. The whole document can exceed the intent size limit, causing a crash which sometimes propagates back to markor.

Blocking the call prevents this

### DIFF
--- a/app/src/main/java/net/gsantner/markor/ui/hleditor/HighlightingEditor.java
+++ b/app/src/main/java/net/gsantner/markor/ui/hleditor/HighlightingEditor.java
@@ -276,7 +276,12 @@ public class HighlightingEditor extends AppCompatEditText {
 
     @Override
     protected void onSelectionChanged(int selStart, int selEnd) {
-        super.onSelectionChanged(selStart, selEnd);
+        // super.onSelectionChanged triggers some accessibility service code.
+        // This appears to cause an exception with larger files
+        // Possible android bug? Commenting out.
+
+        // super.onSelectionChanged(selStart, selEnd);
+
         if (MainActivity.IS_DEBUG_ENABLED) {
             AppSettings.appendDebugLog("Selection changed: " + selStart + "->" + selEnd);
         }


gsantner:
I know there are people using Markor that depend on accessibiliy features and activeley using them, i.e. blind+talkback. Might be that selected text "saying loud" doesnt work then anymore.

Will check bugtracker if can find one of users, maybe we can arrange a testbuild for them

harshad1:
Hmmm. Good point. One other thing we can do is add a 25k character test around this. i.e. allow the upstream call if the file is < 25k characters

gsantner:
potentially we could also limit the super call to only call it when selection range is i.e. over 3000 characters..so like 

```
selStart, ((selEnd-selStart)>3000 ? (selStart+3000) : selEnd);
```

gsantner:
was typing/thinkibg similar, typing slow :D

gsantner:
I have something witk ca 9000 limit in mind (totally 10000 but activity requires some bytes on its owb). With string data requiring around 2 byte per char (i.e. utf16) we should be save with 3000

gsantner:
btw, popup "copy" and "translate" button may also be affected

harshad1:
> btw, popup "copy" and "translate" button may also be affected

These do not seem to be affected

harshad1:
> What do you think about that path?

The bug appears to be triggered even when selStart == selEnd (i.e. for example, when just entering text). Not sure we will be able to filter this by the selection size, but only by the whole document size.

Going to sleep on this. Hopefully we can find a solution which solves the issue with minimal impact on accessibility users.

gsantner:
ah ok good to know then we probably really only can if (length  < x)

harshad1:
Closing as an improved fix is included in https://github.com/gsantner/markor/pull/1129

Issue #1133, opened 2020-11-21
Add new preset color scheme


lokesh-krishna:
#### General information
Could you please consider adding the [Nord colour scheme](https://www.nordtheme.com/) as one of the preset color schemes?

#### Description
If you could tell me what colors need to be specified then I can list all the corresponding colour hex codes from Nord's [official documentation](https://www.nordtheme.com/docs/colors-and-palettes).

gsantner:
Please paste me the 2 to 4 colors for it (Standard Foreground text color, Standard background text color - for dark and light mode respectiveley).

https://github.com/gsantner/markor/blob/master/app/src/main/res/values/colors.xml#L52

lokesh-krishna:
Made a PR: #1134.

PR #1134, opened 2020-11-21
Add Nord colors

lokesh-krishna:
Adds the Nord colors as discussed in #1133.


gsantner:
thanks

can you please also modify the code for it? See AppSettings.java and preferences_master.xml, should be copy-pasteable+rename from one of the exsisting ones.

lokesh-krishna:
I edited `preferences_master.xml` but I'm not sure about what needs to be added in `AppSettings.java`. Could you point me to a reference line number?

gsantner:
SettingsActivity

lokesh-krishna:
Is this alright?

gsantner:
thanks yes looks OK

PR #1139, opened 2020-11-30
Fix for correct file not opened from shortcut

harshad1:
When multi document is turned off and markor is already running, sometimes shortcuts to a file will fail to open the correct file (instead the previously open file will be shown).

This PR improves the launch behavior. Now the correct file is opened correctly.


gsantner:
have you tried after this

* open file in markor
* from browser ie share into markor, select new/existing document

now 2 windows open?

harshad1:
With multiwindow enabled or disabled, the share into behavior is the same as before:

1. Open document in Markor (if multiwindow enabled, several documents can bed open)
1. Open browser and share into Markor
2. Share into always opens in new window.
3. Share into file browser is in the share into window (i.e. separate)
4. Share into works correctly and closes when file is selected.
5. Existing open document(s) are not affected.

H


Nov. 30, 2020 01:40:07 Gregor Santner <notifications@github.com>:

> have you tried after this
> 
* > open file in markor
* > from browser ie share into markor, select new/existing document

PR #1156, opened 2020-12-13
Add ZettelKasten template, close #1143

### LINKED ISSUE #1143: Add Zettelkasten template/sample
#### General information

* **App version:** 2.4.0 Play Store
* **System:** Android 10

#### Description
I've noticed that Markor already has a lot of built-in templates support. But for me, I usually create a new file with the name start with `YYYYMMDDHHmm`, such as 20201205133000，and then append the file name.

But for now, I have to input the prefix on my own. If there is an option to custom the default file name, it will be much easy to use.

I have checked the code if it is acceptable, I'd like to open a PR.

einverne:
close #1143

### DIFF
--- a/samples/zettelkasten.md
+++ b/samples/zettelkasten.md
@@ -1,3 +1,4 @@
 source:
 category:
 tag:
+------------
--- a/samples/zettelkasten.md
+++ b/samples/zettelkasten.md
@@ -1,0 +1,3 @@
+source:
+category:
+tag:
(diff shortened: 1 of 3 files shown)


samples/zettelkasten.md
```diff
@@ -0,0 +1,3 @@
+source:
+category:
+tag:
```
gsantner, line 3: where does normal text content start here? If required/possible add a 

```
------------
```
or empty sep. line here, so it's clear for everybody.

gsantner (changes requested):

gsantner (approved):

PR #1168, opened 2020-12-22
Per file enable/disable syntax highlighting, default value based on initial filesize

harshad1:
This PR addresses the issue discussed here: https://github.com/gsantner/markor/issues/95#issuecomment-749168886


harshad1:
@gsantner here is your 'dummy' PR :). It also solves the issue with highlighting and file size discussed earlier today

harshad1:
@gsantner I believe the unit tests aren't being run by the CI system.

gsantner:
they are still, just no additional JUnit tab. Buildlog contains everything in the CI run for this PR. Also theres logfiles of the PASSED tests

harshad1:
image

For reference, this is what I see.

app/src/main/java/net/gsantner/markor/ui/hleditor/HighlightingEditor.java
```diff
@@ -156,10 +155,17 @@ public void sendAccessibilityEventUnchecked(AccessibilityEvent event) {
...
 
+    public void setBeforeHighlightCallback(final Callback.a0 callback) {
+        _beforeHighlightCallback = callback;
+    }
+
     private void highlightWithoutChange() {
-        final Editable editable = getText();
-        final boolean isFileShortEnough = editable.length() <= (_isDeviceGoodHardware ? 100000 : 35000);
-        if (isFileShortEnough && _hlEnabled) {
+        if (_beforeHighlightCallback != null) {
```
harshad1, line 163: @gsantner What do you think of this approach? This checks the file size on every highlight and disables the highlighting when size > threshold

Another, simpler way is to just check on file load.

gsantner: I think overall it's more complicated than it should be.  
Already thought about it and have something in mind, but not sure when I can make it.

What I would go for:
* Highlighter class(es) 
  * No limit in the  enable/disable supplied from outside
* DocumentActivity
  * `AppSettings::highlightingEnabled(File file, int filesize/characters/whatever)`
  * return first user preference set
  * then check limit and set default false+return in case too big
  * otherwise return global value
 
This not covers the case where a file would grow XXL size, but assuming you already had a file once open it doesn't change that much - and even if so, you can always disable it still then. The MOST important case is "first time open of a huge file".

gsantner:
Did you actually implement what? I meant I would do that, but also fine if you do :D.

Then the goodHardware check can be removed from ` net/gsantner/markor/ui/hleditor/HighlightingEditor.java`

and the overhauled methods/const can be removed too at appsettings `PREF_PREFIX_HIGHLIGHT_SIZE_CHECKED`

gsantner:
not sure, I tried everything to get Junit job away, it just doesn't go away. github bug or something

harshad1:
@gsantner Could you check, I think this PR is ready

gsantner (approved):
thx, did some more removal of unneeded refs

PR #1177, opened 2020-12-29
Allow to use encryption at Android>=M/API23

opensource21:

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Markor is a TextEditor for Android. This project aims to make an editor that is 
 <br/>🌎 Language selection -- use other language than on the system
 <br/>
 <br/>🔃 Markor is an offline app. It works with sync apps, but they have to do syncing respectively. Sync clients known to work in combination include BitTorrent Sync, Dropbox, FolderSync, OwnCloud, NextCloud, Seafile, Syncthing, Syncopoli
-<br/>🔒 Can encrypt your textfiles with AES256. You need to set a password at the settings and use Android device with version KitKat or newer. You can use [jpencconverter](https://gitlab.com/opensource21/jpencconverter) to encrypt/decrypt easily on desktop. Be aware that only the text is encrypted not pictures or attachments.
+<br/>🔒 Can encrypt your textfiles with AES256. You need to set a password at the settings and use Android device with version Marshmallow or newer. You can use [jpencconverter](https://gitlab.com/opensource21/jpencconverter) to encrypt/decrypt easily on desktop. Be aware that only the text is encrypted not pictures or attachments.
 **Hint** This protects you content from other apps or person who are able to read your sd-card. It's **no** protection against anyone who was able to open the markor-app.  
 
 ### Privacy<a name="privacy"></a>
--- a/README.md
+++ b/README.md
@@ -44,7 +44,8 @@ Markor is a TextEditor for Android. This project aims to make an editor that is 
 <br/>🌎 Language selection -- use other language than on the system
 <br/>
 <br/>🔃 Markor is an offline app. It works with sync apps, but they have to do syncing respectively. Sync clients known to work in combination include BitTorrent Sync, Dropbox, FolderSync, OwnCloud, NextCloud, Seafile, Syncthing, Syncopoli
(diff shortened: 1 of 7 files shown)


gsantner:
Hey thanks!

May I ask for a favor? Could you add two "test" files and junit cases with different passwords to the tests directory? Just for the future when we make changes here, that all still works as expected

opensource21:
other.de.stanetz.jpencconverter.JavaPasswordbasedCryption is tested at https://gitlab.com/opensource21/jpencconverter/-/blob/master/src/test/java/de/stanetz/jpencconverter/cryption/JavaPasswordbasedCryptionTest.java.

I could copy it but then I need JUnit5 in the project.

gsantner:
As far I see it's more work to migrate everything to Junit5 than just rewrite the test case. It's anyway assertEquals only after all right?

I can do that also if it makes it oto much trouble

opensource21:
I used there  
```java
    @ParameterizedTest
    @EnumSource(JavaPasswordbasedCryption.Version.class)
```
which ist not so easy to migrate. JUnit4 and 5 in parallel would be a solution, but I think most pragmatic would the rule not to test third-party tests. If you add the tests know they must be adjusted with each new version. I would think it's more worth to write tests for untested code, than to write test for copied code which is tested outside.

gsantner:
well you might change code in your repo (and free todo so). It just might not be always the case that its all equal. And also, unit tests are always executed at markor, so we at least always know the encryption stuff works as expectrd

opensource21:
This problem you always have with third-party products. But I would be really surprised if the test will ever find a real bug. However, if you want to increase the test-coverage I would recommend switching to JUnit5 now. Copy the test than is easy. But if it's important for you feel free to add some tests. For me it's well tested code, I fixed a bug which is tricky to test with Unit-Tests very quickly without any advantage for me. So I want to concentrate the next day on my family and not on double testing code in a project where the code-coverage is low.

gsantner:
I just copied your file and put a loop around, thats it. Commit and pushed to your PR branch.

README.md
```diff
@@ -44,7 +44,8 @@ Markor is a TextEditor for Android. This project aims to make an editor that is
 <br/>🌎 Language selection -- use other language than on the system
 <br/>
 <br/>🔃 Markor is an offline app. It works with sync apps, but they have to do syncing respectively. Sync clients known to work in combination include BitTorrent Sync, Dropbox, FolderSync, OwnCloud, NextCloud, Seafile, Syncthing, Syncopoli
-<br/>🔒 Can encrypt your textfiles with AES256. You need to set a password at the settings and use Android device with version KitKat or newer. You can use [jpencconverter](https://gitlab.com/opensource21/jpencconverter) to encrypt/decrypt easily on desktop. Be aware that only the text is encrypted not pictures or attachments.  
+<br/>🔒 Can encrypt your textfiles with AES256. You need to set a password at the settings and use Android device with version KitKat or newer. You can use [jpencconverter](https://gitlab.com/opensource21/jpencconverter) to encrypt/decrypt easily on desktop. Be aware that only the text is encrypted not pictures or attachments.
+**Hint** This protects you content from other apps or person who are able to read your sd-card. It's **no** protection against anyone who was able to open the markor-app.
```
gsantner, line 48: I don't think this Hint notice is necessary. If your device is under somebody's (unwanted) control, everything is under their control - nothing specific to Markor.

opensource21: I don't think that there only exists total control or no control. Every file-manager is normally able to read the sd-card and can send everything to a server. Against this it's an effective protection. There are other app like keepass, which protect the data even if someone open the app or is able to make screenshots. So I think it's a good idea to make this clear in the readme. Obviously for everyone how use the app should be clear, that if he gives someone the phone and the person open the markor app, that the other person can read the private notes. But therefore you must install the app.

gsantner: I understand, but password for for opening the app ("app lock") and encrypting files with a password are two complete different things

gsantner:
Thanks is there something left from your side or ready to merge?

opensource21:
Is ready to merge

gsantner:
tried on phone with file created from master branch, worked

gsantner (approved):
tried on phone with file created from master branch, worked

PR #1187, opened 2021-01-06
WIP: Reduce flicker x2

harshad1:
This moves the cursor start code to onResume. Having this code run _after_ the view is visible reduces flicker and tearing


app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -719,26 +744,6 @@ private void checkReloadDisk(boolean forceReload) {
...
-                _hlEditor.scrollTo(0, _appSettings.getLastEditPositionScroll(_document.getFile()));
-            } else if (_appSettings.isEditorStartOnBotttom()) {
-                if (!initPreview) {
-                    _hlEditor.requestFocus();
-                }
-                _hlEditor.setSelection(_hlEditor.length());
-            }
-        }
-    }
-
```
harshad1, line 741: This code runs 1ms after `onViewCreated`. Based on what is happening, it can conflict with wrapping code and scroll handler addition etc. Moved to `onResume` which runs later

app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -339,78 +339,98 @@ public void toggleFavouriteFile(File file) {
...
+        // Includes hack to prevent last edit from being set for quicknote and todo
+        if (pathOk(path) && !getTodoFile().equals(path) && !getQuickNoteFile().equals(path)) {
+            setInt(PREF_PREFIX_EDIT_POS_CHAR + path, pos, _prefCache);
+            setInt(PREF_PREFIX_EDIT_POS_SCROLL + path, scrolloffset, _prefCache);
         }
     }
 
     public void setDocumentWrapState(final String path, final boolean state) {
-        setBool(PREF_PREFIX_WRAP_STATE + path, state);
+        if (pathOk(path)) {
```
harshad1, line 355: Defensive code added to all setters and getters

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -241,8 +243,32 @@ public void onResume() {
...
 
-        // Set initial wrap state
         initDocState();
+
+        // Run once in first onResume if visible. Should always be visible here
+        if (_firstStart && getUserVisibleHint()) {
+            _firstStart = false;
+            // 1. Start on bottom if that is the setting or if quicknote file
+            // 2. Do not set focus or set cursor for todo file
+            // 3. If other file and save cursor position exists, start there
```
harshad1, line 253: Cleaned up the logic here. @gsantner Please review and let me know if this behavior is good. It is the same as what exists currently, except for quicknote, which always starts at the bottom now.

app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -339,78 +341,98 @@ public void toggleFavouriteFile(File file) {
...
-        if (!file.equals(getTodoFile()) && !file.equals(getQuickNoteFile())) {
-            setInt(PREF_PREFIX_EDIT_POS_CHAR + file.getAbsolutePath(), pos, _prefCache);
-            setInt(PREF_PREFIX_EDIT_POS_SCROLL + file.getAbsolutePath(), scrolloffset, _prefCache);
+    private boolean pathOk(final String path) {
+        return (path != null && !path.trim().isEmpty());
+    }
+
+    public void setLastEditPosition(final String path, int pos, int scrolloffset) {
+        // Includes hack to prevent last edit from being set for quicknote and todo
+        if (pathOk(path)) {
```
harshad1, line 350: Removed the special case code here and using paths, not `File` objects. Special case logic now handled in the `DocumentEditFragment` directly.

app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -339,78 +341,98 @@ public void toggleFavouriteFile(File file) {
...
-            return -1;
-        }
-        if (file.equals(getTodoFile()) || file.equals(getQuickNoteFile())) {
-            return -2;
+    public int getLastEditPositionChar(final String path) {
+        final int _default = -1;
+        if (pathOk(path)) {
+            return getInt(PREF_PREFIX_EDIT_POS_CHAR + path, _default, _prefCache);
+        } else {
+            return _default;
```
harshad1, line 425: Similarly removed special case code here. Not needed any more

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -241,8 +243,32 @@ public void onResume() {
             _hlEditor.setText(CoolExperimentalStuff.convertEpubToText(_document.getFile(), getString(R.string.page)));
         }
 
-        // Set initial wrap state
         initDocState();
+
+        // Run once in first onResume if visible. Should always be visible here
```
harshad1, line 248: onResume is a better place for this logic as the UI is fully setup here. Flicker was being caused by conflicts between view generation and setting selection etc

harshad1:
Upon further testing, this whole approach does not seem to work consistently :(. Closing for now

PR #1195, opened 2021-01-16
Additional Textactions for Zim Format, closes #1186

### LINKED ISSUE #1186: zim: edit-mode textaction - link insert
#### General information

* **App version: 2.5.0** 

#### Description
Links inserted by Markor tool-menue are html-style, not zim-style. This is ok within Markor, but does not render well in zim

Link given by Markor tool-menue:
``
<a href='Synctest.txt'>Link zu Synctest.txt</a>
``

Link given ZIM style:
``
[[./Synctest.txt|Link zu Synctest.txt]] 
``

fredericjacob:
This PR adds the following, yet missing/incomplete textactions for the Zim Wiki format in the editor:

**Short-Press-Actions:**

- [x] insert image
   * selected image is copied to the folder corresponding to the current zim page (e.g. folder "My_Page" for "My_Page.txt")
      * taken/edited pictures are directly saved in this corresponding folder
   * relative link is created for the copied image so that the image is shown in view mode (and in the Zim desktop application), e.g. "{{./MyImage.jpg}}"
- [x] insert link
   * URLs are kept intact
   * if a file is selected, it is copied in the same manner as for images
   * relative link is created, e.g. "[[./somefile.txt|My cool file]]"
   * zim inter-wiki-links (to other pages in the notebooks) are not yet supported
- [x] include common actions (format-independent) for moving text one line up/down, inserting a new line below

**Long-Press-Actions:**
- [x] delete checkbox prefix (long-press on checkbox icon)
- [x] insert image (no dialog, just empty image syntax)
- [x] insert link (no dialog, just empty link syntax)


harshad1:
It would be nice if the long press to remove checklist could be added to markdown too.

I haven't felt the need for it personally, since I can just double tap the unordered list text action to achieve the same result

gsantner:
Ahead merged the small TOC changes, they seem rather unrelated to textactions itself

app/src/main/java/net/gsantner/markor/ui/AttachImageOrLinkDialog.java
```diff
@@ -216,6 +236,23 @@ public void onFsViewerConfig(FilesystemViewerData.Options dopt) {
         return builder.show();
     }
 
+    private static void copyFileToZimPageFolder(File fileToBeCopied, File currentWorkingFile, Activity activity) {
```
gsantner, line 239: Please keep Zim specific functions out of the dialog class. Use a common Zim class if required, but not spread it all over place.

fredericjacob:
@gsantner The PR is complete by my side and ready-for-review now. :)

@harshad1 Regarding the long-press to remove checkboxes: I could add the same functionality to markdown if desired, but I don't know if it really adds much value (like you said, it is possible to just tap the button two times). My motivation to add this was that Zim has four check states and not only two, so it's more convenient to long-press instead of having to toggle through all states first. @gsantner What is your opinion on that?

app/src/main/java/net/gsantner/markor/ui/AttachImageOrLinkDialog.java
fredericjacob: Changed: extracted to a utility class.

gsantner:
I don't know, the Attach xyu dialog's job should be to have input, and have some output. Can we reduce the amount of Zim (and / or markdown) related code in there?

Whats your thought @harshad1

harshad1:
> I don't know, the Attach xyu dialog's job should be to have input, and have some output. Can we reduce the amount of Zim (and / or markdown) related code in there? I just don't want to "infect" everything and all with just Zim.
> 
> Whats your thought @harshad1

I agree with @gsantner. I can see how maintaining this code can be hard in the future.

One common pattern which may work here is passing a callback which does the formatting and insertion. That way, markdown can pass in a markdown specific callback which lives in MarkdownTextActions and Zim can do the same. This way the dialog code lives in one place and any specific formatting needed for Zim or Markdown (or other future formats) can live in their own files.

Additionally, IMO we should *really* avoid very long ternary operators or ternary operators with more than one condition. @fredericjacob I would suggest assigning the format strings to final static variables and using normal if / then statements. It will take more lines but will be much easier to read. This may become unnecessary if we implement the callback approach, though.

Additionally, after some thought, I think we should not add the long press to remove checkbox to markdown at this stage. It does not add much functionality.

gsantner:
Just to add, the AUDIO / IMAGE constants we pass i.e. is a good first step, instead of the hardcoded "4" in switch cases :D

fredericjacob:
Yes, in fact the code IS already quite hard to maintain - because it's basically one large function and the code contains dependencies to variables which are scattered over the method body. That makes everything "interwoven" and also hard to separate parts of it to own methods (ideally, it should be split to multiple methods with single responsibilities, though).
This is also the reason why I had a hard time to figure out the bits of code that have to be changed for the zim format. It's not very obvious which piece does what and what the effect of a change will be.
So a bigger refactoring would make sense - however I was afraid of changing everything and then potentially breaking the markdown (or general) logic, so I tried to figure out where I can insert zim code without changing too much in the existing code. But I agree that this solution is not optimal.

Some of the functionality is not needed or need a quite different handling for zim, for these parts extracting the logic to format-specific callbacks is easy (e.g. the content of the `FilesystemViewerData.SelectionListenerAdapter` callback methods).
For other parts not, however, sometimes there are a few additions within the code that have to be made for zim only, so it's not trivial to find out which callbacks are suitable there.

Something that just came to my mind while thinking about it: the template method pattern might be a good fit here: https://refactoring.guru/design-patterns/template-method
So we would have an abstract base class `AttachImageOrLinkDialog` and format-specific subclasses e.g. `MarkdownAttachImageOrLinkDialog`, `ZimWikiAttachImageOrLinkDialog` and `GeneralAttachImageOrLinkDialog`. The base class defines the steps that have to be done (for now, most of the existing logic in `showInsertImageOrLinkDialog` could be adopted), but some code fragments in question are realized via abstract methods which are implemented by the subclasses.
I'd also introduce fields for most of the now local variables then which can be used across the methods.

Regarding the ternary operators + constants for format strings: I agree, this should be an easy fix and I can do this. (Many ternary operators won't be necessary when the template pattern is used, anyway.)

So I suggest going for the template method approach and adjusting ternary operators + extracting constants.
@gsantner @harshad1 What do you think?

gsantner:
I think it needs improvement for sure, but splitting that one yet simple dialog to 5 or more classes is not of my interest though.

app/src/main/java/net/gsantner/markor/ui/AttachImageOrLinkDialog.java
```diff
@@ -113,6 +118,13 @@ public static Dialog showInsertImageOrLinkDialog(final int action, final int tex
         final FilesystemViewerData.SelectionListener fsListener = new FilesystemViewerData.SelectionListenerAdapter() {
             @Override
             public void onFsViewerSelected(final String request, final File file) {
+
+                if (textFormatId == TextFormat.FORMAT_ZIMWIKI) {
+                    ZimPageFilePathUtil.copyFileToZimPageFolder(file, currentWorkingFile, activity);
+                    inputPathUrl.setText("./"+file.getName());
+                    return;
+                }
```
gsantner, line 126: This is the main thing of my dislike. It's free to automatic rewrite the filename, or the write directory...but don't reinvent the file copying stuff and, there should really not be a need here to make something Zim specific.

harshad1:
> So we would have an abstract base class `AttachImageOrLinkDialog` and format-specific subclasses e.g. `MarkdownAttachImageOrLinkDialog`, `ZimWikiAttachImageOrLinkDialog` and `GeneralAttachImageOrLinkDialog`. The base class defines the steps that have to be done (for now, most of the existing logic in `showInsertImageOrLinkDialog` could be adopted), but some code fragments in question are realized via abstract methods which are implemented by the subclasses.

Another lighter weight option is to pass in a specialization object, similar to your work on markdown prefix patterns / zim prefix patterns.

> I'd also introduce fields for most of the now local variables then which can be used across the methods.

This would be good, imo. Would also help to factor out some of the code into functions (static or otherwise). The block for extracting file paths for markdown looks like a good candidate, for example.

harshad1:
@fredericjacob @gsantner is there anything I can do to help this along?

fredericjacob:
@harshad1 If you would like to do some of the refactoring, you're welcome from my side. I would also do this, but I can't promise that I'll find time before next weekend.

gsantner:
Please just keep it for now. When I have time after my day-job or weekend I may look into shaping it in a manner which isn't too much changes, but at same time not makes feel like it's not the expected quality.

fredericjacob:
@gsantner Okay, does this also concern your review comment or only refactorings? Should I just keep the whole code now or still change something?

gsantner (approved):

gsantner:
hi @fredericjacob @harshad1 
I did various improvements. Overall it is better to have smaller pull requests and code change chunks to get the job done.

So:
* Kept most of the proposed changes
* Reworked the 2-4 way ternary expressions to use a K-V Map with default value instead.
* As already noted, totally not happy about Zim re-implementing file writing - so dropped @ THIS PR.
  * -> Removed the ZimUtil file and it's references.

Removed code is in history, I suggest to make another PR. (And by merging this one) a smaller PR that tackles the requirements of Zim on file structure level. Everything else is merged via this PR, so another one can be of a more specific scope, can be smaller and cover less different kind of areas.

fredericjacob:
@gsantner Thanks for merging the basic actions, I'll then look at the file insertion again in the future.

Issue #1190, opened 2021-01-10
KaTeX line break not working


radanovicnik:
#### General information

* **App version:** 2.5.0 (121) Google Play
* **System:** Android 10

#### Description

`\\` in math blocks just escapes the `\` instead of creating a line break, as is supposed to happen with KaTeX:

https://katex.org/docs/supported.html#line-breaks

This forces you to write `\newline` for breaking lines, which is (longer and) not compatible with editors that use MathJax for rendering.

[Example](https://user-images.githubusercontent.com/70813474/104121656-38bf3700-5340-11eb-8aab-5231fb8443e5.jpg)

gsantner:
MathJax and KaTex are similar but not the same, and you just found a difference. Given KaTex does it this way and MathJax not it is just like that - Markor has KaTex and you have to use it's syntax. This not sounds like any bug report bug a finding/statement...

radanovicnik:
Sorry, I explained it badly.
KaTeX does support `\\` for line break, it says so in the docs (link above).

For example, `\\` works in StackEdit, which also uses KaTeX.

`\\` just doesn't seem to work in Markor for some reason.

Issue #1154, opened 2020-12-09
Improve Markdown Latex/KaTex/Math when used with complex formulas


dhdgves:
Such as "$$ \left. \begin{array}{ll} \frac{\partial \sigma_p}{\partial \rho} + \frac{1}{\rho} \frac{\partial \tau_{\rho \varphi}}{\partial \varphi} + \frac{\sigma_\rho - \sigma_\varphi}{\rho} + f_\rho = 0 & \ \frac{1}{\rho}\frac{\partial \sigma_\varphi}{\partial \varphi} + \frac{\partial \tau_{\rho \varphi}}{\partial \rho} + \frac{2\tau_{\rho \varphi}}{\rho} + f_\varphi = 0 & \ \end{array} \right} $$"
I am sure this formula is in correct format. I can see my formula display well in Typora.

gsantner:
I pasted the $$ xyz $$ from above into other editors/demo sites that use MathJax. Does not show anything either there.

Not sure if paste error on your or my side, your formula is just wrong, formula is incompatible.

Please upload textfile (/.txt) in question to here.

PR #1196, opened 2021-01-16
Fixed KaTeX line break

radanovicnik:
Just fixed `\\` in LaTeX. It was printing `\` instead of creating a line break. 

**KaTeX docs:**

> Hard line breaks are `\\` and `\newline`.

https://katex.org/docs/supported.html#line-breaks

**Related issue:** #1190

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
+++ b/app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
@@ -181,7 +181,7 @@ public class MarkdownTextConverter extends TextConverter {
         if (appSettings.isMarkdownMathEnabled() && markup.contains("$")) {
             head += HTML_KATEX_INCLUDE;
             onLoadJs += JS_KATEX;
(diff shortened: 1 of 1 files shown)


gsantner:
Can you please verify if the markdown-reference document (which contains math), looks same with and without this, and renders correct?

radanovicnik:
Yes, renders perfectly.

Thank you for the awesome editor!

gsantner:
@radanovicnik 
Would you mind to also take a look at #1154 ?

gsantner:
@radanovicnik 

> Thank you for the awesome editor!

Welcome.


Would you mind to also take a look at #1154 ?  
Can you reproduce the issue? Looked fine on my side.

radanovicnik:
@gsantner

This formula looks like some backslashes were escaped:
- in the middle: should be `\\` instead of `\`
- at the end: should be `\right\}` instead of `\right}`

Both are caused by GitHub md editor.
This guy probably should have written the formula in a code block.

If you place the backslashes where they were, and put the formula in Markor, **it works fine (after this commit)**.
Before the commit, it would just escape the `\` and wouldn't render.

( This is how it looks: )
test_formula

gsantner:
OK then PR will close both issues.

Thanks, I merge it.

PR #1197, opened 2021-01-16
Improve wording of Markdown table string

edi9999:
"Many" should be used instead of "much" because the columns can be counted (2,3,4,...).

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -348,7 +348,7 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="reading">Reading</string>
     <string name="read">Read</string>
     <string name="table">Table</string>
-    <string name="how_much_columns_press_table_button_long_to_start_table">How much columns? Press table button long to start a table (insert header), press short to add a row.</string>
+    <string name="how_much_columns_press_table_button_long_to_start_table">How many columns? Press table button long to start a table (insert header), press short to add a row.</string>
     <string name="example_of_a_markdown_table">Example of a Markdown table</string>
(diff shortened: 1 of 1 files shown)


gsantner:
Please keep the string id and change only the english string. Translations are completely managed by translation service.

app/src/main/res/values/strings.xml
```diff
@@ -348,7 +348,7 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="reading">Reading</string>
     <string name="read">Read</string>
     <string name="table">Table</string>
-    <string name="how_much_columns_press_table_button_long_to_start_table">How much columns? Press table button long to start a table (insert header), press short to add a row.</string>
+    <string name="how_many_columns_press_table_button_long_to_start_table">How many columns? Press table button long to start a table (insert header), press short to add a row.</string>
```
gsantner, line 351: Please keep the string id and change only the english string. Translations are completely managed by translation service.

gsantner (changes requested):

app/src/main/res/values/strings.xml
edi9999: `how_much_columns_press_table_button_long_to_start_table` is technically also not totally correct, and it seems strange to use how_much for the string id, and then "how many" for the real text no ?

edi9999: "Many" should be used instead of "much" because the columns can be counted (2,3,4,...).

gsantner: As said, please update the english strings, as required. Keep the IDs however, they dont matter for this.

edi9999: Ok, done !

PR #1204, opened 2021-01-23
Prevent accessibility and autofill from crashing markor

harshad1:
Autofill, like accessibility, can cause crashes because the _entire text of the editor_ is sent to the autofill service for every change. Sending a large file 10s of times a second can cause a crash.

This PR adds logic to block autofill (not needed for editing a document anyway) when the file is large or on some functions which rapidly change the text.


gsantner:
merge master into PR branch for failure

harshad1:
@gsantner Subject to the CI system passing, this is ready to go.

harshad1:
@gsantner Whats with the JUnit stage?

gsantner:
I tried to get rid of it already, but github still remembers it somewhere. not even completley removing all builds and ci config removes it.......

harshad1:
> I tried to get rid of it already, but github still remembers it somewhere. not even completley removing all builds and ci config removes it.......

Ah. Wanted to make sure the issue was not with my pr

gsantner (approved):

gsantner:
Don't mind JUnit, only the the "build" is important. It also runs JUnit testcases. The JUnit github item is/was just for visualization of the testcase output.

harshad1:
Thanks for merging. Do you know approximately when this will make it into a release?

gsantner:
probably when we got zim finished

PR #1208, opened 2021-01-28
README: Various small improvements

xerus2000:
Just some things I observed


gsantner (approved):

xerus2000:
I don't think this sentence adds anything:

>  The app is working completly offline, no internet connection required!

it is inferrable from the other sentences and doesn't seem grammatically correct to me either.

gsantner:
I initially started writing a big comment to you about the changes but decided then to just rework it as thats faster.

I want to keep that point, offline focus is one of the most important points of Markor - for both usage and privacy.

xerus2000:
I do understand that, but isn't that obvious from the sentence before?

 > Markor does not use your internet connection unless external resources ...

Issue #1205, opened 2021-01-24
+/Add button is unresponsive after the app is open for a while


poisonborz:
#### General information

* **App version: 2.5** 
* **System: Android 8** 

#### Description

When the app is opened again after some sleep time on the phone (10+ minutes) tapping the red "+" button has no effect. The button does display the "hover" color, so the app surely registers the tap. The only solution is to restart the app.

#### Log

There were no new adb log entries created upon tapping "+".

harshad1:
When this happens:

1. Are hidden files / folders (i.e. files / folders starting with '.') visible?
2. Is the date in the file browser (i.e the second line) replaced with the path?

If so, I'm seeing this same issue.

poisonborz:
Yes and yes. Also folders are not clickable.

harshad1:
Yeah, my theory is that this is because the [savedInstanceState not being restored correctly](https://github.com/gsantner/markor/blob/ecf6cc17fe877d2daa805d9978fdad0116cc27d8/app/src/main/java/net/gsantner/opoc/ui/FilesystemViewerFragment.java#L140) / defaults not being sane

harshad1:
I will probably try to add a fix over the next few days.

harshad1:
@poisonborz Could you check if the following reproduces the issue:

1. Enable developer options -> Don't Keep Activities
2. Launch markor
3. Go home
4. Open markor from the recent apps list

You'll want to disable Don't keep activities afterwards :)

poisonborz:
The issue is still reproducible with `Don't Keep Activities` on (Markor was already in recent list ofc).

There is one additional thing, I don't know if it's related - list items were clickable now, but when a file was opened, Markor crashed.

PR #1210, opened 2021-02-02
Fixed logic to handle restore

harshad1:
This PR addresses #1205


gsantner (approved):
The blue line makes me slightly to worry, but I didn't notice a issue yet.

image

Issue #1212, opened 2021-02-02
Toggling checkbox causes autocorrect to insert word _after_ end


harshad1:
This is a strange issue best explained with a video:

![GIF-210202_124406](https://user-images.githubusercontent.com/1940940/106661686-5ace6280-6556-11eb-9f24-ff8dbf57df52.gif)

Essentially:
1. Insert checkbox
2. type a word which will be autocorrected (do not press space or enter to insert autocorrection)
3. Press the checkbox action (will toggle check)
4. Press enter
5. Instead of the word being corrected, the correction is _appended_

harshad1:
Further investigation has shown that this happens because
1. `regexReplaceAction` replaces the _entire line_
2. It happens only when the line length has not changed

PR #1213, opened 2021-02-03
Limit replacement to ROI in regexReplace

harshad1:
This PR addresses #1212


harshad1:
@gsantner this works, but i plan to do a lot more testing as regexReplace is used _everywhere_

harshad1:
@gsantner this is ready

Issue #1219, opened 2021-02-05
Live search for files


harshad1:
The current file search is async. It would be nice to have a synchronous file search which updates as you type, similar to the search _within_ files.

This search system would have to fall back to the async search if the total number of files exceeded a certain limit (say 5000).

PR #1222, opened 2021-02-06
search within search results

harshad1:
This minimal change addresses #1219

In order to live search for files one now just has to:
1. search for `*` or any other 'filter'
2. Search through the results as you type for the exact file needed


gsantner:
nice idea yet simple addition, cool

PR #1226, opened 2021-02-09
Remember last used file format, closes #1220

### LINKED ISSUE #1220: Fix per-file setting of Format
#### General information

* **App version:  2.5.0
* **System: 11

#### Description
Whenever I set the formatting of a note with a .txt file extension (via the three-circles-menu-item>File settings>Format >Markdown menu item) and then reload or close and then return to that note, the formatting of the note is reset to plaintext. Can you please help me resolve this issue? 
Also, is there a way to have Markor load all .txt files as if they were Markdown files (eg: by showing the Markdown Text Actions bar above the keyboard rather than the Plaintext one) by default? If not, would you consider adding an option for doing so? Thanks!!

harshad1:
This will address #1220

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
@@ -320,7 +320,7 @@ public class DocumentEditFragment extends GsFragmentBase implements TextFormat.T
             final int id = formats.getItem(i).getItemId();
             _formatMap.put(id, menu.findItem(id));
         }
-        _formatMap.get(_document.getFormat()).setChecked(true);
+        setFormatToggle();
     }
 
     public void loadDocumentIntoUi() {
@@ -558,10 +558,6 @@ public class DocumentEditFragment extends GsFragmentBase implements TextFormat.T
     }
 
     public void applyTextFormat(final int textFormatId) {
-        // Checkable = single => only one is checked
-        final MenuItem formatItem = _formatMap.get(textFormatId);
-        if (formatItem != null) formatItem.setChecked(true);
-
         _textActionsBar.removeAllViews();
         _textFormat = TextFormat.getFormat(textFormatId, getActivity(), _document, _hlEditor);
         _hlEditor.setHighlighter(_textFormat.getHighlighter());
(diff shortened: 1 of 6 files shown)


app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -396,8 +410,9 @@ public boolean getDocumentHighlightState(final String path, final CharSequence c
         return getBool(PREF_PREFIX_HIGHLIGHT_STATE + path, lengthOk && isHighlightingEnabled());
     }
 
-    public int getLastEditPositionChar(File file) {
-        if (file == null || !file.exists()) {
+    public int getLastEditPositionChar(final String path) {
```
harshad1, line 413: normalizing all per-file settings to take a path

harshad1: Reverted this change - unnecessary

harshad1:
@gsantner I think we should add a check next to the current format in the menu. i.e. to show what the currently selected format is.

gsantner:
I had exactly that already local, not sure why I stashed it. think there is some special case which needs attention.

harshad1:
My approach seems to be working correctly. Would be good if you could test it. Also, please let me know if you remember what the special case was.

gsantner:
probably it was device rotation, app switchting, onResume and savedInsance

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -543,7 +557,11 @@ private Document loadDocument() {
         return document;
     }
 
-    public void applyTextFormat(int textFormatId) {
+    public void applyTextFormat(final int textFormatId) {
+        // Checkable = single => only one is checked
+        final MenuItem formatItem = _formatMap.get(textFormatId);
+        if (formatItem != null) formatItem.setChecked(true);
```
gsantner, line 563: {

app/src/main/java/net/gsantner/opoc/preference/SharedPreferencesPropertyBackend.java
```diff
@@ -588,4 +589,12 @@ public static String getDebugLog() {
...
         _debugLog += "[" + new Date().toString() + "] " + text + "\n";
     }
+
+    public static boolean ne(final String str) {
+        return str != null && !str.trim().isEmpty();
+    }
+
+    public static boolean fexists(final String fp) {
+        return ne(fp) && (new File(fp)).exists();
+    }
```
harshad1, line 599: Ah. nice. :+1:

harshad1:
@gsantner This is ready, imo

harshad1:
@gsantner are there any changes required to this pr?

I'm looking forward to seeing all these recent changes in a release 😁

PR #1241, opened 2021-02-16
Improve scrollbar grab width on hidpi devices

harshad1:
Currently, the draggable scrollbar grab width uses the raw pixel width of the scoll bar. On higher DPI devices (which is most devices these days) the grab region can be too small.

For example, the default on my pixel 3 XL is 0.7mm.

This PR uses a better heuristic which accounts for DPI

### DIFF
--- a/app/src/main/java/net/gsantner/markor/ui/DraggableScrollbarScrollView.java
+++ b/app/src/main/java/net/gsantner/markor/ui/DraggableScrollbarScrollView.java
@@ -15,7 +15,7 @@ public class DraggableScrollbarScrollView extends ScrollView {
     private boolean _fastScrollEnabled = true;
     private boolean _ltr = true;
     private int _thumbHeight;
-    private int _scrollbarWidth;
+    private int _grabWidth;
 
     public DraggableScrollbarScrollView(Context context) {
         super(context);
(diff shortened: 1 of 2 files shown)


harshad1:
@gsantner `DraggableScrollbarScrollView.java` and `DraggableScrollbarWebView.java` are 95% identical. We should probably find a way to de-duplicate them,

app/src/main/java/net/gsantner/markor/ui/DraggableScrollbarScrollView.java
```diff
@@ -80,7 +81,8 @@ protected void onAttachedToWindow() {
         if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.JELLY_BEAN_MR1) {
             _ltr = getLayoutDirection() == View.LAYOUT_DIRECTION_LTR;
         }
-        _scrollbarWidth = getVerticalScrollbarWidth();
+        final DisplayMetrics displayMetrics = getContext().getResources().getDisplayMetrics();
+        _grabWidth = (int) (2.0 * (float) getVerticalScrollbarWidth() * displayMetrics.density);
```
harshad1, line 85: @gsantner It would be good to test this on other devices

harshad1: I have tested this in the emulator and it seems ok

gsantner: working here as well

gsantner (approved):

gsantner:
thanks for so much help and your time. Everything merged. Excpet the last used format, not sure yet if its all fine.

harshad1:
> thanks for so much help and your time. Everything merged. Except the last used format, not sure yet if its all fine.

No problem. 99% of my PRs are things that I want or that directly affect my usage, win-win. Last used format should be ready to merge too

PR #887, opened 2020-03-31
Add option to encrypt files, fixes #60

opensource21:
Added encryption (#867).

Added the ability to save the content AES encrypted based on a password
Added the ability to store the password in a safe way (PasswordStore is for years in production at turtl)
You can only store one password at the moment. So if you need different keys for different documents you must change the one password before open the document.

Add the moment the extension is jenc, but there are only 2 Files with this information, so it could be easily changed.
Please be so kindly to give me feedback what should be improved.

PR #1244, opened 2021-02-17
Backup and restore settings, closes #839

### LINKED ISSUE #839: App Settings backup & restore
It will be useful if we can export and import app settings (such as `settings.json`).

harshad1:
This PR addresses #839. ~~I have liberally borrowed from [OpenLauncher/BackupHelper](https://github.com/OpenLauncherTeam/openlauncher/blob/ec793b34672e2913662e016978d6ac7bb5ff1e6c/app/src/main/java/com/benny/openlauncher/util/BackupHelper.java)~~

This PR now does not export a zip of the shared preference XML files (as openlauncher does). Instead, the preferences are exported as a JSON.


harshad1:
@gsantner This works, but we need to review strings etc

app/src/main/java/net/gsantner/markor/activity/BackupRestoreHelper.java
```diff
@@ -0,0 +1,185 @@
...
+public class BackupRestoreHelper {
+
+    private static final String SAVE_NAME = "markor_settings_backup%s.zip";
+    private static final int BUFFER_SIZE = 2048;
+
+    private static String[] BACKUP_PATHS = {
+            "shared_prefs/action_order.xml",
+            "shared_prefs/app.xml",
+            "shared_prefs/datetime_dialog_settings.xml",
+            "shared_prefs/search_replace_dialog_settings.xml",
```
gsantner, line 35: suggestiong whole folder, and excluding. instead of whitelist

app/src/main/res/xml/preferences_master.xml
```diff
@@ -557,4 +557,17 @@
...
+            android:icon="@drawable/ic_save_black_24dp"
+            android:key="@string/pref_key__backup_settings"
+            android:title="@string/backup_settings"
+            android:summary="@string/backup_save_summary" />
+        <Preference
+            android:icon="@drawable/ic_restore_black_24dp"
+            android:key="@string/pref_key__restore_settings"
+            android:title="@string/restore_settings"
+            android:summary="@string/restore_backup_summary" />
+    </PreferenceCategory>
```
harshad1, line 571: I have this under the main settings. Not sure if it is better here or under one of the other submenus.

app/src/main/java/net/gsantner/markor/activity/BackupRestoreHelper.java
```diff
@@ -0,0 +1,185 @@
...
+        for (final String path : BACKUP_PATHS) {
+            files.add(new File(dataDir, path.replace("%PACKAGENAME%", packageName)));
+        }
+
+        return files;
+    }
+
+    public static void backupConfig(final Context context, final FragmentManager manager) {
+
+        if (context instanceof Activity) {
```
harshad1, line 54: User is asked for export directory

app/src/main/java/net/gsantner/markor/activity/BackupRestoreHelper.java
```diff
@@ -0,0 +1,185 @@
...
+            }
+            Toast.makeText(context, R.string.toast_backup_error, Toast.LENGTH_SHORT).show();
+        }
+    }
+
+    public static void restoreConfig(final Context context, final FragmentManager manager) {
+        if (context instanceof Activity) {
+            final Activity activity = (Activity) context;
+
+            FilesystemViewerCreator.showFileDialog(
```
harshad1, line 112: User is asked to select backup file

app/src/main/res/xml/preferences_master.xml
gsantner: First = Important. Second  = Various stuff people barley need

app/src/main/java/net/gsantner/markor/activity/BackupRestoreHelper.java
harshad1: I thought of this, but this causes some additional complexity.

When restoring the backup we will have to have a whitelist anyway, otherwise we will have to iterate over the contents of the zip file and guess whether a file should be restored or not.

app/src/main/res/xml/preferences_master.xml
```diff
@@ -233,6 +233,19 @@
...
+                android:icon="@drawable/ic_save_black_24dp"
+                android:key="@string/pref_key__backup_settings"
+                android:title="@string/backup_settings"
+                android:summary="@string/backup_save_summary" />
+            <Preference
+                android:icon="@drawable/ic_restore_black_24dp"
+                android:key="@string/pref_key__restore_settings"
+                android:title="@string/restore_settings"
+                android:summary="@string/restore_backup_summary" />
+        </PreferenceCategory>
```
harshad1, line 247: @gsantner Moved to under 'other'

gsantner: however you think it fits better

gsantner:
what do you think of adding SearchOrCustomDialog after file selection?

Data / Data +Settings / Settings

and for data, also add the notebook folder recursive to zip

harshad1:
I think data backup etc is best handled by an external app.

For example, it's pretty easy to compress a folder using any file manager app.

Feb. 18, 2021 01:03:22 Gregor Santner <notifications@github.com>:

app/src/main/java/net/gsantner/markor/activity/BackupRestoreHelper.java
harshad1: @gsantner Have found a good fix - using both a blacklist and a whitelist. This works well and is now ready for review / merge

gsantner:
btw, the password is not contained in the export right (->keystore)? If so, we should strip that one

harshad1:
> btw, the password is not contained in the export right (->keystore)? If so, we should strip that one

Ah good catch. I will address this

gsantner:
@opensource21 
Is it enough to decrypt with these two values? Just recall we (try to) store something in android's key storage, outside of app files.

gsantner:
then it might not work as expected. should be better with `if (pref key is set) { key.remove }` in onStart/onResume

harshad1:
IMO, a much cleaner solution would be to use a regular dialog to set the password. That way it is never stored in plaintext and we don't have to do this delete after storing in password manager thing.

opensource21:
I started with a separate dialog https://github.com/gsantner/markor/pull/887/commits/36d7078107f05bc438a109c3bbfe10969d0ffc4b.
I didn't remember the argument against.

opensource21:
Just to get it clear `pref_key__default_encryption_password` always contains the 3 asterisks "***" and never the real password correct? This would I expect. Then it would be save to put them into the backup file it's not really a risk, because you only have encrypted garbage, which is useless without the real key. Obviously a brute-force could be easier done if you have it in a backup-file.
@gsantner But to remove the 3 keys from export with !key.contains(default_encryption_password) would be fine too.

app/src/main/java/net/gsantner/markor/util/BackupRestoreHelper.java
```diff
@@ -0,0 +1,203 @@
...
+        } while (file.exists());
+        return file;
+    }
+
+    public static void createAndSaveBackup(final Context context, final File saveLoc) {
+        try {
+            final FileOutputStream fos = new FileOutputStream(saveLoc);
+            final BufferedOutputStream bos = new BufferedOutputStream(fos);
+            final ZipOutputStream zos = new ZipOutputStream(bos);
+            final File[] files = getPrefDir(context).listFiles();
```
opensource21, line 112: Why not simply use _preferences.getAll() instead of reading file?

harshad1:
This pr includes code to handle password entry with a separate dialog. It appears to work correctly and completely side side steps this issue. If there isnt a compelling reason to not do this, then going with a separate dialog would be my preference.

app/src/main/java/net/gsantner/markor/util/BackupRestoreHelper.java
harshad1: I'm just copying / exporting the shared preference XML files as is. Not reading them or parsing their contents.

app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
```diff
@@ -94,7 +94,7 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
         final Spinner templateSpinner = root.findViewById(R.id.new_file_dialog__template);
         final String[] typeSpinnerToExtension = getResources().getStringArray(R.array.new_file_types__file_extension);
 
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT && appSettings.hasPasswordBeenSetOnce()) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && !TextUtils.isEmpty(appSettings.getPassword())) {
```
harshad1, line 97: Just checking if password exists. This avoids issues with `hasPasswordBeenSetOnce` being out of sync with actual status after restore of backup

harshad1:
@gsantner @opensource21 This is good to go imo

app/src/main/java/net/gsantner/markor/util/BackupRestoreHelper.java
opensource21: Are you sure that this is an official api? What happens if a future android version will change the file format?
What if the file access will change again? I would recommend to only use official documented api.

app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
opensource21: @gsantner Sure that this will work?

harshad1: It seems to work fine for me. Is there a specific bug you anticipate / see here?

app/src/main/java/net/gsantner/markor/util/BackupRestoreHelper.java
harshad1: 1. If Android breaks the File api, much of markor will be broken. I am comfortable using this.
2. `ApplicationInfo` / `dataDir` appear to be documented
3. As pointed out above, much of this logic is lifted from OpenLauncher.

gsantner: I'm 99% sure internal File access will stay, for private app store, so including prefs.

For external storage it might be different.

gsantner:
Do you guys feel confident that it won't result in any password loss & potential file damage / things lost ("forgot password that was entered before")?

Otherwise we merge of this PR somewhen **after** update release

harshad1:
> Do you guys feel confident that it won't result in any password loss & potential file damage / things lost ("forgot password that was entered before")?
> 
> Otherwise we merge of this PR somewhen **after** update release

I've found several cases where the password can be overwritten etc. Not sure how to fix these for now. I think it is best if we do a lot more testing and merge this after the release

gsantner:
agree

harshad1:
@opensource21 I've updated this PR to export settings as a json with `getAll()`. This is more robust to changes in the underlying API and we can parse settings and filter them etc. Your opinion / review would be appreciated

harshad1:
@gsantner @opensource21 This is ready IMO

app/src/main/java/net/gsantner/markor/util/BackupRestoreHelper.java
```diff
@@ -0,0 +1,211 @@
...
+                for (final String key : map.keySet()) {
+                    if (includeKey(key)) {
+                        final Object value = map.get(key);
+                        if (
+                            (value instanceof Integer) ||
+                            (value instanceof Float) ||
+                            (value instanceof String) ||
+                            (value instanceof Boolean)
+                        ) {
+                            prefSon.put(key, value);
```
opensource21, line 114: Have you checked that the type is stored in the Json-Object or does the information be lost.

app/src/main/java/net/gsantner/markor/util/BackupRestoreHelper.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    public static void createAndSaveBackup(final Context context, final File saveLoc) {
+        try {
+            final JSONObject json = new JSONObject();
+            for (final String _pref : PREF_NAMES) {
+                final String pref = getPrefName(context, _pref);
+                final SharedPreferences sp = context.getSharedPreferences(pref, Context.MODE_PRIVATE);
+                final Map<String, ?> map = sp.getAll();
+                final JSONObject prefSon = new JSONObject();
```
opensource21, line 104: I wonder if it wasn't better to add some version information. This would allows to migrate the config from one version to another.

opensource21 (commented):
Sorry for the late response. I'm very busy at the moment. I found the new approach much cleaner. Please have a look to the further comments.
I only looked at the BackupRestoreHelper.

app/src/main/java/net/gsantner/markor/util/BackupRestoreHelper.java
harshad1: The json correctly stores boolean string and number.

app/src/main/java/net/gsantner/markor/util/BackupRestoreHelper.java
harshad1: Ah good point.

harshad1:
Noticed a bug in our system: A couple settings in AppSettings directly save a resource string option as an integer, rather than as a string. This is brittle. Will be addressing this in this PR.

gsantner:
by the way, there is a easier way then zip of xml: concat all xml files to one

app/src/main/java/net/gsantner/markor/util/BackupRestoreHelper.java
harshad1: @opensource21 Have added a version code to the json

app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -371,15 +379,30 @@ public boolean getDocumentWrapState(final String path) {
 
     public void setDocumentFormat(final String path, final int format) {
```
harshad1, line 380: These routines have been updated to store and retrieve format by _index_. This is now robust to the res value changing on recompilation

harshad1:
> by the way, there is a easier way then zip of xml: concat all xml files to one textfile

We have switched from exporting a ZIP to exporting a Json of all key:value pairs. This helps avoid accidentally exporting or overwriting passwords etc

app/src/main/java/net/gsantner/markor/util/BackupRestoreHelper.java
```diff
@@ -0,0 +1,217 @@
...
+                for (final String key : map.keySet()) {
+                    if (includeKey(key)) {
+                        final Object value = map.get(key);
+                        if (
+                                (value instanceof Integer) ||
```
gsantner, line 113: @harshad1   
what about `long`? not that we much use it currently, but json to my understanding does not support long. however preference input (and it's recovery) is potentially a `long`

harshad1: Long appears to be a valid input / output for the java json library. I have added it as an option. I expect that if it is supported int the library it will be saved and retrieved correctly

app/src/main/java/net/gsantner/markor/util/BackupRestoreHelper.java
```diff
@@ -0,0 +1,236 @@
...
+    private static final String[] PREF_NAMES = {
+            null, // Default pref
+            AppSettings.SHARED_PREF_APP,
+            DatetimeFormatDialog.DATETIME_SETTINGS,
+            TextActions.ACTION_ORDER_PREF_NAME,
+            SearchReplaceDialog.SEARCH_REPLACE_SETTINGS
+    };
+
+    private static final Pattern[] PREF_EXCLUDE_PATTERNS = {
+            Pattern.compile(".*password.*", Pattern.MULTILINE | Pattern.CASE_INSENSITIVE),
```
harshad1, line 46: @gsantner we can't leave this so generic because it will block any per file settings for any file with 'password' in the name.

This affects me, and others too,  I'm sure

gsantner: And what happens then? File settings are taken then for such files with global default values.

gsantner: And what happens then? File settings are taken then for such files with global default values.

On top of that, it will not expose the filepath to your plaintext passwords file you have still laying around somewhere. (Everybody has a passwords txt somewhere, even if its just for test accounts)

app/src/main/java/net/gsantner/markor/util/BackupRestoreHelper.java
gsantner: Beside version info of app, also backup-json format revision number (constant 1) is there now

just in case we make changes to exchange protocol we can evaluate this value.

app/src/main/java/net/gsantner/markor/util/BackupRestoreHelper.java
harshad1: I am of the opinion that we shouldn't include such hidden special cases or presume the user's intent.

How about if we add a negative look-behind to allow keys which start with `PREF_PREFIX_`? That will allow per file settings while still catching all other possible password related keys.

gsantner: i'd prefer everything with password, but if you say OK with pref_ prefix...also OK

gsantner:
@harshad1 

I would like to use the new  BackupRestore class also in my other open source apps, like dandelion, and maybe also reworking openlauncher.

Would you be okay with relicensing that file to shared code `opoc/utils/`, like ContextUtils?

That usually comes with increase on code quality, documentation, reusability and use safety / overall improvements. Thus if I work on dandelion and fix stuff I find, I also improve Markor at same time :smiley:

harshad1:
Your changes look good.

> @harshad1
> 
> I would like to use the new BackupRestore class also in my other open source apps, like dandelion, and maybe also reworking openlauncher.
> 
> Would you be okay with relicensing that file to shared code `opoc/utils/`, like ContextUtils?
> 
> That usually comes with increase on code quality, documentation, reusability and use safety / overall improvements. Thus if I work on dandelion and fix stuff I find, I also improve Markor at same time smiley

Sure. I would be fine with that. Apache - 2? Anything I need to do to enable this?

gsantner:
I will move/rename it later on the computer

harshad1:
> I will move/rename it later on the computer

Cool. I have pushed up my fix for the password regex. Will pull after you've moved, if needed

app/src/main/java/net/gsantner/markor/activity/SettingsActivity.java
```diff
@@ -199,10 +199,10 @@ public void doUpdatePreferences() {
...
 
             setPreferenceVisible(R.string.pref_key__is_multi_window_enabled, Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP);
-            setPreferenceVisible(R.string.pref_key__default_encryption_password, Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT);
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && _as.hasPasswordBeenSetOnce()) {
-                updateSummary(R.string.pref_key__default_encryption_password, "****");
-                setDialogMessage(R.string.pref_key__default_encryption_password, getString(R.string.password_already_set_setting_a_new_password_will_overwrite));
+
+            setPreferenceVisible(R.string.pref_key__encryption_password, Build.VERSION.SDK_INT >= Build.VERSION_CODES.M);
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && !TextUtils.isEmpty(_as.getPassword())) {
+                updateSummary(R.string.pref_key__encryption_password, getString(R.string.hidden_password));
```
gsantner, line 205: @harshad1  / @opensource21  
Really required to change the settings key here? Want to make sure people not get their password reset somehow.  
Also I'd say `default_..._password` is still a valid name to use, because that is what it is - default password for everything.  
There is user interest for entering the encryption password per-file (`Per File Settings`) and/or forgetting the general password. 

Thus, we might have in future `Per-File password` and `Global default password`, similar to `Syntax Highlighting, Format Per-File`.

gsantner:
@harshad1 

I'm done with what I found to improve. Do you have something remaining or ready for merge?  
From me, only one point is open

app/src/main/java/net/gsantner/markor/activity/SettingsActivity.java
harshad1: I have reverted the key used to store the password to `pref_key__default_encryption_password`.

gsantner:
I found a issue with the recent changes of format selection.  
I.e. I opened markor and main "ToDo" didn't have any format set (no divider lines).  
Manually set it, then swiped to quicknote, saw that nothing is preselected (-> I'm aware of a "undefined" state which is no format - we should get rid of that).

So guess this new, third format selection preference setting doesn't work as expected.  
What do you think of getting rid of all of it alltogether and have just a traditional string for selection in settings ("markdown", "todotxt", "zim", ..) - and map that to Format dropdown for example?  
Potentially we can use the syntax highlighter getClass.getName(), this is effectifley always the formats correct selection.

Screenshot_20210323-145107_Trebuchet

harshad1:
@gsantner It may be due to the issue I mentioned here earlier - we were saving format by integer, instead of by string. Because of this, the format retrieval can break if the resource integer values change during recompilation. I have fixed this issue in this PR.

gsantner:
> Because of this, the format retrieval can break if the resource integer values change during recompilatio

I can nothing but agree, totally right. However I'm not sure if not a string is better instead of index

harshad1:
> I can nothing but agree, totally right. However I'm not sure if not a string is better instead of index

Its a trade-off. A string is more explicit, but now we need to map strings to and from the int values. In the end, index approach seemed to be cleaner and introduced less new stuff + was less work.

gsantner:
Than index based. However the unknown/undefined state should not be possible / exist

harshad1:
> However the unknown/undefined state should not be possible / exist

Yeah. I need to investigate and possibly remove it.

gsantner:
If we cannot detect something, PlainText should be fine I guess

harshad1:
> If we cannot detect something, Plaintext should be fine I guess

Looks like this is the current behavior. When a document is first created, it is set to `FORMAT_UNKNOWN`. Then, in `loadDocument`, if format is `FORMAT_UNKNOWN`, the file name is parsed to detect the correct format. If nothing matches, the fallback is Plain text.

i.e. `FORMAT_UNKNOWN` is only used to indicate that the format hasn't been detected yet and theoretically no file should have `FORMAT_UNKNOWN` after being loaded. Given this, I am happy with the code as it is currently

app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -763,12 +782,10 @@ public String getHuuidDeviceId() {
...
     }
 
-    public boolean hasPasswordBeenSetOnce() {
-        return getBool(R.string.pref_key__default_encryption_password_set_once, false);
-    }
-
-    public void setPasswordHasBeenSetOnce(boolean b) {
-        setBool(R.string.pref_key__default_encryption_password_set_once, b);
+    @RequiresApi(api = Build.VERSION_CODES.M)
+    public String getPassword() {
```
opensource21, line 786: Would be better to use char[] here. Lesser conversion and lesser String-Caching.

app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
opensource21: Why not simply add a method appSettings.isPasswortSet()? Then the information how it is stored is encapsulate at one place. We could have an extra-field or check if the char[] is null.

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
```diff
@@ -442,6 +446,27 @@ public static void showPriorityDialog(Activity activity, char selectedPriority,
...
+            baseConf(activity, dopt);
+            dopt.isSearchEnabled = true;
+            dopt.titleText = R.string.file_encryption_password;
+            final boolean hasPassword = !TextUtils.isEmpty(as.getPassword());
+            dopt.messageText = hasPassword ? activity.getString(R.string.password_already_set_setting_a_new_password_will_overwrite) : "";
+            dopt.searchHintText = hasPassword ? R.string.hidden_password : R.string.empty_string;
+            dopt.callback = password -> {
+                if (!TextUtils.isEmpty(password)) {
+                    final String key = activity.getString(R.string.pref_key__default_encryption_password);
+                    new PasswordStore(activity).storeKey(password, key, PasswordStore.SecurityMode.NONE);
```
opensource21, line 462: Wouldn't it be besser to have an AppSettings.storePassword?

harshad1: I agree. Would you be able to push these two changes to this PR?

app/src/main/java/net/gsantner/markor/util/AppSettings.java
gsantner: To be honest, a char[] to string and vice versa, of lets say 64 chars max .... is not really performance critical.  
operations on the user file (text content) is important.

Whatever is easier working for the use here [[ less code ]] , is better.

opensource21: string vs char[] is not an performance issue, it's an security issue - at least at normal java runtimes.

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
opensource21: @harshad1 I can't make changes to this PR, but I create a PR for your repo, so be so kindly an merge it into your branch.

harshad1: @opensource21 Thanks. Merged

harshad1:
@gsantner @opensource21 I think we are good to go :)

gsantner (changes requested):
I'm sorry that I have to block this still, but it does not work. Tried it. 

* Set 22 as value just above the backup settings. exited app to make sure it saved the 22.
* then made backup
* Changed to 10, exited again to make sure it's saved and not pending.
* Then restored the backup. Checked settings and it's still at 10 - it should be 22 now.
* I also exited / restarted the app multiple times after loading the backup - the value is still at 10, so backup-restore didn't do anything.

`https://user-images.githubusercontent.com/6735650/112771640-1da1c080-902d-11eb-9e0e-28bf076ee997.mp4`

https://user-images.githubusercontent.com/6735650/112771640-1da1c080-902d-11eb-9e0e-28bf076ee997.mp4

Screenshot_20210329-011009 Screenshot_20210329-011137

This is probably the price of converting back-and-forth formats, potential loss of information.

harshad1:
For some reason, the `.apply()` does not appear to consistently apply the results before `exit()` after the recent changes. I have reverted to `.commit()`

@gsantner could you check importing now?

harshad1:
> This is probably the price of converting back-and-forth formats, potential loss of information.

To be clear, the issue was not data loss (all data appears to be preserved with 100% fidelity), but that `.apply()` would not finish writing to disk before `sys.exit()`

This is fixed now

gsantner:
may I suggest to replace

```
sys.exit()
```

with 

```
postDelayed( lambda()->sys.exit(), 150ms)
```

then we might be more certain the disk i/o has been finished with writing - even though the preference method call already should have ensured it.

harshad1:
It appears we may not even need the system exit. Could you test and confirm?

harshad1:
After much testing, I have decided that I agree with @gsantner and that we should store the file type as strings rather than indices. I will be making these changes later today.

gsantner:
OK as you prefer, if you think indices is good enough, also OK if it works

app/src/main/java/net/gsantner/markor/activity/ActionOrderActivity.java
```diff
@@ -146,20 +146,14 @@ protected void onPause() {
...
-                _textActions = new ZimWikiTextActions(this, null);
-                break;
+        if (documentType == R.string.pref_key__markdown__reorder_actions) {
+            _textActions = new MarkdownTextActions(this, null);
+        } else if (documentType == R.string.pref_key__todotxt__reorder_actions) {
+            _textActions = new TodoTxtTextActions(this, null);
+        } else if (documentType == R.string.pref_key__zimwiki__reorder_actions) {
+            _textActions = new ZimWikiTextActions(this, null);
+        } else { // Default to Plaintext
+            _textActions = new PlaintextTextActions(this, null);
```
harshad1, line 156: Just using the pref key directly simplifies this

using if-else as recommended by google

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -446,11 +446,11 @@ public boolean onOptionsItemSelected(final MenuItem item) {
...
-            case R.id.action_format_zimwiki:
-            case R.id.action_format_keyvalue:
-            case R.id.action_format_todotxt:
-            case R.id.action_format_plaintext:
-            case R.id.action_format_markdown: {
+            case R.string.action_format_zimwiki:
+            case R.string.action_format_keyvalue:
+            case R.string.action_format_todotxt:
+            case R.string.action_format_plaintext:
+            case R.string.action_format_markdown: {
```
harshad1, line 453: Switched all type formats to strings

app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -29,34 +32,38 @@
...
+import other.de.stanetz.jpencconverter.PasswordStore;
+
+import static net.gsantner.markor.format.TextFormat.FORMAT_UNKNOWN;
+
+@SuppressWarnings({"SameParameterValue", "WeakerAccess", "FieldCanBeLocal"})
 public class AppSettings extends SharedPreferencesPropertyBackend {
     private final SharedPreferences _prefCache;
     private final SharedPreferences _prefHistory;
     public static Boolean _isDeviceGoodHardware = null;
+    private final ContextUtils _contextUtils;
```
harshad1, line 48: Created a class level `_contextUtils` since it is used in multiple places.

@gsantner I have not called `freeContextRef`. All it seems to do is set the internal context reference to null. Is it really necessary?

app/src/main/java/net/gsantner/opoc/preference/SharedPreferencesPropertyBackend.java
```diff
@@ -202,11 +202,15 @@ public String getString(@StringRes int keyResourceId, @StringRes int defaultValu
     }
 
     public String getString(String key, String defaultValue, final SharedPreferences... pref) {
-        return gp(pref).getString(key, defaultValue);
+        try {
+            return gp(pref).getString(key, defaultValue);
+        } catch (ClassCastException e) {
+            return defaultValue;
```
harshad1, line 208: Added defensive code to all the get methods in this class. This prevents an exception when `setInt(key)` followed by `getString(key)` etc.

app/src/main/java/net/gsantner/opoc/preference/SharedPreferencesPropertyBackend.java
```diff
@@ -278,11 +282,15 @@ public void setInt(String key, int value, final SharedPreferences... pref) {
     }
 
     public int getInt(@StringRes int keyResourceId, int defaultValue, final SharedPreferences... pref) {
-        return gp(pref).getInt(rstr(keyResourceId), defaultValue);
+        return getInt(rstr(keyResourceId), defaultValue, pref);
```
harshad1, line 285: Deferring to the overload which includes the exception handling

app/src/main/java/net/gsantner/markor/util/AppSettings.java
harshad1: ?

app/src/main/java/net/gsantner/markor/activity/ActionOrderActivity.java
gsantner: Not looks much better to me, but I don't care, ok

app/src/main/java/net/gsantner/markor/util/AppSettings.java
gsantner: not strictly, but generally speaking no references to Context's and Activity's  is better than references that cannot be freed.

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
gsantner: R.string is again a build-time generated, non constant value, not?

gsantner:
Note, this is how the format is saved now in the end:

grafik

gsantner (approved):

Issue #1242, opened 2021-02-16
Improve editor cursor - (at file open & at edit/view swap)


gsantner:
#### General information

* **App version:** 2.5
#### Description
Combined issue for improving mouse/cursor position handling at editor.

Replaces:

* #1238
* #1237 
* #1149 
* #1004

Current thoughts:

* On Main page with the 4 tabs - always start on bottom. Not affected by any settings. Especially important to see the hot "ShareInto" stuff immediately after appeneded.
* Remember cursor position only for 12 hours (ie save timestamp). When file was last opened 12h+ ago, ignore remembered position.
* If nothing else, use the preferred start position from settings (start on top/bottom).

krisfremen:
Hey @gsantner,

As I mentioned in the other issue did a screencap, maybe it will reveal the issue:

This is from latest master, but same behaviour as v2.5.0.

https://user-images.githubusercontent.com/1108071/108118775-d6f39a80-706c-11eb-8576-78f1a9892beb.mp4

Cheers!

krisfremen:
hey @gsantner,

I ended up digging into the code for a bit, did some quick changes, let me know if this is good or not.

At least this seems as the most user friendly behaviour while respecting the app settings.

Happy to change anything if you have any comments.

Cheers.

gsantner:
There has been numerous improvements on this over the last couple of months. Closing.

PR #1245, opened 2021-02-18
WIP: Improve editor cursor.

krisfremen:
Improve editor cursor.

ToDo/QuickNote tabs and files use the last position based on app setting first, position second.

Partial fix for #1242.


gsantner:
as noted in other ticket, I want to keep always on bottom on the main tabs.

harshad1:
> as noted in other ticket, I want to keep always on bottom on the main tabs.

@krisfremen Currently this is solved by always starting quicknote and todo at the bottom. A better way would be start at the bottom is by checking if `activity instanceof MainActivity`. This way, qn and todo will start at the bottom in the main tab view but will respect the user's setting if started from a shortcut or through the file manager.

gsantner:
I'm closing for now due inactivity - we can always reopen later when improved

PR #1281, opened 2021-03-14
Move/Copy file improvements, closes #1264 , closes #1265

### LINKED ISSUE #1264: File manager: Add file copy support
Right now, Markor supports moving files in the file manager. Copy should be supported as well

### LINKED ISSUE #1265: File manager: Ask for overwrite confirmation at copy/move
Moving files can overwrite files with the same name at the destination. Instead we should warn the user and present the following options:

1. Skip
2. Overwrite
3. Keep both (will find non - conflicting name by adding a suffix)
4. Skip all
5. Overwrite all
6. Keep all

4, 5, 6 will be needed in case of move / copy multiple files

harshad1:
This includes
1. Copy Support
2. Support for handling duplicate files (overwrite, skip, skip all etc)

Addresses #1264 and #1265 

ToDo:
- [x] Handle 1 vs more files in dialog
- [x] Folder select dialog should show folder name
- [x] Cancel option
- [x] Lots of testing


harshad1:
I am satisfied with my testing. This is ready for review / merging

gsantner:
I will do some code auto-reformat and try if it works as expected

gsantner:
The problem I see with it is still the same (and one of the reasons why not further expanded yet) - it doesn't work for sdcard and SAF.

gsantner:
Have to recheck somewhen else, it's already late here

harshad1:
@gsantner is there a specific change / fix you're waiting for here?

gsantner (approved):

PR #1282, opened 2021-03-14
todo.txt: Fix trailing space resulting in contexts/projects/due-dates to be entered twice

harshad1:
i.e.
```
foo_  <- insert due date
```
results in
```
foo due:2021-03-13 due:2021-03-13
```
(`_` = space)


harshad1:
This is caused as `\s*$` matches a trailing space 2x for some reason

@gsantner this may be worth a bugfix release

PR #1290, opened 2021-03-22
Editor: Allow top-menu back button also when the file is empty

harshad1:
The back arrow when the file is empty does not work, the UI does not react. This is unexpected and confusing. This change makes the back arrow behave like the back button and the behavior is consistent and as expected.


gsantner:
This change will allow to close the file without making sure it has saved successfully.

I suggest instead to check length and `if trimmged.length == 0`, if this is wanted.

gsantner (approved):

PR #1291, opened 2021-03-23
WIP: Correctly set extension spinner selection

harshad1:
In this PR, the new file dialog is updated so that the correct extension is selected in the spinner


app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -787,14 +787,6 @@ public void setNewFileDialogLastUsedExtension(String v) {
         setString(R.string.pref_key__new_file_dialog_lastused_extension, v);
     }
 
-    public int getNewFileDialogLastUsedType() {
-        return getInt(R.string.pref_key__new_file_dialog_lastused_type, 0);
```
harshad1, line 791: These are redundant as the actual extension is saved

app/src/main/res/layout/new_file_dialog.xml
```diff
@@ -67,7 +67,6 @@
                 android:layout_marginRight="32dp"
                 android:layout_weight="30"
                 android:gravity="end"
-                android:hint=".md"
```
harshad1, line 70: This default looks like an extension is present when it isn't. The UI is clearer when it is not present.

gsantner:
I also found a issue with the recent changes of format selection.  
I.e. I opened markor and main "ToDo" didn't have any format set (no divider lines).  
Manually set it, then swiped to quicknote, saw that nothing is preselected (-> I'm aware of a "undefined" state which is no format - we should get rid of that).

So guess this new, third format selection preference setting doesn't work as expected.  
What do you think of getting rid of all of it alltogether and have just a traditional string for selection in settings - and mirror that to Format dropdown for example?

Screenshot_20210323-145107_Trebuchet

gsantner:
Oh, the changes seems to be in #839 actually, not live yet. So guess that one needs improvements still

app/src/main/res/layout/new_file_dialog.xml
gsantner: maybe .extension?

harshad1: I'll test it. Mostly I'm fed up with accidentally creating `foo` when I wanted `foo.md` :)

harshad1:
I'm happy with the state here now.

app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -771,14 +771,6 @@ public void setPasswordHasBeenSetOnce(boolean b) {
         setBool(R.string.pref_key__default_encryption_password_set_once, b);
     }
 
-    public boolean getNewFileDialogLastUsedEncryption() {
-        return getBool(R.string.pref_key__new_file_dialog_lastused_encryption, false);
-    }
-
-    public void setNewFileDialogLastUsedEncryption(boolean b) {
-        setBool(R.string.pref_key__new_file_dialog_lastused_encryption, b);
```
harshad1, line 779: We use the `.jenc` extension to detect encryption everywhere else. These options are redundant. We can just use the remembered extension.

app/src/main/res/layout/new_file_dialog.xml
gsantner: The main issue is - we cant know what the extension is and what is not, otherwise we would have just single line text input

like

```
myfile.md.txt
a.txt.jenc
my.cool.text.txt
my.learning.with.zim.txt
```

because you might or might not have multiple file extenions

gsantner:
If select "Markdown" from the spinner, and change the extension to ".mydown.txt" just for example, create the new file - it is loaded in "Plaintext" mode.  
Also subsequent new file creations will go for Plaintext, plus show ".mydown.txt" as Format - which is not one of the supported formats.

gsantner:
If select "Markdown" from the spinner, and change the extension to ".mydown.txt" just for example, create the new file - it is loaded in "Plaintext" mode.  
Also subsequent new file creations will go for Plaintext, plus show ".mydown.txt" as Format - which is not one of the supported formats.  

The format spinner selection however should pre-set the "Per-File" format for the new created file, and apply it in same manner for when new files are created.  
Some people for example always want markdown for normal.txt files, or other way round, always plaintext for file.md.

gsantner:
lets wait with this till backup pr is merged. due we change format selection there

Issue #1283, opened 2021-03-14
Add format: Txt2tags (Highlighter, TextActions)


farvardin:
Hello,

I'm working on a [txt2tags](https://txt2tags.org/) support on Markor. At the moment it's rather functional, but I've just replaced Zim (which I don't use), by txt2tags. Btw, txt2tags syntax is 80% like the Zim one. Another similar markup langage is [Creole](https://en.wikipedia.org/wiki/Creole_(markup)), which came to almost the same reasoning as txt2tags did. 

I've put the fork there: https://github.com/farvardin/markor/

I'm using it every day, but I'm not sure if other people would be interested too. In this case, it might worth considering to make a cleaner version, in complement to Zim and not in replacement, but I suppose it would need some extra work to support this.

gsantner:
hi @farvardin 

I'm open for additional format at Markor.  
Though, I do not know txt2tags. If you say "OK this thing is very very similar to Zim", my suggestion is:

* Use class hierachy (`extends ZimClassXyz`)
* OR make smaller adjustments to ZimClasses so it also works for txt2tags by a different boolean switch

Overall, also add the format to the list of available options. The root/main point of the where the available formats is configured...is the `TextFormat` class: <https://github.com/gsantner/markor/blob/master/app/src/main/java/net/gsantner/markor/format/TextFormat.java#L37>

But please don't destroy/make Zim malfunction.  
So, happy to merge when it's ready! Please open a pull request when you have something minimal viable available, so we can take a look at the state and next steps.

Thanks

farvardin:
Hello,

I've looked how to integrate the txt2tags format in addition, instead of replacing zimwiki (which was just a quick hack to check if it could work as expected).

It seems to work well. 

> But please don't destroy/make Zim malfunction.

Sure! I wasn't my intention of course. Every markup format can coexist with the others ;)

I'll work more on it for a better integration, and I'll propose a pull request when I'll be satisfied!

farvardin:
@gsantner I've made the PR!

It compiles fine for me.

PR #1300, opened 2021-04-01
txt2tags support

farvardin:
hello, here is the support to the txt2tags markup! As announced on https://github.com/gsantner/markor/issues/1283

Thank you.


app/src/main/java/net/gsantner/markor/activity/SettingsActivity.java
```diff
@@ -376,9 +376,10 @@ public void onFsViewerConfig(FilesystemViewerData.Options dopt) {
...
                 case R.string.pref_key__zimwiki__reorder_actions:
+                case R.string.pref_key__txt2tags__reorder_actions:
                 case R.string.pref_key__todotxt__reorder_actions: {
                     Intent intent = new Intent(getActivity(), ActionOrderActivity.class);
-                    intent.putExtra(ActionOrderActivity.EXTRA_FORMAT_KEY, (keyResId == R.string.pref_key__markdown__reorder_actions) ? R.id.action_format_markdown : (keyResId == R.string.pref_key__todotxt__reorder_actions ? R.id.action_format_todotxt : (keyResId == R.string.pref_key__zimwiki__reorder_actions ? R.id.action_format_zimwiki : R.id.action_format_plaintext)));
+                    intent.putExtra(ActionOrderActivity.EXTRA_FORMAT_KEY, (keyResId == R.string.pref_key__markdown__reorder_actions) ? R.id.action_format_markdown : (keyResId == R.string.pref_key__todotxt__reorder_actions ? R.id.action_format_todotxt : (keyResId == R.string.pref_key__zimwiki__reorder_actions ? R.id.action_format_zimwiki : (keyResId == R.string.pref_key__txt2tags__reorder_actions ? R.id.action_format_txt2tags : R.id.action_format_plaintext))));
```
harshad1, line 382: Can we replace this with a switch? A line this long is pretty unreadable

farvardin: yes, sure, I've just adapted the code to what was already there. I must admit I don't know much java, so I've looked how it was before.

gsantner: I know it's not pretty but keep it in the PR farvardin.

If it's just that line, preferrably we/I take care outside of the PR, after merge.

harshad1:
Making a note here so I remember:

If https://github.com/gsantner/markor/pull/1244 is merged before this PR, we will have to add the new format to the list of indices

app/src/main/java/net/gsantner/markor/activity/SettingsActivity.java
farvardin: > I know it's not pretty but keep it in the PR farvardin.

ok, I won't touch it.

app/src/main/java/net/gsantner/markor/format/txt2tags/Txt2tagsHighlighter.java
```diff
@@ -0,0 +1,201 @@
...
+import net.gsantner.markor.format.ListHandler;
+import net.gsantner.markor.model.Document;
+import net.gsantner.markor.ui.hleditor.Highlighter;
+import net.gsantner.markor.ui.hleditor.HighlightingEditor;
+
+import java.util.regex.Pattern;
+
+import other.writeily.format.txt2tags.WrTxt2tagsHeaderSpanCreator;
+
+public class Txt2tagsHighlighter extends Highlighter {
```
gsantner, line 26: Suggestion: As you said it's mostly the same: extend from ZimHighlighter instead and use inheritance/overrides where required. Guess for example Patterns are 99% same, and we better maintain these only at **one place**.

farvardin: I'm afraid I don't know how to do this inheritance :( I'm not a (java) developper.

It's easier for me to use a diff tool to compare the two in the case there is a modification.

gsantner: extends = Inheritance. extend from zim highliter instead of the generic one

gsantner:
@farvardin   
Can you please list differencies between zim and txt2tags? Want to understand how big the difference is. And if it's really just some minor things, it not really makes sense to re-develop everything again (everything requires maintainence), instead of adding minor changes to zim.

farvardin:
> @farvardin
> Can you please list differencies between zim and txt2tags? Want to understand how big the difference is. And if it's really just some minor things, it not really makes sense to re-develop everything again (everything requires maintainence), instead of adding minor changes to zim.

- the heading are reversed: in txt2tags, heading level 1 is one =, with zim it's ====== (you only have to count the number of surrounding = in txt2tags to know the level)
- links are [description http://url.com]. In zim I can't see external links like that, they only recognise URL and other links are internal wiki links
- lists are * in zim, - in txt2tags
- numbered lists are 1. 2. 3. in zim, + in txt2tags
- no checkboxes in txt2Tags
- bold italic and underline are similar
- strike through is ~~ in zim, and -- in txt2tags
- images are [image.jpg] in txt2tags and {{image.jpg}} in zim.
- commented line (won't be displayed) is % at the beginning of a line in txt2tags. Doesn't exists in zim.


It's not that much different, but I suppose it would be quite complicated to just make txt2tags a class from zim (I can't code this myself). And the copied and adapted code from zim is just working now, so no need to further development. They are different markups, it would be rather strange to not treat them as two separate things (and txt2tags is older, it's even older than markdown).

What do you think?

PR #1301, opened 2021-04-01
Disable fallback spacing

harshad1:
Fallback spacing causes the text height to change annoyingly when I am editing text. This is especially apparent in todo.txt with the extra ascent.


harshad1:
@gsantner Are you able to reproduce the build error here? Seems to build fine locally for me.

gsantner (approved):

PR #1305, opened 2021-04-03
todo.txt: Settings option for always-visible @contexts, closes #1302

### LINKED ISSUE #1302: todo.txt: Settings option for always-visible @contexts
Markor will list all the projects and contexts available in the `add projects` and `add contexts` dialogs respectively.

However, if I delete (or mark as done and archive) the last task with a context or project, it is no longer in the list and I have to type it again.

We should have a setting where a user can enter a (space separated) list of projects and contexts.

I.e for example

```
@home @pc @work +work +garage 
```
These will then always be available in the context

harshad1:
This PR addresses #1302

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
+++ b/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
@@ -32,6 +32,7 @@ import net.gsantner.opoc.util.FileUtils;
 import net.gsantner.opoc.util.StringUtils;
 
 import java.io.File;
+import java.lang.reflect.Array;
 import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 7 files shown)


app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -37,6 +37,8 @@
 import java.util.Arrays;
 import java.util.Calendar;
 import java.util.List;
+import java.util.SortedSet;
+import java.util.TreeSet;
```
gsantner, line 42: we have uniq(list-string).sorted available, so sinple list append with duplicates would be totally fine

harshad1: Good point. I have cleaned up that code and used it

harshad1:
@gsantner I am done with my changes

gsantner (approved):

PR #1306, opened 2021-04-03
Decrease scrollbar width for better usability

harshad1:
1.5 is a better value than 2.0 in my testing. At 2.0, text selection near the edge was more affected.


harshad1:
@gsantner title should be _decrease_ scrollbar width :)

PR #1308, opened 2021-04-07
Additional contexts & projects - cleanup of PR #1305 changes, closes #1302

### LINKED ISSUE #1302: todo.txt: Settings option for always-visible @contexts

harshad1:
@gsantner Couple of simplifications to the code

### DIFF
--- a/app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
+++ b/app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
@@ -37,6 +37,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.TreeSet;
 
 import static net.gsantner.markor.format.todotxt.TodoTxtTask.SttTaskSimpleComparator.BY_CONTEXT;
 import static net.gsantner.markor.format.todotxt.TodoTxtTask.SttTaskSimpleComparator.BY_CREATION_DATE;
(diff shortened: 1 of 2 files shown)


app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
```diff
@@ -262,7 +263,7 @@ public static void showSttSortDialogue(Activity activity, final Callback.a2<Stri
     public static void showSttContextDialog(Activity activity, List<String> availableData, Callback.a1<String> callback) {
         SearchOrCustomTextDialog.DialogOptions dopt = new SearchOrCustomTextDialog.DialogOptions();
         baseConf(activity, dopt);
-        dopt.data = sortUniqNonEmpty(availableData);
+        dopt.data = new ArrayList<>(new TreeSet<>(availableData));
```
harshad1, line 266: Replace function with one-liner

app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -813,7 +813,7 @@ public boolean getSetWebViewFulldrawing(boolean... setValue) {
     }
 
     public String getTodotxtAdditionalContextsAndProjects() {
-        return getString(R.string.pref_key__todotxt__additional_projects_contexts, "+music +video @home @shop").replaceAll("\\s+", " ").trim();
+        return getString(R.string.pref_key__todotxt__additional_projects_contexts, "+music +video @home @shop");
```
harshad1, line 816: We don't require the replace and trim code. The additional string is parsed by the same logic which parses tasks so multiple and trailing spaces are ignored.

gsantner: edittext preference is multiline (no android: option). Task class is for single line of text?

harshad1: The task parser can handle multiple lines (i.e. text with `\n`). 

image

Works fine

gsantner (approved):

PR #1332, opened 2021-05-06
Per-file font size, closes #1326

### LINKED ISSUE #1326: Per-file font size
#### General information

* **App version:** 
v2.6.0(122)

* **System:** 

Android 10 (galaxy tab A7 SM-T500)

#### Description

This is probably a feature request.

It appears that Markor allows you to change the Font size at under Settings which is three hops away.  Most readers and editors that i've used allow you to change this right from the document window you are editing / viewing because it's such a common operation.

Sure, setting a global default for font size and font once in settings will cover a lot of cases.  But for whatever reason, certain documents and document types just look better and are easier to work with using a larger font size and being able to switch this back and forth as you change between different documents on the edit interface itself might be useful to some.

Just a suggestion.  Thoughts?

harshad1:
Addresses #1326

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
@@ -532,12 +532,11 @@ public class DocumentEditFragment extends GsFragmentBase implements TextFormat.T
                 return true;
             }
             case R.id.action_set_font_size: {
-                final DisplayMetrics displayMetrics = getContext().getResources().getDisplayMetrics();
-                final int oldSize = (int) (_hlEditor.getTextSize() / displayMetrics.scaledDensity);
-                SearchOrCustomTextDialogCreator.showFontSizeDialog(getActivity(), oldSize, (v) -> {
+                final String path = getPath();
+                SearchOrCustomTextDialogCreator.showFontSizeDialog(getActivity(), _appSettings.getDocumentFontSize(path), (v) -> {
                     final int newSize = Integer.parseInt(v);
                     _hlEditor.setTextSize(TypedValue.COMPLEX_UNIT_SP, (float) newSize);
-                    _appSettings.setDocumentFontSize(getPath(), newSize);
+                    _appSettings.setDocumentFontSize(path, newSize);
                 });
             }
         }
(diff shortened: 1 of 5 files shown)


app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -454,6 +456,7 @@ public boolean onOptionsItemSelected(final MenuItem item) {
                 if (_document != null) {
                     _document.setFormat(itemId);
                     applyTextFormat(itemId);
+                    _appSettings.setDocumentFormat(getPath(), _document.getFormat());
```
harshad1, line 459: Moved all the setting of per-document options into the action code. This way we don't set per-document values unless the user actually sets them

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -526,6 +531,17 @@ public void onFsViewerConfig(FilesystemViewerData.Options dopt) {
...
             }
+            case R.id.action_set_font_size: {
+                final DisplayMetrics displayMetrics = getContext().getResources().getDisplayMetrics();
+                final int oldSize = (int) (_hlEditor.getTextSize() / displayMetrics.scaledDensity);
+                SearchOrCustomTextDialogCreator.showFontSizeDialog(getActivity(), oldSize, (v) -> {
+                    final int newSize = Integer.parseInt(v);
+                    if (newSize != oldSize) {
+                        _hlEditor.setTextSize(TypedValue.COMPLEX_UNIT_SP, (float) newSize);
+                    }
+                    _appSettings.setDocumentFontSize(getPath(), newSize);
```
harshad1, line 542: Setting saved only if user explicitly sets font size.

gsantner: I don't think we need the displayMetrics & oldSize. Just 

```
callback {
  editor.setTextSize  
  setDocumentFontsize
}
```

should be fine? And in case user presses cancel (thus opposite of "explicitley set"), nothing happens as the callback not invoked.

whats your view/thought?

harshad1: I have removed the unnecessary test. The displayMetrics and oldSize are used to highlight the current font size. I think that is useful to have

harshad1: Ha! Didn't think of that. Yes, that will work.

harshad1: I have pushed this change

PR #1337, opened 2021-05-07
 Recursive file search with in-content search, closes #1295, by @adelobosko #1337 

### LINKED ISSUE #1295: Search in files with Queue
I am not strong on android / java development.
Could we search by context, simple steps:

1) Create a queue.
      var queue = new System.Collections.Generic.Queue<DirectoryInfo>();

2) Get current directory:
      queue.Enqueue(rootDirectory);

3) while(queue.Count > 0) =>
var currentDir = queue.Dequeue();
Get all files and subdirs of the current directory

4) Find directories and files by name
    (add to the search result: paths to dir/file, count of results (for configuring search and taking by chunks if apparat is low (for ex. get first 100, then skip 100 and take next 100, skip 200 take 100...))

    4.1) Search context in all files in the current directory 
    (create something like StreamReader)

    4.2) read file line by line
    (or by chunk if you configure, but do not forget about offset that depends on search text length (for ex. chunk = 100, search text length = 10 symbols, get(0,100), get(80-180), get(160-260,get(240-349))

    4.3) search all includes of search text into the current line / chunk

    4.4) add the current line and character index to the search result 
    (group by file path)
    (and if configure.IsShowPreview == true then move intto searchResult configure.PreviewSymbolsCount behind search index:
    ex. line "abcd 12345 6 123 TEST 123545 gsads", searchText = "test", if PreviewSymbolsCount == 10 => resultPreview = line.Skip(index-PreviewSymbolsCount ).Take(searchText.Length+PreviewSymbolsCount 2)
    => resultPreview = line.Skip(16-10).Take(4+102) => line.Skip(6).Take(24) => "2345 6 123 TEST 123545 "

5) Get all directories of current directory
    foreach (var directoryInfo in currentDir.GetDirectories())
    {
        queue.Enqueue(directoryInfo);
    }

6) got to next iteration (Step 3)
 
7)  mainTask.Wait()

8) show searchResult...

it does not take too much memory and CPU, as you can configure a search:
- read file by lines or chunks;
- (get all) get every N search result (then click the "get next" button and get next N);
- (not)use infile search;
- (not)show preview;

Could somebody do this, as I think people who know Java will do it faster and better than I...
if not I will try it do myself then.

adelobosko:
https://github.com/gsantner/markor/issues/1295


adelobosko:
https://user-images.githubusercontent.com/56083544/118863265-5b152280-b8e7-11eb-989a-023e7ddfd408.mp4

adelobosko:
https://user-images.githubusercontent.com/56083544/118864668-e0e59d80-b8e8-11eb-9850-4484bc86ad20.mp4

gsantner:
I did various improvements all over the plcae. Also ~ halfed the amount of changed files, which makes it _less hard_ to review.

For that, already transitioned some generic code parts to master branch.

gsantner:
I did various improvements all over the placed. Also ~ halved the amount of changed files, which makes it _less hard_ to review.

For that, already transitioned some generic code parts to master branch.

gsantner:
@harshad1  
If you are interested, please feel free to help review.  
It's just a quite massive PR on many areas. It's already open quite some time, and shelfed off many things already - but still quite big.

app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
```diff
@@ -66,13 +67,17 @@
 
     private static boolean nextLaunchTransparentBg = false;
 
-    public static void launch(Activity activity, File path, Boolean isFolder, Boolean doPreview, Intent intent) {
+
+    public static void launch(Activity activity, File path, Boolean isFolder, Boolean doPreview, Intent intent, final int... arg0lineNumber) {
```
harshad1, line 71: It seems the variadic `final int... arg0lineNumber` is not used for multiple values, but as a nullable integer. If this is the case, I suggest replacing it with `Integer lineNumber`, which can also be set to null

app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
```diff
@@ -271,14 +278,10 @@ public void setDocumentTitle(final String title) {
         }
     }
 
-    public void showTextEditor(@Nullable Document document, @Nullable File file, boolean fileIsFolder) {
-        showTextEditor(document, file, fileIsFolder, false);
-    }
-
-    public void showTextEditor(@Nullable Document document, @Nullable File file, boolean fileIsFolder, boolean preview) {
-
+    public void showTextEditor(@Nullable Document document, @Nullable File file, boolean fileIsFolder, boolean preview, final int... arg0fileLineNumber) {
```
harshad1, line 281: Same observation about using an overload. Or just use `Integer lineNumber` and pass in null wherever needed

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -491,7 +482,7 @@ public boolean onOptionsItemSelected(final MenuItem item) {
             case R.id.action_load_epub: {
                 FilesystemViewerCreator.showFileDialog(new FilesystemViewerData.SelectionListenerAdapter() {
                                                            @Override
-                                                           public void onFsViewerSelected(String request, File file) {
+                                                           public void onFsViewerSelected(String request, File file, final Integer... lineNumber) {
```
harshad1, line 485: variadic as optional

app/src/main/java/net/gsantner/markor/activity/SettingsActivity.java
```diff
@@ -264,7 +264,7 @@ public Boolean onPreferenceClicked(Preference preference, String key, int keyRes
                         FragmentManager fragManager = getActivity().getSupportFragmentManager();
                         FilesystemViewerCreator.showFolderDialog(new FilesystemViewerData.SelectionListenerAdapter() {
                             @Override
-                            public void onFsViewerSelected(String request, File file) {
+                            public void onFsViewerSelected(String request, File file, final Integer... lineNumber) {
```
harshad1, line 267: Variadic as optional

app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -592,6 +609,82 @@ public boolean isMultiWindowEnabled() {
         }
     }
 
+    public boolean isShowSearchResultOnCancel() {
+        return getBool(R.string.pref_key__is_show_search_result_on_cancel_enabled, true);
```
harshad1, line 613: Note: We can use backup-restore to change these values (if necessary for testing etc)

harshad1:
Still reviewing, will take a couple days for me given my time constraints, but the code looks pretty good so far. Only issue I have seen is this pattern of using a variadic (Varargs) as an optional arg. This isn't a great approach, IMO, but feel free to convince me otherwise.

gsantner:
Yeah the variadic actually are from me by PR changes. This way its optional to specify for callback

adelobosko:
>  I prefer only one single streamedline.
i have changed Integer params to one Integer parameter, I think it will be better.

app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
harshad1: @adelobosko missed this one? :)

harshad1:
> Yeah the variadic actually are from me by PR changes. This way its optional to specify for callback.
> 
> There were previously 2 callback methods for selected and I prefer only one single streamedline.

Then the Integer method is the way to go, IMO. Using variadic is really confusing unless one studies the code underneath. Personally, I am much happier with the changes @adelobosko just made

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -680,18 +666,54 @@ public boolean onBackPressed() {
...
+            if (_isPreviewVisible) {
+                _appSettings.setLastViewPosition(_document.getFile(), _webView.getScrollX(), _webView.getScrollY());
+            } else {
+                _appSettings.setLastEditPosition(_document.getFile(), _hlEditor.getSelectionStart());
+            }
+        }
+    }
+
+    public void restoreDocumentPositions() {
+        if (!Arrays.asList(_hlEditor, _webView, _document.getFile(), _primaryScrollView).contains(null) && !isTodoOrQuickNote()) {
```
harshad1, line 690: Neat trick!

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -377,14 +360,17 @@ public boolean onOptionsItemSelected(final MenuItem item) {
                 return true;
             }
             case R.id.action_preview: {
+                saveDocumentPositions();
                 setDocumentViewVisibility(true);
```
harshad1, line 364: Does it make more sense to call `saveDocumentPositions` inside `setDocumentViewVisibility`

app/src/main/java/net/gsantner/opoc/ui/FileSearchDialog.java
```diff
@@ -0,0 +1,224 @@
...
+        // TextView
+        messageTextView.setText(R.string.recursive_search_in_current_directory);
+        dialogLayout.addView(messageTextView, margins);
+
+        // EdiText: Search query input
+        searchEditText.setHint(R.string.search);
+        searchEditText.setSingleLine(true);
+        searchEditText.setMaxLines(1);
+        searchEditText.setTextColor(textColor);
+        searchEditText.setHintTextColor((textColor & 0x00FFFFFF) | 0x99000000);
```
harshad1, line 101: Should make the color a resource

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
```diff
@@ -120,26 +123,33 @@ public static void showInsertTableRowDialog(final Activity activity, final boole
...
+            final Integer maxSearchDepth = appSettings.getSearchMaxDepth();
+            final List<String> ignoredDirs = appSettings.getIgnoredSearchDirNames();
+            final List<String> ignoredFiles = appSettings.getIgnoredSearchFileNames();
+            final boolean isShowMatchPreview = appSettings.isShowMatchPreview();
+            final List<String> searchByContentExtensions = appSettings.getSearchByContentExtensions();
+            appSettings.setSearchQueryRegexUsing(cb_options._isRegexQuery);
+            appSettings.setSearchQueryCaseSensitivity(cb_options._isCaseSensitiveQuery);
+            appSettings.setSearchInContent(cb_options._isSearchInContent);
+            appSettings.setOnlyFirstContentMatch(cb_options._isOnlyFirstContentMatch);
+            SearchEngine.Config config = new SearchEngine.Config(searchDir, cb_options._query, isShowResultOnCancel, maxSearchDepth, ignoredDirs, ignoredFiles, cb_options._isRegexQuery, cb_options._isCaseSensitiveQuery, cb_options._isSearchInContent, cb_options._isOnlyFirstContentMatch, isShowMatchPreview, searchByContentExtensions);
```
harshad1, line 145: Would it be possible to use a dopt like mechanism here? These long function calls are hard to follow

app/src/main/java/net/gsantner/opoc/ui/FileSearchDialog.java
gsantner: It is a color filter applied to a existing color. should be fine I'd say

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
gsantner: > may not work if all are null.

why? [null, null, null].contains(null) => true

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
```diff
@@ -120,26 +123,33 @@ public static void showInsertTableRowDialog(final Activity activity, final boole
...
+
+        Callback.a1<FileSearchDialog.CallBackOptions> fileSearchDialogCallback = (cb_options) -> {
+
+            final boolean isShowResultOnCancel = appSettings.isShowSearchResultOnCancel();
+            final Integer maxSearchDepth = appSettings.getSearchMaxDepth();
+            final List<String> ignoredDirs = appSettings.getIgnoredSearchDirNames();
+            final List<String> ignoredFiles = appSettings.getIgnoredSearchFileNames();
+            final boolean isShowMatchPreview = appSettings.isShowMatchPreview();
+            final List<String> searchByContentExtensions = appSettings.getSearchByContentExtensions();
+            appSettings.setSearchQueryRegexUsing(cb_options._isRegexQuery);
```
harshad1, line 141: Do we really want to remember these options for next time? @gsantner

app/src/main/java/net/gsantner/opoc/ui/FileSearchDialog.java
harshad1: 👍

app/src/main/java/net/gsantner/opoc/ui/SearchEngine.java
```diff
@@ -0,0 +1,570 @@
...
+                }
+            } catch (Exception ignored) {
+            }
+
+            return subQueue;
+        }
+
+
+        @Override
+        protected void onProgressUpdate(Integer... values) {
```
harshad1, line 377: Looks like we always expect exactly 4 args. Should just name them.

harshad1:
Right now, this searches through PDF, JPG and other binary files. We should add logic to test if a file is a text file before searching it.

app/src/main/java/net/gsantner/opoc/ui/SearchEngine.java
gsantner: this is a method from Android AsyncTask class. Signature should match

gsantner:
This is a common issue where we dont have a generic solution yet. But please not lets implement this in the PR.

Instead, lets have a simple if check to not read files bigger than 800kb. For example.

app/src/main/java/net/gsantner/opoc/ui/SearchEngine.java
harshad1: Ah sorry, nevermind

harshad1:
> This is a common issue where we dont have a generic solution yet. But please not lets implement this in the PR.

We have `Files.probeContentType` and `URLConnection.guessContentTypeFromStream` built into java. Additionally we already have the `TextFormat.isTextFile` routine. 

I can add a function to do this test as a separate PR. We can then use it here. Should have it up in a few hours
    
> Instead, lets have a simple if check to not read files bigger than 800kb. For example.

I don't think this is a very good heuristic. Most of the PDFs and images I am getting in my search are < 800 kb. They really pollute the search results and increase the search time.

app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
gsantner: changed

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
gsantner: might be that setting cursor happens after saving (=->always 0), caution

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
gsantner: I think we should carefully think what all settings we want to have. however putting PR review focus atm on the funcationalties and not breaking existing code

gsantner:
> We have Files.probeContentType and URLConnection.guessContentTypeFromStream built into java. Additionally we already have the TextFormat.isTextFile routine.

yeah we can try. however mentioned java methods often do not work as expected, or are not supported on older android(=java) versions.

> I can add a function to do this test as a separate PR. We can then use it here. Should have it up in a few hours

ehm I think @adelobosko  already gave you push rights to his repository. can you try this way, push directly to his PR?

harshad1:
> ehm I think @adelobosko already gave you push rights to his repository. can you try this way, push directly to his PR branch?

Looks like I do not have push rights. Code is ready though. @adelobosko could you give me permission?

app/src/main/res/xml/preferences_master.xml
```diff
@@ -71,6 +71,50 @@
...
+                android:title="@string/ignored_search_dir_names" />
+            <android.support.v7.preference.EditTextPreference
+                android:defaultValue="@string/empty_string"
+                android:dialogMessage="@string/ignored_search_file_names_dialog"
+                android:icon="@drawable/ic_baseline_rule_folder_24"
+                android:key="@string/pref_key__ignored_search_file_names"
+                android:title="@string/ignored_search_file_names" />
+            <android.support.v7.preference.EditTextPreference
+                android:defaultValue="@string/empty_string"
+                android:dialogMessage="@string/search_by_context_have_to_include_dialog"
```
harshad1, line 105: Instead of having this and the previous setting, I think it may be better to just have a filename filter directly in the dialog. 

Default filter file would be `*.*`, and a user would be able to enter, say, `*.md`.

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
harshad1: :+1

app/src/main/java/net/gsantner/opoc/ui/SearchEngine.java
```diff
@@ -0,0 +1,570 @@
...
+            super.onProgressUpdate(values);
+
+            Integer queueLength = values[0];
+            Integer queueDepth = values[1];
+            Integer filesFound = values[2];
+            Integer countCheckedFiles = values[3];
+            String snackBarText = "f:" + filesFound + " qu:" + queueLength + "|" + queueDepth + " c:" + countCheckedFiles + "\n" + _config._query;
```
harshad1, line 384: Not very important for this PR, but we should consider replacing the letters with icons in the future.

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
adelobosko: > 
> 
> might be that setting cursor happens after saving (=->always 0), caution

yes, it is a reason that I do not move `saveDocumentPositions` into `setDocumentVisibility`

app/src/main/res/xml/preferences_master.xml
adelobosko: Yes, it is more convenient

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
adelobosko: > 
> Looks like I do not have push rights. Code is ready though. @adelobosko could you give me permission?

Sure, I sent you the invite

harshad1:
@adelobosko Could you give me push permissions for your branch? I have more improvements I would like to make to filtering and it would be easier to push rather than to create PRs

harshad1:
> see opoc utils, there is a similar but better working version already...with text/markdown for example

Sorry, what is the context for this?

gsantner:
grafik

app/src/main/java/net/gsantner/opoc/ui/FileSearchResultSelectorDialog.java
```diff
@@ -0,0 +1,247 @@
...
+
+        if (searchResults.isEmpty()) {
+            dialogBuilder.setMessage("     ¯\\_(ツ)_/¯     ");
+        }
+
+        return dialogBuilder;
+    }
+
+
+    private static Pair<ArrayList<Map<String, Object>>, ArrayList<ArrayList<Map<String, Object>>>> filter(final List<SearchEngine.FitFile> searchResults, String query) {
```
harshad1, line 179: @adelobosko If you are able to document this data structure, that would be helpful for future maintainece.

adelobosko:
> @adelobosko Could you give me push permissions for your branch? I have more improvements / bug fixes I would like to make to filtering and it would be easier to push rather than to create PRs

Hmm, i sent in ite to you yesterday, couldnyou check?

harshad1:
> Hmm, i sent in ite to you yesterday, couldnyou check?

image

I believe you'll have to add me as a collaborator:

https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/managing-access-to-your-personal-repositories/inviting-collaborators-to-a-personal-repository

adelobosko:
> image
> 
> I believe you'll have to add me as a collaborator:

I have done it. 
maybe this link will help you:
https://github.com/adelobosko/markor/invitations

image

harshad1:
@adelobosko Thanks! got it.

harshad1:
IMO one more thing we do need to add is shortening very long lines in the search results.

Perhaps something like  `. . . < first match in line > . . . `

harshad1:
> IMO one more thing we do need to add is shortening very long lines in the search results.

I have added support for this.

@adelobosko  @gsantner IMO this is ready to be merged.

There will always be things which can be tweaked, but I think they should be discussed and made in subsequent PRs, if needed.

harshad1:
Actually, one more thing: I am noticing that the keyboard no longer opens when I open a file. Was this change made intentionally?

adelobosko:
> Was this change made intentionally?
Yes, need to tweak it. Let me do this

harshad1:
> Yes, need to tweak it. Let me do this

👍

harshad1:
@adelobosko are your changes ready for me to test?

adelobosko:
> @adelobosko are your changes ready for me to test?

@harshad1 yes, sorry that did not inform you

harshad1:
@adelobosko Not working consistently. I am working on ways to consistently show the keyboard after the animation (using AnimatorListener), but none seem to work

It may be best to just use setSelection for now and come back to this in a later PR. Will push my changes tomorrow, if they are successful

adelobosko:
I think just need to increase post delay in the showSoftKeyboard()

adelobosko:
@harshad1 could you try it again?

gsantner:
> animation (using AnimatorListener), but none seem to work

sure that the keyboard call happens on UI thread?

harshad1:
> @harshad1 could you try it again?

Keyboard now opens, but the following issues remain:

1. Keyboard remains open in file explorer view. To reproduce
    1. Open document
    1. press back arrow at the top of the document
2. Quicknote and todo.txt have no selection set (and keyboard does not open). They are supposed default jump to the end of the file.
3. If I switch to another app and then back to the document, the selection and keyboard are not restored

I strongly feel we should restore all the logic in `DocumentEditFragment`, except for the minimum needed to support launching search intents, and address these changes it in a separate PR. There are a lot of nuances here related to activity lifecycle.

DarkP0wer:
> Keyboard remains open in file explorer view. To reproduce 
Maybe remove the smooth selection / delays when open, and just select a position?

>     2. Quicknote and todo.txt have no selection set (and keyboard does not open). They are supposed default jump to the end of the file.

This is not supported in the current PlayMarket's version

>     3. If I switch to another app and then back to the document, the selection and keyboard are not restored

and this as well, so it is not a bug from this PR and we can add it in the future with the next pull.

adelobosko:
> Keyboard remains open in file explorer view. To reproduce 
Maybe remove the smooth selection / delays when open, and just select a position?

>     2. Quicknote and todo.txt have no selection set (and keyboard does not open). They are supposed default jump to the end of the file.

This is not supported in the current PlayMarket's version

>     3. If I switch to another app and then back to the document, the selection and keyboard are not restored

and this as well, so it is not a bug from this PR and we can add it in the future with the next pull.

harshad1:
> This is not supported in the current PlayMarket's version, so it is not a bug from this PR and we can add it in the future with the next pull.

Hmmm. I am using markor from F-Droid and it works fine is the play store version that different. To be clear, I mean Quicknote and todo.txt as standalone files, not in the tabs. (Create by long pressing the markor icon -> dragging todo.txt).

On my device, both these files jump to the bottom.

gsantner:
play store may lag behind in releases, but generally speaking it's 99.99% same code.

adelobosko:
> To be clear, I mean Quicknote and todo.txt as standalone files, not in the tabs.

Hmm yes, I will fix it

gsantner:
yeah please lets tidy the PR as far the feature is concered. but not lets add general/unrelated fixes to it

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
```diff
@@ -409,9 +409,9 @@ public void onFsViewerDoUiUpdate(FilesystemViewerAdapter adapter) {
                 }
 
                 @Override
-                public void onFsViewerSelected(String request, File file) {
+                public void onFsViewerSelected(String request, File file, final Integer lineNumber) {
                     if (TextFormat.isTextFile(file)) {
-                        DocumentActivity.launch(MainActivity.this, file, false, null, null);
+                        DocumentActivity.launch(MainActivity.this, file, false, null, null, lineNumber);
```
harshad1, line 414: @adelobosko  Is `lineNumber` actually used / passed in anywhere in `onFsViewerSelected`? If not, we should revert the signature and pass in null here?

adelobosko:
> yeah please lets tidy the PR as far the feature is concered. but not lets add general/unrelated fixes to it

ok, I can sort it tomorrow or at next day after tomorrow

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
adelobosko: > 
> 
> @adelobosko Is `lineNumber` actually used / passed in anywhere in `onFsViewerSelected`? If not, we should revert the signature and pass in null here?

Sorry, I do not understand the question.

adelobosko: > 
> 
> @adelobosko Is `lineNumber` actually used / passed in anywhere in `onFsViewerSelected`? If not, we should revert the signature and pass in null here?

if we replace lineNumber here to null then we could not move to this line in the file
image

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -680,18 +663,75 @@ public boolean onBackPressed() {
...
+                    hideSoftKeyboard();
+                } else {
+                    if ((v = _appSettings.isEditorStartOnBottom() || isDocTodoOrQuickNote() ? _hlEditor.length() : _appSettings.getLastEditPositionChar(_document.getFile())) >= 0) {
+                        final int selPosition = v;
+                        _hlEditor.postDelayed(() -> {
+                            if (!_hlEditor.hasFocus()) {
+                                _hlEditor.requestFocus();
+                            }
+                            _hlEditor.setSelection(selPosition);
+                            new ActivityUtils(getActivity()).showSoftKeyboard().freeContextRef();
```
harshad1, line 703: So far, i have not found any issues in my testing this latest version. Only some cleanup needed if I can't find more behavior bugs. :+1:

harshad1: @adelobosko I do not think that this postDelayed is actually necessary. `restoreDocumentPositions` is called in `onResume`. UI actions should be fine there without any delay. Works fine on my device without `postDelayed`

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -680,18 +663,75 @@ public boolean onBackPressed() {
...
+        final boolean isTodoFile = _document.getFile().getPath().equals(_appSettings.getTodoFile().getPath());
+        final boolean isQuickNoteFile = _document.getFile().getPath().equals(_appSettings.getQuickNoteFile().getPath());
+        return isTodoFile || isQuickNoteFile;
+    }
+
+    private void hideSoftKeyboard() {
+        new ActivityUtils(getActivity()).hideSoftKeyboard().freeContextRef();
+        _hlEditor.postDelayed(() -> {
+            _hlEditor.clearFocus();
+            new ActivityUtils(getActivity()).hideSoftKeyboard().freeContextRef();
```
harshad1, line 731: @adelobosko Same comment about postDelayed may not be necessary

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
adelobosko: > 
> 
> @adelobosko I do not think that this postDelayed is actually necessary. `restoreDocumentPositions` is called in `onResume`. UI actions should be fine there without any delay. Works fine on my device without `postDelayed`

@harshad1 hmm, for me delay is needed. I tried without the delay but the selection could not set. Try to open a big file. I think need another way to set selection. After the editor text is loading... i going to remove restore selection in this pull

adelobosko: > After the editor text is loading...

But seems 300 milliseconds is enough for loading a big file (selection is set). Need to try a very big file :)

adelobosko (commented):
i think it is ready

gsantner:
@adelobosko 

Why did you push `reverted the "start editor on bottom" option`? Intentionally did the change, as it works slightly different now.

adelobosko:
> @adelobosko 
> 
> Why did you push `reverted the "start editor on bottom" option`? Intentionally did the change, as it works slightly different now, and all clients have it enabled currently as it's the old default.

@gsantner 
Because I thought that is related to the "restore position" commit that was reverted before...

gsantner:
@adelobosko 

Just one "not" filename filter should be fine. I tend to suggestion of combining into one. This is also what I'm used to for example at Syncthing <https://docs.syncthing.net/users/ignoring.html#example>. Having two options is overkill.

screenshot-2021-05-31__21-03-15

Do you agree or is there a bigger issue at implementation with that? I mean `.*\.mp4` can be applied folderpaths `/storage/emulated/0/video.mp4` just in the same manner.

adelobosko:
@gsantner so do I need to discard my revert :) (`start on the bottom`)?

gsantner:
Well I wanted to know why you reverted back to the original key :D

adelobosko:
> I mean `.*\.mp4`
> 

Yes, it will be fine

gsantner:
I'm sorry but I'm not happy with multiple signatures and callback methods, carrying usually the same result ("finished search").

Trying to make efforts for instead: Callback.a1<String, Hashmap>, with the second one containing key value pairs of additional metadata.


grafik


grafik

adelobosko:
Not clear to me, do i need back the revert of `start on the bottom commit`?

I reverted `the restore position commit` because i want to create a new pull with this.
And I reverted `the start on bottom commit`, because I thought that is related to `restore the position commit`

gsantner:
I had the finding that, by at least changes of the PR - start on bottom and recent position restore didn't work as expected. But with the default configuration that was in the PR it worked better for me

gsantner:
@adelobosko 

unused method, can be removed right?

grafik

adelobosko:
@harshad1 getIndexFromLineOffset working bad... please move my function back...

adelobosko:
@harshad1 for example:
String t1 = "2123";
int index = getIndexFromLineOffset(t1, 0, 0);
and sometimes it moves to the end of the previous line but not to start...

adelobosko:
> 
> 
> @harshad1 for example:
> String t1 = "2123";
> int index = getIndexFromLineOffset(t1, 0, 0); //index = 4 not 0
> and sometimes it moves to the end of the previous line but not to start...

adelobosko:
@harshad1 and try to check big files what will be faster...
```java
int index = getIndexFromLineOffset(t1, 100, 0);
```
or
```java
int index = getIndexByLineNumber(t1, 100);
```
getIndexByLineNumber will be faster 4 times as does not have additional calculations, just counts the `\n` char...

adelobosko:
> unused method, can be removed right?
@gsantner sure, (it used for a history listview, now it is a spinner)

gsantner:
* guess it's ok to get `getIndexByLineNumber` back
* about "ignored dir/file" - should I look into it or have you already started adelobosko?

harshad1:
@adelobosko @gsantner I am investigating the performance and behavior difference, I will revert if I cannot fix `getIndexFromLineOffset`

harshad1:
@adelobosko Note that `getIndexFromLineOffset(foo, n, 0)` will move cursor to the _end_ of the nth line, not the start. The difference didn't matter for scrolling to the line, so I did not add a `getLineStart` call.

I do not think performance is an issue, `getIndexFromLineOffset` is used a *lot* in code which must be responsive (such as `regexReplaceAction`) and no issues with behavior or performance are observed. Unless you are calling it in a tight loop, the current code will do.

adelobosko:
> about "ignored dir/file" - should I look into it or have you already started adelobosko

Ok, sure. I think better that you look at it.

adelobosko:
> 
> Note that `getIndexFromLineOffset(foo, n, 0)` will move cursor to the _end_ of the nth line, not the start. 

ok

gsantner:
btw, when this is merged I want to release Markor update

gsantner:
whats the status on the cursor moving? Fine now or are you both happy with it?

gsantner:
whats the status on the cursor moving/line-index? Fine now / are you both happy with it?

harshad1:
> whats the status on the cursor moving/line-index? Fine now / are you both happy with it?

I believe we have agreed do use the old version for now. 

If performance or behavior becomes a concern in the future, I will update the existing function to address it.

adelobosko:
> I believe we have agreed do use the old version for now.

yes, i am confirming

harshad1:
Anything I can do to help move this along?

gsantner:
@harshad1 
getIndexByLineNumber was wrongly merged ahead, now removed again as we keep the old solution with getIndexFromLineOffset. (Which is now slightly improved).

Is this correct? / So reverted as expected already in the PR?

grafik

harshad1:
@gsantner 
> Is this correct? / So reverted as expected already in the PR?

Yes. The current state in the PR is correct. No revert is necessary.

gsantner:
ok thx

I think it's still quite clunky, wanna chop of some more. From functionality remaining is merge of the ignored dir/file settings & functions to a single one

harshad1:
I agree. Let me know if you need me test something

gsantner:
yeah thanks

and sory that it not moves fast currently. lots of personal stuff todo to take care off

gsantner:
yeah thanks

and sorry that it not moves fast currently. lots of personal stuff to take care off.  
(because I've said on weekend I'll merge)

gsantner:
~200-300 lines less again

app/src/main/java/net/gsantner/markor/ui/fsearch/FileSearchResultSelectorDialog.java
```diff
@@ -108,7 +109,7 @@ public void beforeTextChanged(final CharSequence arg0, final int arg1, final int
 
         expandableListView.setOnChildClickListener((parent, v, groupPosition, childPosition, id) -> {
             GroupItemsInfo groupItem = (GroupItemsInfo) parent.getExpandableListAdapter().getGroup(groupPosition);
-            ChildItemsInfo childItem = (ChildItemsInfo) parent.getExpandableListAdapter().getChild(groupPosition, childPosition);
+            SearchEngine.FitFile.ContentMatchUnit childItem = (SearchEngine.FitFile.ContentMatchUnit) parent.getExpandableListAdapter().getChild(groupPosition, childPosition);
```
harshad1, line 112: Why not just use a `Pair<Integer, String>` in both places?

gsantner: I already thought about it and on my own thought that it's easier to follow if it's explicit with name "linenumber".

However changed this now, and at the end it's not that bad as we don't access it much often.

gsantner:
@adelobosko @harshad1   
Thank you very much for programming this. I further slimmed this down throughout the week and I'm gonna merge it now.  
I'm not 100% happy about everything, but it's mature enough for users 😄 . 

Again, thanks! :D

PR #1353, opened 2021-05-30
Multi-select capability to add projects and contexts

harshad1:
In this PR I add multi-select capability to the `SearchOrCustomTextDialog`.

Multi select is so far used to:
1. Insert multiple contexts / projects
2. Select multiple lines (todo only)

I have also applied many clean ups.

Here is what we can do with multi select:

Multi select contexts in context search (long press `@`):
image

Multi select filtered results:
image

Selected tasks are highlighted for editing:
image

Single tap works as before. If multiple tasks are selected they will be selected appropriately:
- If continuous (eg 3, 4 and 5), they will be selected in place.
- If discontinuous (eg, 3, 5, 9) they will be moved to the top and selected

Multi-selected can be used to add multiple contexts, projects and in search in todo.txt


app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
```diff
@@ -195,13 +208,34 @@ protected FilterResults performFiltering(final CharSequence constraint) {
...
+                });
+            } else if (adapter._dopt.neutralButtonCallback != null && adapter._dopt.neutralButtonText != 0) {
+                button.setVisibility(Button.VISIBLE);
+                button.setText(adapter._dopt.neutralButtonText);
+                button.setOnClickListener((v) -> adapter._dopt.neutralButtonCallback.callback());
+            } else {
+                button.setVisibility(Button.INVISIBLE);
+            }
+        }
+    }
```
harshad1, line 230: Biggest change is that if multiple items are selected, the neutral button becomes a 'clear' button. It will revert to the specified button once clear has been pressed (or all items manually deselected).

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTask.java
```diff
@@ -170,7 +167,7 @@ public char getPriority() {
             if (ret.length() == 1) {
                 priority = ret.charAt(0);
             } else {
-                priority = 0;
+                priority = '~'; // No priority == lowest priority
```
harshad1, line 170: `~` has a 'high' value per ascii. This ensures that tasks with no priority have low priority

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -303,27 +281,21 @@ public boolean onLongClick(View v) {
...
+        if (lines.contains("\n") || _appSettings.isTodoAppendProConOnEndEnabled()) {
             runRegexReplaceAction(
                     // Replace existing item with itself. i.e. do nothing
-                    new ReplacePattern(String.format("\\s\\Q%s\\E(:?\\s|$)", item), "$0"),
+                    new ReplacePattern(pattern, "$0"),
                     // Append to end
                     new ReplacePattern("\\s*$", " " + item)
             );
-        } else {
+        } else if (!lines.matches(pattern)) {
```
harshad1, line 294: Update to prevent duplication of contexts and projects

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
```diff
@@ -266,42 +270,50 @@ public static void showSttContextDialog(Activity activity, List<String> availabl
...
+            final List<Integer> filteredIndices = StringUtils.filterIndices(Arrays.asList(allTasks), task ->
+                    StringUtils.containsAny(searchKeys, Arrays.asList(isProjects ? task.getProjects() : task.getContexts())));
+            dopt2.data = StringUtils.map(StringUtils.slice(Arrays.asList(allTasks), filteredIndices), TodoTxtTask::getLine);
+            dopt2.titleText = dopt.titleText;
+            dopt2.searchHintText = R.string.search;
+            dopt2.highlighter = highlighter;
+            dopt2.multiSelectCallback = posns -> userCallback.callback(StringUtils.slice(filteredIndices, posns));
+            dopt2.positionCallback = posn -> dopt2.multiSelectCallback.callback(Collections.singletonList(posn));
+            SearchOrCustomTextDialog.showMultiChoiceDialogWithSearchFilterUI(activity, dopt2);
+        };
```
harshad1, line 313: De-duplicating code for search contexts and search projects. Now supports multi-select. User can now:

1. Multi-select projects (or contexts)
2. See tasks which contain _any_ of the selected projects
3. Multi select tasks
4. Selected tasks will be highlighted

If a user does not multi-select, the dialog works exactly like the old version.

app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
```diff
@@ -99,29 +109,33 @@
...
     }
 
-    private static class WithPositionAdapter extends ArrayAdapter<Pair<String, Integer>> {
+    private static class WithPositionAdapter extends ArrayAdapter<Integer> {
         @LayoutRes
         final int _layout;
         final LayoutInflater _inflater;
         final DialogOptions _dopt;
-        final List<Pair<String, Integer>> _filteredItems;
+        final List<Integer> _filteredItems;
```
harshad1, line 117: Simplified the Adapter to just use List<Integer> instead of the pair, which was unnecessary

app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
```diff
@@ -227,20 +320,43 @@ public void beforeTextChanged(final CharSequence arg0, final int arg1, final int
...
+
+        LinearLayout.LayoutParams slp;
+        slp = new LinearLayout.LayoutParams(WRAP_CONTENT, MATCH_PARENT, 1);
+        slp.gravity = Gravity.START | Gravity.BOTTOM;
+        searchLayout.addView(searchEditText, slp);
+
+        final ImageView clearButton = new ImageView(activity);
+        clearButton.setImageResource(R.drawable.ic_crop_black_24dp);
+        TooltipCompat.setTooltipText(clearButton, activity.getString(R.string.clear));
+        clearButton.setColorFilter(ContextCompat.getColor(activity, dopt.isDarkDialog ? android.R.color.white : R.color.grey));
```
harshad1, line 339: Added a clear button for the searchEditText

app/src/main/java/net/gsantner/opoc/util/Callback.java
```diff
@@ -84,4 +84,24 @@
...
+    public interface r3<R, A, B, C> {
+        R callback(A arg1, B arg2, C arg3);
+    }
+
+    public interface r4<R, A, B, C, D> {
+        R callback(A arg1, B arg2, C arg3, D arg4);
+    }
+
+    public interface r5<R, A, B, C, D, E> {
+        R callback(A arg1, B arg2, C arg3, D arg4, E arg5);
```
harshad1, line 105: New callbacks for arbitrary return type

app/src/main/java/net/gsantner/opoc/util/StringUtils.java
```diff
@@ -317,4 +325,41 @@ public static int tryParseInt(final String value, int defaultValue) {
...
+        return false;
+    }
+
+    public static <T> List<T> slice(final List<T> input, final Iterable<Integer> indices) {
+        final List<T> result = new ArrayList<>();
+        for (final int i : indices) {
+            result.add(input.get(i));
+        }
+        return result;
+    }
```
harshad1, line 364: These functions do not work on strings (they are generic), but I couldn't find a better place for them

app/src/main/java/net/gsantner/markor/format/general/CommonTextActions.java
```diff
@@ -170,12 +170,11 @@ public boolean runAction(String action) {
                 return true;
             }
             case ACTION_SEARCH: {
-                SearchOrCustomTextDialogCreator.showSearchDialog(_activity, _hlEditor.getText(), StringUtils.getSelection(_hlEditor), null, (text, lineNr) -> {
+                SearchOrCustomTextDialogCreator.showSearchDialog(_activity, _hlEditor.getText(), StringUtils.getSelection(_hlEditor), lineNr -> {
```
harshad1, line 173: Simplified the callback.

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -145,20 +132,18 @@ public void onClick(View view) {
...
                 case R.string.tmaid_todotxt_add_context: {
                     final List<String> contexts = new ArrayList<>();
-                    contexts.addAll(Arrays.asList(TodoTxtTask.getContexts(TodoTxtTask.getAllTasks(_hlEditor))));
+                    contexts.addAll(Arrays.asList(TodoTxtTask.getContexts(TodoTxtTask.getAllTasks(_hlEditor.getText()))));
                     contexts.addAll(Arrays.asList(new TodoTxtTask(_appSettings.getTodotxtAdditionalContextsAndProjects()).getContexts()));
-                    SearchOrCustomTextDialogCreator.showSttContextDialog(_activity, contexts, (context) -> {
-                        insertUniqueItem((context.charAt(0) == '@') ? context : "@" + context);
-                    });
+                    SearchOrCustomTextDialogCreator.showSttContextDialog(_activity, contexts, context ->
+                        insertUniqueItem((context.charAt(0) == '@') ? context : "@" + context));
```
harshad1, line 138: Simplifications

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
```diff
@@ -266,42 +271,74 @@ public static void showSttContextDialog(Activity activity, List<String> availabl
         dopt.data = new ArrayList<>(new TreeSet<>(availableData));
         dopt.callback = callback;
         dopt.titleText = isTodoTxtAlternativeNaming(activity) ? R.string.category : R.string.context;
-        dopt.searchHintText = R.string.search_or_custom;
-        //dopt.messageText = activity.getString(R.string.add_x_or_browse_existing_ones_witharg, activity.getString(R.string.context));
+        dopt.multiSelectCallback = (result) -> {
+            for (final Integer pi : result) {
+                callback.callback(dopt.data.get(pi).toString());
+            }
+        };
```
harshad1, line 278: Add multiple contexts or projects with multi-select

harshad1:
Here is what we can do with multi select:

Multi select contexts in context search (long press `@`):
image

Multi select filtered results:
image

Selected tasks are highlighted for editing:
image

Single tap works as before. If multiple tasks are selected they will be selected appropriately:
- If continuous (eg #3, #4 and #5), they will be selected in place.
- If discontinuous (eg, #3, #5, #9) they will be moved to the top and selected

Multi-selected can be used to add multiple contexts, projects and in search in todo.txt

harshad1:
@gsantner This is ready.

If you merge this in and there are conflicts in the search PR, I can resolve the conflicts directly in the search PR

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTask.java
```diff
@@ -79,16 +77,15 @@ public static boolean isTodoFile(String filepath) {
...
-    public static TodoTxtTask[] getAllTasks(final TextView view) {
-        final int[] sel = StringUtils.getSelection(view);
-        return getTasks(view, 0, view.length());
+    public static TodoTxtTask[] getAllTasks(final CharSequence text) {
+        return getTasks(text, 0, text.length());
     }
 
     public static String[] getProjects(final TodoTxtTask[] tasks) {
-        final Set<String> set = new HashSet<>();
+        final TreeSet<String> set = new TreeSet<>();
```
harshad1, line 88: Sorted

harshad1:
@gsantner I have rebased and fixed all conflicts after merge of searchFeature

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
gsantner: so you just fire the callback which does in turn insert/append ... multiple times?

btw what does "pi" mean here?

harshad1: > so you just fire the callback which does in turn insert/append ... multiple times?

Yes. The call back maintains the selection correctly so this works well.

Leaving the variable as `pi` was a mistake. It was originally 'project index' before I combined the project and context functions. Should change it to `di` for data index or something.

gsantner: just i or index is finr

harshad1: changed

gsantner:
cleaned up PR comments a little, removed outdated comments

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -145,20 +130,18 @@ public void onClick(View view) {
...
```
gsantner, line 136:
```suggestion
                    SearchOrCustomTextDialogCreator.showSttContextDialog(_activity, contexts, (context) -> {
                        insertUniqueItem((context.charAt(0) == '@') ? context : "@" + context)
                    });
```

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -231,7 +214,7 @@ public void onClick(View view) {
                     SearchOrCustomTextDialogCreator.showSttSortDialogue(_activity, (orderBy, descending) -> new Thread() {
                         @Override
                         public void run() {
-                            List<TodoTxtTask> tasks = Arrays.asList(TodoTxtTask.getAllTasks(_hlEditor));
+                            List<TodoTxtTask> tasks = Arrays.asList(TodoTxtTask.getAllTasks(_hlEditor.getText()));
```
gsantner, line 217:
```suggestion
                            final List<TodoTxtTask> tasks = Arrays.asList(TodoTxtTask.getAllTasks(_hlEditor.getText()));
```

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -286,7 +262,7 @@ public boolean onLongClick(View v) {
                 case R.string.tmaid_todotxt_sort_todo: {
                     final List<String> last = new AppSettings(getContext()).getStringList(LAST_SORT_ORDER_KEY);
                     if (last != null && last.size() == 2) {
-                        List<TodoTxtTask> tasks = Arrays.asList(TodoTxtTask.getAllTasks(_hlEditor));
+                        List<TodoTxtTask> tasks = Arrays.asList(TodoTxtTask.getAllTasks(_hlEditor.getText()));
```
gsantner, line 265:
```suggestion
                        final List<TodoTxtTask> tasks = Arrays.asList(TodoTxtTask.getAllTasks(_hlEditor.getText()));
```

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -303,27 +279,21 @@ public boolean onLongClick(View v) {
 
     private void insertUniqueItem(String item) {
         item = item.trim().replace(" ", "_");
-        if (!selectionIsSingleTask() || _appSettings.isTodoAppendProConOnEndEnabled()) {
+        final Pattern pattern = Pattern.compile(String.format("\\s\\Q%s\\E(:?\\s|$)", item));
```
gsantner, line 282: can you please add a short inline comment what this pattern is for / does?

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
```diff
@@ -259,42 +262,68 @@ public static void showSttContextDialog(Activity activity, List<String> availabl
...
+        dopt.data = Arrays.asList(isProjects ? TodoTxtTask.getProjects(allTasks) : TodoTxtTask.getContexts(allTasks));
+        if (isTodoTxtAlternativeNaming(activity)) {
+            dopt.titleText = isProjects ? R.string.tag : R.string.category;
+        } else {
+            dopt.titleText = isProjects ? R.string.project : R.string.context;
         }
-        return values;
-    }
+        dopt.searchHintText = R.string.search_or_custom;
+        dopt.iconsForData = Collections.nCopies(dopt.data.size(), R.drawable.ic_search_black_24dp);
```
gsantner, line 284: nice

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
```diff
@@ -259,42 +262,68 @@ public static void showSttContextDialog(Activity activity, List<String> availabl
...
+            for (final int i : keyIndices) {
+                searchKeys.add(dopt.data.get(i).toString());
             }
-        }
-        return data;
+
+            final List<Integer> filteredIndices = new ArrayList<>();
+            final List<String> taskLines = new ArrayList<>();
+            for (int i = 0; i < allTasks.length; i++) {
+                final TodoTxtTask task = allTasks[i];
```
gsantner, line 299:
```suggestion
            for (final TodoTxtTask task : allTasks) {
```

app/src/main/java/net/gsantner/opoc/util/StringUtils.java
```diff
@@ -298,4 +305,22 @@ public static int tryParseInt(final String value, int defaultValue) {
             return defaultValue;
         }
     }
+
+    // So one doesn't have to define all three always
+    public static class SimpleTextWatcher implements TextWatcher {
```
gsantner, line 310: Recommendation: TextWatcherAdapter (i.e. normal java interface listener vs adapter)

app/thirdparty/java/other/writeily/model/WrMarkorSingleton.java
```diff
@@ -169,7 +169,7 @@ private void _moveOrCopySelected(
                     // Put the file back in
                     files.push(file);
                     SearchOrCustomTextDialogCreator.showCopyMoveConflictDialog(
-                            activity, file.getName(), destDir.getName(), files.size() > 1, (name, option) -> {
+                            activity, file.getName(), destDir.getName(), files.size() > 1, option -> {
```
gsantner, line 172:
```suggestion
                            activity, file.getName(), destDir.getName(), files.size() > 1, (option) -> {
```

gsantner (commented):
# Review 1
added some suggested changes, can be directly commited/accepted in github web interface

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
harshad1: I use the index in line 303

```Java
                      filteredIndices.add(i);
```

app/src/main/java/net/gsantner/opoc/util/StringUtils.java
harshad1: Just got rid of it. Only used in 2 places anyway.

harshad1:
Your changes look fine to me

app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
```diff
@@ -38,26 +41,33 @@
...
 public class SearchOrCustomTextDialog {
 
     public static class DialogOptions {
-        public Callback.a1<String> callback;
-        public Callback.a2<String, Integer> withPositionCallback;
-        public List<? extends CharSequence> data;
-        public List<? extends CharSequence> highlightData;
+        public Callback.a1<String> callback = null;
+        public Callback.a1<Integer> positionCallback = null;
+        public Callback.a1<List<Integer>> multiSelectCallback = null;
```
gsantner, line 68: To be honest: I prefer to have a single callback ("way out" / single return value). Thats what I also thought when I stumbled across it recently with the second "positionCallback" addition.

My suggestion is:
* `Callback.a1<String, MapPropertyBackend<String>>`
* with the first parameter being the usual selection (i.e. first selection)
* second parameter is a hashmap with getString / getInt etc
* There you can store i.e. position int and any additional data not required for the simple search

harshad1: I thought of unifying all the callbacks into a single  `Pair<String, List<Integer>>`. The problem here is the calling code has to then have logic to decide what to do based on what flags were set and what is in the callback.

With the current system in this PR we have

1. `callback`: string in search or item selected
2. `positionCallback` : position of item selected
3. `multiSelectCallback`: all selected positions. Also serves as a flag that multi-select is enabled.

Calling code just sets the appropriate callback(s) to do the correct thing and doesn't have to have any if-then logic.

One option we do to simplify things here is unify the position callbacks and have an explicit multi-select flag.

gsantner: Only place we need one position:   In editor content search

Places where multiple positions:  
todo tag  
todo project  

everything else just needs the string result, correct?

gsantner: what do you think off `<String, Integer...>`? Should cover all cases as well

harshad1:
I changed it to just 2 callbacks to see how it looks:
1. `callback` -> just the string
2. `positionCallback` -> the list of indices. 

I can try the combined callback and look at that too, but I think the separate callbacks are cleaner as it allows the calling code to be simpler. Don't need to handle cases which aren't relevant - you can provide _only_ `callback`, leave `positionCallback = null` and get the old behavior without any downsides.

harshad1:
@gsantner Any opinions on the current system (2 callbacks). I added some documentation and an explicit `@Nullable` flag to the callbacks too

gsantner (approved):

app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
```diff
@@ -143,12 +143,10 @@ public View getView(int pos, @Nullable View convertView, @NonNull ViewGroup pare
             final TextView textView;
             if (convertView == null) {
                 textView = (TextView) _inflater.inflate(_layout, parent, false);
-                textView.setBackgroundResource(R.drawable.search_dialog_selection);
```
harshad1: @gsantner This needs to be reverted. Without the border-only highlight, the visibility of highlighted text against the selected background is very poor.

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
```diff
@@ -282,7 +282,6 @@ public static void showSttKeySearchDialog(final Activity activity, final CharSeq
             dopt.titleText = isProjects ? R.string.project : R.string.context;
         }
         dopt.searchHintText = R.string.search_or_custom;
-        dopt.iconsForData = Collections.nCopies(dopt.data.size(), R.drawable.ic_search_black_24dp);
```
harshad1: @gsantner I thought you liked this ?!

It makes it very clear that we are in the search dialog and not in the insert dialog

app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
gsantner: default android behaviour is way better to understand what is highlighted and what not to me:

screenshot-2021-07-14__18-04-55


screenshot-2021-07-14__18-04-41

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
gsantner: I liked the line of code, but in use it looked bad. 10 identical search icons are bloat

harshad1: Ah. Ok.

Other options are:

1. Dialog level icon (relatively simple addition, this is my preference)
2. Change the dialog title text

gsantner: Why do you see the need for a change? It's not that much bad IMO

harshad1: I would change it to 'Search tag' or 'search context'

Insert context is activated by short press, search context is activated by long press. 

Right now those two dialogs look identical.

app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
harshad1: image

Issue is when highlighting is present

gsantner: I'd rather pefer to make the text bold, or i.e. accent colored

Or/And disable syntax highlighting for selected elements

harshad1: Of these, the checkmark icon is the cleanest solution, IMO. I will work on a PR for it.

Issue #1025, opened 2020-08-09
Reading position of a note gets lost after device rotation


geniuslittleteddy:
#### General information
Reading position of a long note could get lost after device orientation changes. It's annoying to manually scroll to the target position I read before.
* **App version:** 2.2.10
* **System:** Android 7

#### Description
Some screen shots here
![Screen Shot 2020-08-09 at 4 16 21 PM](https://user-images.githubusercontent.com/67744580/89728001-22911000-da5c-11ea-9ed1-616c92ce2618.png)
![Screen Shot 2020-08-09 at 4 16 32 PM](https://user-images.githubusercontent.com/67744580/89728002-245ad380-da5c-11ea-982c-34e8a20ed6bc.png)

matobarny:
hey @gsantner I would like to solve this issue. Can you assign me?

matobarny:
If this issue is still available I would be thankfull if I could try it and learn something new. Thank you.

gsantner:
Hey matobarny, thanks for help!

One note: At same time the content gets loaded, not everything might be fully loaded. So i.e. you probably will require a delay timer of like 500ms after setting content prior to apply cursor pos. Good luck!

gsantner:
@matobarny I haven't heard something back, so I guess you never started?

matobarny:
I was working on it and then I was really bussy but I want to finish it
soon.

pi 4. 12. 2020 o 0:08 Gregor Santner <notifications@github.com> napísal(a):

matobarny:
I hope till the next week it will be done :)

pi 4. 12. 2020 o 10:18 Martin Baran <martin.baran5@gmail.com> napísal(a):

DrRajivKulkarni:
New search is fantastic. But once, any one search result is opened the searched results are lost.

gsantner:
1. this is completley unrelated to the #1025
2.  There is a history of recent searches at file search, which you can use to redo search

CutieDeng:
Maybe I can provide some ideas for it: #1733.

gsantner:
From 2195 XYIheng

> In a long text file. I scroll down and append some txt at the end of the file, then click view mode. The file goes to the beginning instead of keeping scroll position

https://github.com/gsantner/markor/assets/31406320/8314a63b-0825-422a-82b4-a5355c82ca1a

jcfyre:
[...]

PR #1379, opened 2021-06-30
Restore scroll position (#1025, #1242)

adelobosko:
Need help: 
 - Keyboard is not hiding when moving the app to the background.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
@@ -49,6 +49,7 @@ import net.gsantner.markor.format.general.DatetimeFormatDialog;
 import net.gsantner.markor.model.Document;
 import net.gsantner.markor.ui.AttachImageOrLinkDialog;
 import net.gsantner.markor.ui.DraggableScrollbarScrollView;
+import net.gsantner.markor.ui.DraggableScrollbarWebView;
 import net.gsantner.markor.ui.FileInfoDialog;
 import net.gsantner.markor.ui.FilesystemViewerCreator;
 import net.gsantner.markor.ui.SearchOrCustomTextDialogCreator;
@@ -78,8 +79,14 @@ public class DocumentEditFragment extends GsFragmentBase implements TextFormat.T
     public static final int HISTORY_DELTA = 5000;
     public static final String FRAGMENT_TAG = "DocumentEditFragment";
     private static final String SAVESTATE_DOCUMENT = "DOCUMENT";
-    private static final String SAVESTATE_CURSOR_POS = "CURSOR_POS";
     private static final String SAVESTATE_PREVIEW_ON = "SAVESTATE_PREVIEW_ON";
+    public static DocDisplayType _docInitType;
+
+    private enum DocDisplayType {
+        FILE_MANAGER,
+        TODO,
+        QUICK_NOTE
+    }
 
     private AppSettings _appSettings;
     private HorizontalScrollView hsView;
@@ -122,7 +129,7 @@ public class DocumentEditFragment extends GsFragmentBase implements TextFormat.T
     TextView _textSdWarning;
 
     @BindView(R.id.document__fragment_view_webview)
-    WebView _webView;
+    DraggableScrollbarWebView _webView;
 
     @BindView(R.id.document__fragment__edit__content_editor__scrolling_parent)
     DraggableScrollbarScrollView _primaryScrollView;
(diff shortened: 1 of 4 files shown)


app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -78,8 +79,14 @@
...
     private static final String SAVESTATE_DOCUMENT = "DOCUMENT";
-    private static final String SAVESTATE_CURSOR_POS = "CURSOR_POS";
     private static final String SAVESTATE_PREVIEW_ON = "SAVESTATE_PREVIEW_ON";
+    public static DocDisplayType _docInitType;
+
+    private enum DocDisplayType {
+        FILE_MANAGER,
+        TODO,
+        QUICK_NOTE
+    }
```
gsantner, line 89: We don't need to track the filetype, as long we are at MainActivity it is OK to do the special handling. We don't care if it's a todo.txt, markdown etc

gsantner:
> Keyboard is not hiding when moving the app to the background.

There should be a existing hidekeyboard function in opoc. And we have activity lifecycle methods like onPause etc

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
adelobosko: > 
> 
> We don't need to track the filetype, as long we are at MainActivity it is OK to do the special handling. We don't care if it's a todo.txt, markdown etc. At same time, we also have filepath and appsettings for comparision. And for fragment visible detection there is also something.

I need to know what is used in the main activity (ToDo or QuickNote), because they both will be opened, and the keyboard will open for ToDo and QuickNote as a result keyboard will be bugged.
We need to open the keyboard only for active doc, and do not open for docs that in the background. That why I have created this detection.

adelobosko: > and the keyboard will open for ToDo and QuickNote as a result keyboard will be bugged.

you can try without checking:
then try to open ToDo in the main activity then hide the app, and then back to the app.
At first, the keyboard will select the cursor position for ToDo and open the keyboard for ToDO, then for QuickNote (that in the background) as result will be a bug for ToDo will not show a keyboard after resume, and for QuickNote the keyboard will be shown as it opened last...

https://user-images.githubusercontent.com/56083544/124092942-0de8ac80-da60-11eb-80d1-c2ddf20d08ed.mp4

adelobosko:
> 
> 
> > Keyboard is not hiding when moving the app to the background.
> 
> There should be a existing hidekeyboard function in opoc. And we have activity/fragment lifecycle methods like onPause etc

I move `new ActivityUtils(getActivity()).hideSoftKeyboard().freeContextRef();` to the onStop but it did not help.
The function call all sub-functions but the keyboard is not closed...
These android / java annoying me a little bit... ;)

adelobosko:
Hallelujah, this helped me `imm.toggleSoftInput(InputMethodManager.SHOW_FORCED,0 );`

adelobosko:
> 
> 
> Hallelujah, this helped me `imm.toggleSoftInput(InputMethodManager.SHOW_FORCED,0 );`

it is working, but it is a toggle...

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
gsantner: OK I see, you need to know which of the files is opened currently. In this case you can also use AppSettings to get the filepath and compare it to the current Document.getFile

harshad1:
@adelobosko I am not sure I understand exactly what this PR is trying to solve? Could you explain in a line or two?

gsantner:
@adelobosko 

I pushed some changes to reduce amount of duplicate code. Can you please check if it still works as you expect from these changes? QuickNote and ToDo properly start on bottom for me with this PR.

app/src/main/java/net/gsantner/markor/util/MarkorWebViewClient.java
```diff
@@ -89,19 +87,7 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
...
     @Override
     public void onPageFinished(WebView view, String url) {
-        if (_restoreScrollYEnabled) {
-            for (int dt : new int[]{50, 100, 150, 200, 250, 300}) {
-                view.postDelayed(() -> view.setScrollY(_restoreScrollY), dt);
-            }
-            _restoreScrollYEnabled = false;
-        }
         super.onPageFinished(view, url);
     }
```
gsantner, line 91: if only @ override and super. is left, method can be removed

gsantner:
@adelobosko 

How "save" is the PR to merge? Do you have additional improvements in your mind,or do you see a bigger remaining/new issue from these changes?

adelobosko:
@gsantner yes, your changes are worked, but I think need to decline or enhance this PR, other keyboard-related bugs appear.
i will try to update it on the weekend

gsantner:
OK then I don't think it will make it for the update. Need to release update before weekend.

PR #1388, opened 2021-07-10
Search encrypted files, closes #1375

### LINKED ISSUE #1375: Support in-content filesearch for encrypted textfiles
#### General information

* **App version:** 2.7.0
* **System:** Android 10

#### Description
I wonder if it is possible that even encrypted files will be part of the search.
Are there any reason why this isn't implemented? Is the password there not accessible?

opensource21:
I added the functionality to search even encrypted files. #1337 #1375


app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
```diff
@@ -125,11 +125,17 @@ public static void showInsertTableRowDialog(final Activity activity, final boole
...
+                password = new AppSettings(activity).getDefaultPassword();
+            } else {
+                password = new char[0];
+            }
             Callback.a1<SearchEngine.SearchOptions> fileSearchDialogCallback = (searchOptions) -> {
                 searchOptions.rootSearchDir = searchDir;
                 SearchEngine.queueFileSearch(activity, searchOptions, searchResults -> {
                     FileSearchResultSelectorDialog.showDialog(activity, searchResults, callback);
-                });
+                }, password);
```
gsantner, line 138: there is a data container searchOptions which should fit instead of creating an specific additional parameter

opensource21:
If it okay in principle I would make some performance-measurements.

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
opensource21: Good Point. I fixed it.

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
```diff
@@ -125,8 +125,15 @@ public static void showInsertTableRowDialog(final Activity activity, final boole
 
     public static void showSearchFilesDialog(Activity activity, File searchDir, Callback.a2<String, Integer> callback) {
         if (!SearchEngine.isSearchExecuting) {
+            final char[] password;
+            if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M) {
+                password = new AppSettings(activity).getDefaultPassword();
+            } else {
+                password = new char[0];
```
harshad1, line 132: Maybe `null` would be better flag?

gsantner: can be removed complete, for the default case there is default value at decl

app/src/main/java/net/gsantner/markor/ui/fsearch/SearchEngine.java
```diff
@@ -429,5 +439,24 @@ private boolean isIgnored(File directory) {
...
+                }
+            } else {
+                return new FileInputStream(file);
+            }
+        }
+    }
+
+
+    private static boolean isEncryptedFile(File file) {
+        return Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && file.getName().endsWith(JavaPasswordbasedCryption.DEFAULT_ENCRYPTION_EXTENSION);
```
gsantner, line 460: also mind encryption for isIgnored function (.*.txt filter at .txt.enc VS .txt file)

app/src/main/java/net/gsantner/markor/ui/fsearch/SearchEngine.java
```diff
@@ -429,5 +439,24 @@ private boolean isIgnored(File directory) {
...
+
+        private InputStream getInputStream(File file) throws FileNotFoundException {
+            if (isEncryptedFile(file)) {
+                final byte[] encryptedContext = FileUtils.readCloseStreamWithSize(new FileInputStream(file), (int) file.length());
+                if (encryptedContext.length > JavaPasswordbasedCryption.Version.NAME_LENGTH && _config.password.length > 0) {
+                    final byte[] decrypt = new JavaPasswordbasedCryption(getVersion(encryptedContext), null).decryptBytes(encryptedContext, _config.password.clone());
+                    return new ByteArrayInputStream(decrypt);
+                } else {
+                    return new ByteArrayInputStream(encryptedContext);
+                }
```
gsantner, line 451: don't we have a function for this already somewhere. looks a lot like duplicated code here again

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
```diff
@@ -125,8 +125,15 @@ public static void showInsertTableRowDialog(final Activity activity, final boole
 
     public static void showSearchFilesDialog(Activity activity, File searchDir, Callback.a2<String, Integer> callback) {
         if (!SearchEngine.isSearchExecuting) {
+            final char[] password;
+            if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M) {
+                password = new AppSettings(activity).getDefaultPassword();
```
gsantner, line 130: please have the assignment in FileSearchDialog.java, there we also assign other AppSettings default values

grafik

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
```diff
@@ -125,8 +125,15 @@ public static void showInsertTableRowDialog(final Activity activity, final boole
 
     public static void showSearchFilesDialog(Activity activity, File searchDir, Callback.a2<String, Integer> callback) {
         if (!SearchEngine.isSearchExecuting) {
+            final char[] password;
```
gsantner, line 128: please have the assignment in FileSearchDialog.java, there we also assign other AppSettings default values

grafik

gsantner (changes requested):

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
opensource21: Good Point!

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
opensource21: Remove completly

app/src/main/java/net/gsantner/markor/ui/fsearch/SearchEngine.java
opensource21: Not exactly, but it was a form of pre optimization. Fixed

app/src/main/java/net/gsantner/markor/ui/fsearch/SearchEngine.java
opensource21: I would say that the caller should be able to define an exact filter. So if he want to ignore encrypted files he can do it. Or if he want to search only in encrypted file. It's tricky to implement there an automatism. 
BTW Base on variable names it looks like ignoring directories, but it seems that it ignores files...

gsantner (approved):

PR #1392, opened 2021-07-16
Release Markor update v2.8.0

gsantner:


gsantner:
@harshad1 @adelobosko @fredericjacob @opensource21 

I plan to release a smaller Markor update this month. If you have some important fixes or alike on the stage, please make a PR soon enough.  
This will be the last Google Play update for a while, at least till the next bigger release. Regarding appbundle, giving Google my signing keys, regarding minSdk 30, fs limitations and other stuff. Have to think about it.

Thanks, Gregor

opensource21:
The only thing I miss is the global search in encrypted file. The PR should be fine.
Search-replace would be nice in editor, but I think that needs more refinements.
You create a really nice tool!

harshad1:
> Search-replace would be nice in editor, but I think that needs more refinements.

We already have search replace though.

opensource21:
Indeed - don't know why I didn't found it. Sorry.

gsantner:
Please help to test / check if it's OK or if you notice something new bad on master.  
There are quite some changes to dialogs these days.

I want to start with the release process tomorrow.

harshad1:
> This will be the last Google Play update for a while

Do you plan to continue releasing on F-Droid in the meanwhile?

gsantner:
I dont have thought on gplay further yet, but fdroid and github release procedure will not be different

gsantner:
(so yes)

PR #1396, opened 2021-07-17
Improve multi-select dialog UI

harshad1:
In this PR I have re-worked the multi select feature

Now, if multi select is enabled
1. Single click selects / unselects
2. Long click goes directly to the item


app/src/main/res/values/strings.xml
```diff
@@ -433,4 +433,13 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
...
     <string name="infinite">Infinite</string>
+    <string name="search_project">Search projects</string>
+    <string name="search_context">Search contexts</string>
+    <string name="search_category">Search categories</string>
+    <string name="search_tag">Search tags</string>
+    <string name="insert_project">Insert projects</string>
+    <string name="insert_context">Insert contexts</string>
+    <string name="insert_category">Insert categories</string>
+    <string name="insert_tag">Insert tags</string>
+    <string name="clear_selected">Clear selected</string>
```
harshad1, line 444: Much better wording. No ambiguity now

app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
```diff
@@ -431,6 +349,79 @@ public void beforeTextChanged(final CharSequence arg0, final int arg1, final int
...
+            } else {
+                directActivate.callback(pos);
+            }
+        });
+
+        // long click always activates
+        listView.setOnItemLongClickListener((parent, view, pos, id) -> directActivate.callback(pos));
+
+        // Initialize neutral button state
+        setNeutralButtonState.callback();
```
harshad1, line 425: Simpler logic, moved into the main body

app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
```diff
@@ -217,99 +220,12 @@ protected FilterResults performFiltering(final CharSequence constraint) {
...
-        }
-
-        // Long click always selects, if multi select is possible
-        if (dopt.isMultiSelectEnabled) {
-            listView.setOnItemLongClickListener((parent, view, pos, id) -> {
-                toggleSelection.callback(pos, view);
-                return true;
-            });
-        }
-    }
```
harshad1, line 305: No longer need the setDialogState state machine

harshad1:
image

app/src/main/res/values/strings.xml
gsantner: there is also alternative todo wording. Better "Insert %s"

harshad1:
I have the alternative naming titles explicitly set right now.

I thought of using 'Insert %' but figured this would make it harder for translators.

PR is not quite ready btw. I still need to tweak the view heights.

Jul. 16, 2021 23:51:41 Gregor Santner ***@***.***>:

harshad1:
I am happy with the state now

harshad1:
@gsantner will you be releasing this with 2.8?

app/src/main/res/values/strings.xml
gsantner: Clear selection

harshad1:
@gsantner Are there other changes you'd like?

app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
```diff
@@ -41,12 +41,14 @@
...
 import android.widget.ArrayAdapter;
 import android.widget.Button;
+import android.widget.CheckedTextView;
 import android.widget.Filter;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.ListView;
 import android.widget.TextView;
 
+import net.gsantner.markor.R;
```
gsantner, line 51: Do not use/reference `markor.R` in this component.

gsantner:
Items are not pre-checked, but thats what you meant with the comment on the oher "view mode" issue I guess.

Also noticed that the UI pushes up when clicking something, which is kinda awkward

<https://user-images.githubusercontent.com/6735650/127147182-76af6078-4fff-465b-82df-23bce2c65b3b.mp4>

gsantner:
* Removed the custom layout, use aosp one instead (automatically takes appcompat/androidx design anyway)
* Seen that your items where smaller in height (height=wrap_content.) This is now applied for both, multi and normal selection.
* Lowered the "Clear selection (x)" text to "Deselect (x)". At least in english, layout then not reordered and stays compact without big additional whitespace. Was too long.

Looks same like with your suggested custom layout to me at last

screenshot-2021-07-27__14-39-16

gsantner:
* Removed the custom layout, use aosp one instead (automatically takes appcompat/androidx design anyway)
* Seen that your items where smaller in height (height=wrap_content.) This is now applied for both, multi and normal selection.
* Lowered the "Clear selection (x)" text to "Deselect (x)". At least in english, layout then not reordered and stays compact without big additional whitespace. Was too long.
* Also increased margin at the X clear search input button, now better aligned with checkboxes

Looks same like with your suggested custom layout to me at last

without X change

screenshot-2021-07-27__14-39-16


with X 

screenshot-2021-07-27__14-44-31

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
```diff
@@ -391,6 +395,7 @@ public static void showSearchDialog(Activity activity, Editable edit, int[] sel,
     public static void showTodoSearchDialog(Activity activity, Editable edit, int[] sel, Callback.a1<Spannable> highlighter, Callback.a1<List<Integer>> userCallback) {
         SearchOrCustomTextDialog.DialogOptions dopt2 = basicSearchDialogopts(activity, edit, sel);
         dopt2.isMultiSelectEnabled = true;
+        dopt2.neutralButtonText = R.string.deselect;
```
harshad1, line 398: @gsantner 

Now the search / replace option isn't present

With the old approach you got `Deselect` when multiple options were selected and `Search / replace` when none were selected.

gsantner: Where do we need this? Search/Replace is only important for the document top-toolbar search button, which not has Multi-Selection anyway

harshad1: This change removed search / replace for all file types.

> which not has Multi-Selection anyway

I had multi selection enabled for the top search for Todo file type, where it is useful.

I'll make a PR to fix this tomorrow

harshad1: The other option is to keep the neutral button constant and turn _cancel_ into Deselect when more than 1 item selected.

Do you prefer this?

gsantner: Humble said I generally don't see much of the need for "Unselect all". Especially not for todo, think it might hurt more than it brings [oh no, need to recheck all my finished stuff].

So if you goal to improve for todo.txt I rather suggest to keep the neutral slot free for an additional action.

> The other option is to keep the neutral button constant and turn cancel into Deselect when more than 1 item selected.

No, at dialog it's important to have a button to go out without changes saved. We should not add a dual-use for it

gsantner: An important point also, in case you think off synchronizing the checked states with editor: You can only set cursor across one rectangle of items, if there is any non-selected in between, it will be spanned by the selection as well.

harshad1: > plan to do with the checkboxes atm - i.e. apply the "done" states

Apply done, change priority, add context or project, add or change due date. I do all of these for multiple tasks all the time.

> I generally don't see much of the need for "Unselect all"

Ok. I guess the user could always cancel and come back if they needed to. What we can do is just add the selection count in parenthesis after the _title_? This will also simplify the logic a great deal.

gsantner: So what did you have in mind with "Unselect all", do you really use or plan to use it often? Or just a nice to have?

Not wanna destroy your use case. Just guess most of the time it won't be used

>  What we can do is just add the selection count in parenthesis after the title? This will also simplify the logic a great deal.

It's a option, in case we even want to display the amount of selected items. Could be Title [x] or also at OK [x] button

harshad1: It seemed natural to have an unselect all. I don't seem to actually use it that often so. We can add it back if I find that I or others really feel the need for it.

> or also at OK [x] button. 

I like this more than title (x) actually. Makes it clearer.

PR #1402, opened 2021-07-29
Fix multiselect item height

harshad1:
I found that I often mis-clicked items with the smaller height. This corrects that

### DIFF
--- a/app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
+++ b/app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
@@ -145,8 +145,7 @@ public class SearchOrCustomTextDialog {
             final TextView textView;
             if (convertView == null) {
                 textView = (TextView) _inflater.inflate(_layout, parent, false);
-                ViewGroup.LayoutParams params = textView.getLayoutParams();
-                params.height = ViewGroup.LayoutParams.WRAP_CONTENT;
+                textView.setLayoutParams(new LinearLayout.LayoutParams(LinearLayout.LayoutParams.MATCH_PARENT, LinearLayout.LayoutParams.WRAP_CONTENT));
                 textView.setMinHeight(_layoutHeight);
             } else {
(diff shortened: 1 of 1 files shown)


app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
```diff
@@ -110,23 +111,30 @@
...
+        private final int _layoutHeight;
         private final LayoutInflater _inflater;
         private final DialogOptions _dopt;
         private final List<Integer> _filteredItems;
         private final Set<Integer> _selectedItems;
         private final Pattern _extraPattern;
 
         public static Adapter create(final Context context, final DialogOptions dopt) {
-            return new Adapter(context, dopt, new ArrayList<>());
+            return new Adapter(context, dopt, dopt.isMultiSelectEnabled ? android.R.layout.simple_list_item_multiple_choice : android.R.layout.simple_list_item_1, new ArrayList<>());
```
harshad1, line 124: Moved layout selection to constructor. Decision is made in one place only

app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
```diff
@@ -136,15 +144,17 @@ public View getView(int pos, @Nullable View convertView, @NonNull ViewGroup pare
...
+                params.height = ViewGroup.LayoutParams.WRAP_CONTENT;
+                textView.setMinHeight(_layoutHeight);
             } else {
                 textView = (TextView) convertView;
             }
 
             if (textView instanceof Checkable) {
                 ((Checkable) textView).setChecked(_selectedItems.contains(index));
             }
-            textView.setLayoutParams(new LinearLayout.LayoutParams(LinearLayout.LayoutParams.MATCH_PARENT, LinearLayout.LayoutParams.WRAP_CONTENT));
```
harshad1, line 147: Layout setting moved into if - don't need to set layout every time. Just on creation.

PR #1417, opened 2021-08-21
Allow to install APK files from filebrowser on click, by @gsantner

gsantner:
Allow to install APK's from the file browser.

I have about 3 file & document managers and multiple of browsers on my device but seemingly nothing is capable of requesting installation of APK in 2021.

You have to allow the app to install apps once, similar like below.

grafik


harshad1:
FWIW, I use Simple File Manager (A simple file manager for browsing and editing files and directories.) - https://f-droid.org/packages/com.simplemobiletools.filemanager.pro

Issue #1413, opened 2021-08-14
Todo.txt should support preserving priority on completion (via pri:A)


harshad1:
Currently, when a todo item is completed, the priority is discarded. Instead it should be preserved as a `pri:` tag. The priority should be restored if the todo is uncompleted.

[See todo.txt readme](https://github.com/todotxt/todo.txt)

gsantner:
it also says `Many Todo.txt clients discard priority on task completion.`, which is perfectly fine. I've used quite some todo.txt software and I did not use a single one which would have added such a tag

harshad1:
Several of the todo.txt managers I use seem to support it.

IMO having it enabled by default (i.e. not optional) it won't disrupt the workflow for anyone who doesn't need it.

gsantner:
how does it behave when un-checking? remove pri:?

harshad1:
Remove pri:X and use that value as the priority.

Aug. 15, 2021 05:38:44 Gregor Santner ***@***.***>:

gsantner:
yeah makes sense

Hyperlynx2:
Today I learned!

In my manual, non-phone todo.txt I had just been adding an x to the start of the line and then the date, keeping the priority as is. I was _just_ about to open an issue complaining that this "pri:" business is nonstandard and could I please have it the other way, when I found this ticket and the corresponding link to the spec. I've been doing it wrong!

Nice work @gsantner and @harshad1!

PR #1420, opened 2021-08-24
Using pri:X tags to preserve priority

harshad1:
This addresses #1413

Everything seems to work, ~~but marked WIP as there are a ton of edge cases which require more testing.~~


harshad1:
I am happy with this

gsantner (approved):

PR #1427, opened 2021-09-05
Add option to view headings up to H6 in the table of contents

fredericjacob:
This PR adds the option to view headings up to H6 in the table of contents in view mode.

The original checkbox in the markdown settings allowed to choose between no table of contents (unchecked) and a table of contents with headings up to H3 (checked).
With these changes, the checkbox is replaced with a radio button preference that allows to choose between no table of contents, a simple table of contents (like the existing, up to H3) and an "extended" table of contents with headings up to H6.

Background: I have quite a few files in my notebook with headings beyond H3 which I was missing in the view mode table of contents. I feel others might profit from more heading levels as well, while some may still prefer a smaller table of contents which shows only the topmost (presumably most relevant) heading levels up to H3.

grafik

grafik

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
+++ b/app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
@@ -180,8 +180,7 @@ public class MarkdownTextConverter extends TextConverter {
             }
 
             head += CSS_TOC_STYLE;
-            final int levels = TocOptions.getLevels(appSettings.getMarkdownTableOfContentLevels());
-            options.set(TocExtension.LEVELS, levels)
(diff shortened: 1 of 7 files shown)


gsantner (changes requested):
What I suggest instead: To have checkboxes to toggle on/off all levels. If we add option to select none/23/2456, one comes up why there is no 1, or they don't want 6.

fredericjacob:
@gsantner That's a good point - I changed the setting accordingly. Now it's possible to toggle each heading level separately:

H1 to H3:
grafik
grafik

H2, H5, H6:
grafik
grafik

No table of contents:
grafik
grafik

gsantner (approved):

PR #1434, opened 2021-09-11
ZimWiki: Syntax highlighting for code, global setting

fredericjacob:
This PR adds the same functionality for ZimWiki for highlighting/formatting code which is already present for Markdown.

Before, code (inline and blocks) was always formatted as monospace, which led to severe performance problems in larger files with a lot of code samples.

Now, formatting code as monospace is optional (and turned off by default), but code is highlighted with gray background color (like for markdown). There's also the option to disable highlighting.

grafik
grafik

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/markdown/MarkdownHighlighter.java
+++ b/app/src/main/java/net/gsantner/markor/format/markdown/MarkdownHighlighter.java
@@ -38,9 +38,9 @@ public class MarkdownHighlighter extends Highlighter {
         super(hlEditor, document);
         _highlightLinks = false;
         _highlightLineEnding = _appSettings.isMarkdownHighlightLineEnding();
-        _highlightCodeChangeFont = _appSettings.isMarkdownHighlightCodeFontMonospaceAllowed();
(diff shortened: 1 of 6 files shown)


gsantner:
Can you please move the setting to editmode -> syntay hl? So one setting that is applied to all formats. Intlstead of introducing duplicate for zim.

fredericjacob:
Yes, I will change this. 👍

By the way, how about also moving the markdown table of contents setting to the general view mode settings? As zim view mode makes use of the MarkdownTextConverter, the setting affects both markdown and zim table of contents. (I can create a separate PR for this if you like)

gsantner:
toc is under rework

fredericjacob:
Et voilà, I moved both settings to the general "Editor > Syntax Highlighting" section, they are now commonly used by markdown and zim wiki.

> toc is under rework, dont touch now

Ah okay, didn't know that the table of contents is worked on at the moment, is there some PR or are you just doing some changes? Then it could make sense to just keep the inconsistent TOC setting location in mind and consider moving it in the future (to make more clear that it's used by markdown and zim)

PR #1467, opened 2021-10-10
Save todo search views

harshad1:
This PR adds a button to save a todo search view.

1. Open the browse dialog
2. Browse by (say) context
3. Select context(s) and hit ok
4. Neutral button will allow the current view to be saved (user will be prompted for name)
5. Saved view will be available in browse dialog

Todo:
- [x] Add a clear button (either in the dialogs or in the settings)
- [x] Consider making the saved views per-file
- [x] Add documentation / comments


gsantner:
When I "save a view", named hello, (consisting +family +home) -- I somehow never get to see it again in any dialog. Where should I see it?

I see theres also some new any / exact match toggle there, however it seems not to make any difference? I enter "blo" and blog comes up. But it shows up with both exact and any ... so whats the difference? xD

harshad1:
> When I "save a view", named hello, (consisting +family +home) -- I somehow never get to see it again in any dialog. Where should I see it?

Tap on the title. There you will see the new filter dialogs and any saved dialogs

PR #1473, opened 2021-10-20
Fix v2.8.1 saving issues

harshad1:
* Merge Document and DocumentIO classes, debloat
* flush() after write()
* less save/read calls in activitiy lifecycle
* ..


harshad1:
Issue is that the file is not written before it is read.

The text sent to disk in `onPause` is not the same as the text which is read in `onResume`. If you wait 10s, the process succeeds.

harshad1:
This issue is weirder than I thought.

1. Open file
2. Add text
3. Switch apps
4. Open file in other editor <- new text present
5. Open markor again <- new text gone

gsantner:
then maybe cached string in Document, saveinstancestate or textview?

gsantner:
I wrote DocumentIO stuff from scratch, including undo/redo history, file saving etc. Over time most stuff got replaced by other means (i.e. said undo/redo which worked better in the said textview xlass)

gsantner:
when you are on mainactivity, or editor?

on main its due each of quicknote and todo.txt own lifecycle

Issue #1482, opened 2021-11-04
v2.8.2: ShareInto/QuickNote Markor not saving


Ssmur:
## General information
* App version: 2.8.2 (F-Droid)
* Android version: 8.1
* Device: OnePlus 5T
* I understand that this issue will be closed without further notice in case I didn't describe the problem including general information, or didn't [search](https://github.com/gsantner/markor/issues?q=) for similar issues & [requests](https://github.com/gsantner/markor/discussions?discussions_q=).

### Description

Today I've updated Markor (2.8.1) to (2.8.2).
 In Firefox, press Share button to save URL, paste the note on a large file I have of links saved, and Markor when ask where to insert the note, I choose that file (bookmarksff.md) and open the file cmpletly erased everything, 0 bytes.  I have create other dummy file (markdown), write something random, Save, close file and open it again, there are the text. I repeat the process of save a Url from Firefox to Markor and choose that dummy file, and again, the file auto-delete its content, does not even enter the Url I want to save.
I wil try with a Xiaomi 6a, and update the result.

**[UPDATE]** Well, in a Xiaomi 6a  Stock rom MIUI 11.0.8 reproduces the same behaviour. In this  case, I open a app of F-Droid catalog, press ShareInto button, select Markor, select file of notes and got erased its  content.

gsantner:
pronably from #1473 , though using this a lot haven't seen it yet

harshad1:
Relevant code 

https://github.com/gsantner/markor/blob/master/app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java#L203-L220

I don't see an obvious bug in the code

gsantner:
I'm setting up android dev on a new OS. As far I read there is also an issue with QuickNote/todo.txt/Main page, so might be outside the ShareIntoFragment

harshad1:
On fail we return content = ""

https://github.com/gsantner/markor/blob/master/app/src/main/java/net/gsantner/markor/model/Document.java#L230-L231

I think it would be better to not hide the exception and force the calling code to handle an exception. That way, we can prevent writes if there was an exception on read.

harshad1:
I can reproduce this issue. Happens with encrypted files.

gsantner:
hehe, once more "we don't save empty files, too much tears about lost files" :smiley:

harshad1:
> hehe, once more "we don't save empty files, too much tears about lost files" 😃

Sigh.

[...]

PR #1487, opened 2021-11-05
WIP: More defensive file IO

harshad1:
This PR is meant to address #1482

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
@@ -39,6 +39,7 @@ import net.gsantner.opoc.format.plaintext.PlainTextStuff;
 import net.gsantner.opoc.preference.GsPreferenceFragmentCompat;
 import net.gsantner.opoc.ui.FilesystemViewerAdapter;
 import net.gsantner.opoc.ui.FilesystemViewerData;
+import net.gsantner.opoc.util.StringUtils;
 
 import java.io.File;
 import java.util.regex.Matcher;
@@ -207,12 +208,16 @@ public class DocumentShareIntoFragment extends GsFragmentBase {
             final Context context = getContext();
             final Document document = Document.fromArguments(context, args);
             String content = document.loadContent(context);
+
             if (content != null) { // Don't do anything if could not load file
                 content = content.trim();
-                document.saveContent(context, content + (content.isEmpty() ? "" : "\n") + separator + _sharedText);
+                content += (content.isEmpty() ? "" : "\n") + separator + _sharedText;
+                document.saveContent(context, content);
                 document.resetModTime(); // Force load of content in document
 
                 if (showEditor) {
+                    // Open at line
+                    document.setInitialLineNumber(StringUtils.getLineOffsetFromIndex(content, content.length() - 1)[0]);
                     showInDocumentActivity(document);
                 }
 
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
@@ -201,22 +201,24 @@ public class DocumentShareIntoFragment extends GsFragmentBase {
         }
 
         private void appendToExistingDocument(final File file, final String separator, final boolean showEditor) {
-            Bundle args = new Bundle();
+            final Bundle args = new Bundle();
             args.putSerializable(Document.EXTRA_PATH, file);
             args.putBoolean(Document.EXTRA_PATH_IS_FOLDER, false);
             final Context context = getContext();
             final Document document = Document.fromArguments(context, args);
-            String trimmedContent = document.loadContent(context).trim();
-            String currentContent = TextUtils.isEmpty(trimmedContent) ? "" : (trimmedContent + "\n");
-            document.saveContent(context, currentContent + separator + _sharedText);
-            document.resetModTime();
+            String content = document.loadContent(context);
+            if (content != null) { // Don't do anything if could not load file
+                content = content.trim();
+                document.saveContent(context, content + (content.isEmpty() ? "" : "\n") + separator + _sharedText);
+                document.resetModTime(); // Force load of content in document
 
-            if (showEditor) {
-                showInDocumentActivity(document);
-            }
+                if (showEditor) {
+                    showInDocumentActivity(document);
+                }
 
-            if (file != null) {
-                _appSettings.addRecentDocument(file);
+                if (file != null) {
+                    _appSettings.addRecentDocument(file);
+                }
             }
         }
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTextActions.java
```diff
@@ -192,21 +193,20 @@ public void onClick(View view) {
...
-                                final String tasksString = TodoTxtTask.tasksToString(keep);
-                                _hlEditor.setText(tasksString);
-                                _hlEditor.setSelection(
-                                        StringUtils.getIndexFromLineOffset(tasksString, selStart),
-                                        StringUtils.getIndexFromLineOffset(tasksString, selEnd)
-                                );
+                            final Document doneDoc = new Document(new File(_document.getFile().getParentFile(), callbackPayload));
+                            final Context context = getContext();
+                            String doneStr = doneDoc.loadContent(context);
+                            if (doneStr != null) {
```
harshad1, line 199: Don't do anything if the read file was null. i.e. error

app/src/main/java/net/gsantner/markor/model/Document.java
```diff
@@ -214,41 +212,46 @@ public static Document fromArguments(Context context, Bundle arguments) {
         return document;
     }
 
-    public synchronized String loadContent(final Context context) {
+    public String loadContent(final Context context) {
+        return loadContent(this, context);
+    }
+
+    public static synchronized String loadContent(final Document document, final Context context) {
```
harshad1, line 219: Wrap static synchronized calls - only write one document at a time.

app/src/main/java/net/gsantner/markor/model/Document.java
```diff
@@ -214,41 +212,46 @@ public static Document fromArguments(Context context, Bundle arguments) {
...
-                Log.e(Document.class.getName(), "loadDocument:  File " + getFile() + " not found.");
-                content = "";
+                Log.e(Document.class.getName(), "loadDocument:  File " + file + " not found.");
+                content = null;
             } catch (JavaPasswordbasedCryption.EncryptionFailedException | IllegalArgumentException e) {
                 Toast.makeText(context, R.string.could_not_decrypt_file_content_wrong_password_or_is_the_file_maybe_not_encrypted, Toast.LENGTH_LONG).show();
-                Log.e(Document.class.getName(), "loadDocument:  decrypt failed for File " + getFile() + ". " + e.getMessage(), e);
-                content = "";
+                Log.e(Document.class.getName(), "loadDocument:  decrypt failed for File " + file + ". " + e.getMessage(), e);
+                content = null;
```
harshad1, line 238: error = return null

app/src/main/java/net/gsantner/opoc/util/FileUtils.java
```diff
@@ -55,7 +55,8 @@ public static String readTextFileFast(final File file) {
         } catch (IOException e) {
             e.printStackTrace();
         }
-        return "";
+
+        return null;
```
harshad1, line 59: error = return null

app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
```diff
@@ -201,22 +202,28 @@ public void setText(String text) {
...
+
+            if (content != null) { // Don't do anything if could not load file
+                content = content.trim();
+                content += (content.isEmpty() ? "" : "\n") + separator + _sharedText;
+                document.saveContent(context, content);
+                document.resetModTime(); // Force load of content in document
+
+                if (showEditor) {
+                    // Open at line
+                    document.setInitialLineNumber(StringUtils.getLineOffsetFromIndex(content, content.length() - 1)[0]);
```
gsantner, line 220: maybe INT_MAX  works, so editor just jumps to the end of the file...without calculation

app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
```diff
@@ -201,22 +202,28 @@ public void setText(String text) {
...
+                    showInDocumentActivity(document);
+                }
 
-            if (file != null) {
-                _appSettings.addRecentDocument(file);
+                if (file != null) {
+                    _appSettings.addRecentDocument(file);
```
gsantner, line 225: should be in the editor class

app/src/main/java/net/gsantner/markor/model/Document.java
gsantner: maybe

```
catch (encryriptonexception) {
}
catch (exception) {
}
```

to catch all errors at this point. the error-or-not info is contained as null

gsantner:
do you think makes sense to release 2.8.3 with the current master, before merging this? I  heard it's pretty fine on master now. Not that we release another buggy version

harshad1:
> do you think makes sense to release 2.8.3 with the current master, before merging this? I heard it's pretty fine on master now. Not that we release another buggy version

Yes. I think we should.

I would like to test / refine these changes much more. This PR is probably only for 2.9 / 2.9.1.

gsantner:
https://github.com/gsantner/markor/releases/tag/v2.8.3

harshad1:
@gsantner So the whole modtime thing is there to prevent unnecessary reads and writes

The downside of using modtime is that some reads / writes may not happen when we need them to - this requires using resetModTime which is brittle, imo.

What are the downsides of just removing the whole modtime mechanism and allowing the multiple reads and writes (we will still have the hash mechanism to prevent duplicate writes inside `Document`)?

gsantner:
> What are the downsides of just removing the whole modtime mechanism and allowing the multiple reads and writes (we will still have the hash mechanism to prevent duplicate writes inside Document)?

Duplicate read = slow

gsantner:
> What are the downsides of just removing the whole modtime mechanism and allowing the multiple reads and writes

Duplicate read = slow. But if you think it's too much hassle - we can remove it again. Though we need to make sure we don't make File::read calls like 5 times when opening up one file (like it was at 2.8.0).

I know it might be tricky, but we actually need some kind of read Timer that is started, and can only be running one at a time at one fragment. On timeout read async, apply text in uithread and unlock the hlEditor (so at that point it is allowed to save from this editor, and gets editable by users). Beside double read by lifecycle - this also eliminates freezing when opening the activity if the file is big.

harshad1:
Instead of loading the file in the background, we can just show a dialog saying the file has been changed and asking the user if they want to reload. If they do, then we can do a blocking load. This is kind of how vim does it iirc.

I think with 2.8.2 we have 1 load or save at a time. I will investigate removing the modtime.

gsantner:
>Instead of loading the file in the background, we can just show a dialog saying the file has been changed and asking the user if they want to reload. If they do, then we can do a blocking load. This is kind of how vim does it iirc.

no, please not. dialogs are especially tedious on main sceen QuickNote/ToDo

gsantner:
>Instead of loading the file in the background, we can just show a dialog saying the file has been changed and asking the user if they want to reload. If they do, then we can do a blocking load. This is kind of how vim does it iirc.

preferably not. Dialogs are especially bad on main screen QuickNote/ToDo. There could be both changed at same time, so two dialogs popping at same time xD. BUT: Loading indicator / progress overlay in fragment?

> Is this an actual issue? I edit 50-100k character files all the time and I have no issues (but I am on a pixel 3 ... which is a fast phone)

Yes. The bigger the file is, or the worse the phone specs are. I can also notice it when phone is in ultra-battery-saving-mode.

Though, it's not forcefully in scope of PR to load async. I just wanted to share thoughts.

PR #1489, opened 2021-11-06
Basic defensive file io

harshad1:
More basic defensive file IO. 

1. Removed modtime
2. Added time based duplicate save prevention
3. Added no save empty file
4. Removed several duplicate reads / writes


app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -224,11 +234,6 @@ public void onResume() {
 
         loadDocument();
 
-        int cursor = _hlEditor.getSelectionStart();
-        cursor = Math.max(0, cursor);
-        cursor = Math.min(_hlEditor.length(), cursor);
-        _hlEditor.setSelection(cursor);
```
harshad1, line 230: `loadDocument` contains this same code

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -317,26 +322,13 @@ public boolean onQueryTextChange(String text) {
...
-            DocumentActivity da = ((DocumentActivity) activity);
-            da.setDocumentTitle(_document.getTitle());
-            da.setDocument(_document);
-        }
-
-        // Upon construction, the document format has been determined from extension etc
-        // Here we replace it with the last saved format.
-        _document.setFormat(_appSettings.getDocumentFormat(getPath(), _document.getFormat()));
-        applyTextFormat(_document.getFormat());
-        _textFormat.getTextActions().setDocument(_document);
```
harshad1, line 339: Moved to `onViewCreated` no need to call this on each onResume

gsantner:
replaces (can close) #1487 , or supplement?

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -662,10 +654,13 @@ public boolean onBackPressed() {
...
     }
 
+    public boolean saveDocument() {
+        return saveDocument(false);
+    }
 
     // Save the file
     // Only supports java.io.File. TODO: Android Content
-    public boolean saveDocument() {
+    public boolean saveDocument(boolean dontIgnoreEmpty) {
```
gsantner, line 663: Suggest passing on the flag named something with "manually" in it. So, it should be possible to save empty file (and possibly future other special cases), only with a manual user button press

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -662,10 +654,13 @@ public boolean onBackPressed() {
         return false;
     }
 
+    public boolean saveDocument() {
+        return saveDocument(false);
+    }
```
gsantner, line 659: suggestion: lets just have one, and always explicitly specify whether saving happens based on manual or automatic action

app/src/main/java/net/gsantner/markor/model/Document.java
```diff
@@ -52,7 +53,7 @@
     private final String _fileExtension;
     private int _format = TextFormat.FORMAT_UNKNOWN;
     private String _title = "";
-    private long _modTime = 0;  // Modtime as of when the file was last loaded / written
+    private long _lastSaveTime = -1;
```
gsantner, line 56: I suggest to keep 0: `(NEW-OLD) => NEW - (-1) => (NEW+1)`. Not that it will matter much, but with 0 highest the result can get is NEW (-> `==` / `<=` / `>=` are true)

harshad1:
> replaces (can close) #1487 , or supplement?

This is a much more basic PR which I think will be suitable to merge now / soon. I think we can close 1487. I will revisit those changes at a later date (may not be necessary)

gsantner:
OK! The `readerror -> return null;` of the other one is a good addition

harshad1:
Yeah. But imo we should make changes to document saving / loading in smaller increments.

I will add that change in a later PR

Nov. 7, 2021 08:39:48 Gregor Santner ***@***.***>:

harshad1:
This is ready for review / merging imo

gsantner:
merging on weekend

Issue #1368, opened 2021-06-20
Feature Request: Share Into Timestamp/Append-Prepend Options


smartinick:
Been using Markor for a while now, synching with other devices and sharing markdown files between dokuwiki and obsidian.

what i really like on markor is the share into function as some sort of log - when there's some interesting link to revisit later, i use share into a special file which is meant to be some sort of history/log of interesting links. adding tags for reminders,... this works quite well.

Currently the share into always prepends the shared text with a line containing "---"
-> could this be configurable?
i'd e.g. like to have date/time in a simple bullet list for each shared element:
 - [2021-06-20 10:16] shared-text #mytaglist
using this timeformat (or a setting that allows to configure it using some format-function) i'd also be able to sort the liste later on.

Personally i also add headings for each month, but that's easily done manually once a month.

Adding the date-time text manually each time is a bit of a hassle, therefore this feature request ;)

Thanks for considering!, Martin.

harshad1:
I like this idea and think I would like it too.

I'll start working on it.

Biggest challenge would be to make it clean and user-friendly.

Starting point is going to be todo.txt format, I think. If you share into your default todo file, you'll get a dete-stamp (only date though, unfortunately).

gsantner:
I suggest to make use of the filemanager description line format code

smartinick:
harshad1 - Thanks for your efforts!

From my perspective there could be a setting that has 2 parts:
- checkbox: use prefix for share-into
- text: the prefix text to use, e.g. in the current version this would be "---\r\n", and it could be something similar to this pseudo-code: "---\r\n - [\format('YYYY-MM-DD hh:mm',now()] "

Off course i have no idea of the internals of markor, therefore consider this nothing more than a educated guess/wish!

Thanks!!!

harshad1:
Note that CRLF will likely not work well anywhere in Markor. We rely on '\n' for a _lot_ of parsing code

pudah:
> Currently the share into always prepends the shared text with a line containing "---"

In the latest version 2.7.0, this functionality seems to have been removed. Was that done on purpose?

gsantner:
threshold is 400 characters ("long text"), for short text no separator added.

harshad1:
The biggest challenge here is to design a good, consistent UI / behavior. I am willing to write the code if someone else develops a plan.

gsantner:
[...]

PR #1495, opened 2021-11-12
Share into prefix

harshad1:
This PR addresses #1368.

1. Added a field in settings for the prefix
2. Prefix is added to any text shared into app
3. Prefix + shared text is shown in share into dialog so user can edit it as needed

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
@@ -218,7 +218,6 @@ public class DocumentShareIntoFragment extends GsFragmentBase {
             String trimmedContent = document.loadContent(context).trim();
             String currentContent = TextUtils.isEmpty(trimmedContent) ? "" : (trimmedContent + "\n");
             document.saveContent(context, currentContent + separator + _sharedText);
-            document.resetModTime();
 
             if (showEditor) {
                 showInDocumentActivity(document);
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
@@ -122,7 +122,7 @@ public class DocumentShareIntoFragment extends GsFragmentBase {
                 return result;
             }
         } catch (IllegalFormatException ignored) { }
-        return format + " " + value;
+        return format + (format.length() > 0 ? " " : "") + value;
     }
 
     @OnTextChanged(value = R.id.document__fragment__share_into__highlighting_editor, callback = OnTextChanged.Callback.TEXT_CHANGED)
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
@@ -252,7 +252,7 @@ public class DocumentShareIntoFragment extends GsFragmentBase {
 
                 @Override
                 public void onFsViewerSelected(String request, File file, final Integer lineNumber) {
-                    appendToExistingDocument(file, getSeparator(_sharedText), true);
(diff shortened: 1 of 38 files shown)


gsantner:
Suggestion:

* Naming - ShareInto Format
* Usage - With %s. So `\n---\n%s\n---` inserts in the middle

what do you think?

app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
```diff
@@ -99,17 +103,28 @@ public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
...
         _hlEditor.setTextSize(TypedValue.COMPLEX_UNIT_SP, as.getFontSize());
         _hlEditor.setTypeface(Typeface.create(as.getFontFamily(), Typeface.NORMAL));
 
-        if (sharedText.trim().isEmpty()) {
+        if (sharedText.isEmpty() || sharedText.equals(prefix)) {
             _hlEditor.requestFocus();
         }
     }
 
+    private static String formatOrPrefix(final String format, final String value) {
```
gsantner, line 118: simple:
```
format = format.contains("%s") ? format : (format + "%s");
result = format.replace("%s", value);
```

doesn't need real string formatting, just a token which is %s

app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
```diff
@@ -99,17 +103,28 @@ public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
...
+        try {
+            final String result  = String.format(format, value);
+            if (!result.equals(format) && !TextUtils.isEmpty(value)) {
+                return result;
+            }
+        } catch (IllegalFormatException ignored) { }
+        return format + " " + value;
```
gsantner, line 125:
```suggestion
String fmt = format.contains("%s") ? format : (format + "%s");
return fmt.replace("%s", TextUtils.isEmpty(value) ? "" : value);
```

app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
```diff
@@ -99,17 +103,28 @@ public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
...
+    }
```
gsantner, line 126:
```suggestion
    private static String formatOrPrefix(String format, final String value) {
        format = format.contains("%s") ? format : (format + "%s");
        return format.replace("%s", TextUtils.isEmpty(value) ? "" : value);
    }
```

harshad1: I thought of this, but felt that there were edge cases with escaped '\%s' etc. In the end It's only a couple lines extra

app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
```diff
@@ -116,13 +116,7 @@ public void onViewCreated(final @NonNull View view, final Bundle savedInstanceSt
...
 
     private static String formatOrPrefix(final String format, final String value) {
-        try {
-            final String result  = String.format(format, value);
-            if (!result.equals(format) && !TextUtils.isEmpty(value)) {
-                return result;
-            }
-        } catch (IllegalFormatException ignored) { }
-        return format + (format.length() > 0 ? " " : "") + value;
+        return (format + (format.contains("%s") ? "" : " %s")).replace("%s", value);
```
harshad1, line 119: Edge cases either way. Went with simpler solution

gsantner:
seems like this broke "shareinto - todo.txt". 

as it generally applies the prefix/format on the input field, instead of "at append time".  
and the todo.txt field is only shown when the input is singleline (which makes sense).  
thus, need to fix it so format is applied again AT time of appending, not at the time of creating the inputfield

PR #1510, opened 2021-11-27
Support spaces in image url #1365

### LINKED ISSUE #1365: Support [my image.jpg](<assets/my image.jpg>) syntax
Some file or image names may contains spaces. When inserting into Markor, Markor can handle them in this for now: 

1. Transcode the space then put it in, like `[my image.jpg](assets/my%20image.jpg)`

Obviously that way is not so tidy.  

However, with this problem, I looked into the [CommonMark Spec](https://spec.commonmark.org/0.29/#links) , surprisingly, I found that Markdown do support inserting links with spaces! 

> The destination can only contain spaces if it is enclosed in pointy brackets:
>
> [Example 485](https://spec.commonmark.org/0.29/#example-485)
>
> ```
> [link](/my uri)
> <p>[link](/my uri)</p>
> ```
>
> [Example 486](https://spec.commonmark.org/0.29/#example-486)
>
> ```
> [link](</my uri>)
> <p><a href="/my%20uri">link</a></p>
> ```

In terms of the example I mentioned ahead, it would become 

```
[my image.jpg](<assets/my image.jpg>)
```

Which is really tidy and standard. 

But it seems Markor don't support that syntax, right? Will it be cooperated into Markor?

tifish:
Fix #1365


gsantner:
Can you check if its possible to use a stringbuffer at the loop instead of concatting the whole time? something like


```
stringbuffer sb
int indexLastContent=0
while(){
  sb.append(original.subsbtring(indexLastContent, beforeMatch))

String updatedLink = ...%20....

sb.append(updatedLink);
indexLastContent = matcher.indexAfterMatch
}
// one final append for rest of string
sb.append(..
```

tifish:
I have optimized the code with StringBuilder. Another bug was fixed: cannot parse link with title.

gsantner:
@tifish  
Thanks, think performance shuold be way better this way.

Do you think it's easily possible to just have a regex that detects the space characters in such a URL and automatically replace it with normal regex String.replaceAll? Then all it needs is one string replace and the proper regex.

tifish:
We need to replace a capturing group with uncertain number of spaces. As far as I know, the regular expression cannot do it.

gsantner (approved):
I did a few improvements and added documentation to the code.

Tried it and looked OK to me 😄 .

Thank you, I'm merging this now.

tifish:
@gsantner Thank you :). I am looking forward the new version. Absolutely I want to add more features if I got time.

Issue #1174, opened 2020-12-28
Privacy feature: Disallow screenshots (Secure flag)


gsantner:
#### General information

* **App version:**  2.5
* **System:** Android 11

#### Description
* The goal is to add feature to app to disallow screenshots of the running Markor app (so other apps cannot screenshot or see what you write).
* Should be a settings option to enable/disable globally for all activities in the Markor app
* Make it enabled by default
* Optionally, if you find some other security/privacy improvements&tweaks, can also be added (comment if you have something)

Check documentation for FLAG_SECURE and related Android Developers security/privacy guides.  
Overall it should be quite easy to implement this, settings are more or less copy paste from existing checkboxes 😉 .

Ln11211:
@gsantner, As it reads good first issue and you said it's more or less copy paste, canI try this ?

gsantner:
Yes it is one of the easiest.

detherminal:
Hello. I want to make this. How can I give you the answer

gsantner:
Merge/Pull request.

Ln11211:
@gsantner , I found a nice resource and it says that you can use a built in [mechanism](https://medium.com/nomtek/screenshot-preventing-on-mobile-apps-9e62f51643e9) to prevent screenshots. Check out the link and let me know if we can use this to implement it.

I also found one in [stackoverflow](https://stackoverflow.com/questions/9822076/how-do-i-prevent-android-taking-a-screenshot-when-my-app-goes-to-the-background) that prevents taking screenshot when an app goes to background.. 

But, this one from geeksforgeeks seems good and simple, [geeksforgeeks](https://www.geeksforgeeks.org/how-to-prevent-screenshot-or-screen-recorder-in-android/) . 

I would like to go with the code by geeksforgeeks...

gsantner:
yes java FLAG_SECURE is what you need.

Ln11211:
@gsantner , I don't think I can do this issue as I lack the knowledge of working with android code, I'm sorry for wasting your time.

gsantner:
No worries.

By the way, it's never too late to get started :-)

gsantner:
do it

gsantner:
@Zoo-M0    
It happens every year in october (/hacktoberfest) that many people want to start on many things, but mostly nothing happens.  
Unless there is merge request open, generally consider all issues/bugs/features open. Currently there are 0 open ... so everything 😄 .

Zoo-M0:
[...]

PR #1514, opened 2021-11-30
#1174 disallow screenshots

Zoo-M0:
Add FLAG_SECURE to activities to avoid screenshots.


gsantner:
@Zoo-M0  
I made a change and now all activites inherit from MarkorBaseActivity.  
So you can now make the change there instead.  
Note: It should be a additional settings option (preferences xml file), not "just add the flag"

Zoo-M0:
@gsantner perfect, I will move it to the new activity, and work on the preferences flag.

gsantner:
I reworked it a bit, merging now 

thanks! 😄

gsantner (approved):
I reworked it a bit, merging now 

thanks for your help 😄 

as first-time-contribution  - happy to hear more! 🎉

Issue #1311, opened 2021-04-10
File modified indicator 


tomtom:
#### General information

* **App version:** 2.6.0
* **System:** Chrome OS, Android R

#### Description

There currently is no indicator whether a file was modified or not. Most editors provide some marker to tell users whether the file is "dirty" and needs to be saved. It would be nice if Markor did this too. Some editors append a star/asterisk (*) to the filename (displayed in the tab) in order to indicate the file status.

Issue #1303, opened 2021-04-03
[feature request] option to disable auto save


JunkyardCat:
Hi is it possible to add an option to manually save files?. The default settings will still be auto save so that other people who likes auto save dont need to change anything

gsantner:
There is an button to manually save at any time. disable auto save is not planned currently, too many tears because people kill the app and forget to save.

JunkyardCat:
@gsantner 
Im not saying to disable auto save by default. Just have the option to disable auto save for users like me.

What this means is that auto save will still be here but can be disabled by going into option

harshad1:
Note that many actions operate on the file (file information, for example) and not the text view contents. These implicitly save the file.

JunkyardCat:
@harshad1 
I see but im fine looking at outdated info if ever i will look up file information while editing before saving the file

JunkyardCat:
i checked the thread you mentioned @dnngll but i have to disagree. i just want an option to switch to normal mode i ma not saying to remove auto save i just want an option to choose to manual save.

so what this means is that all user will be defaulted to have auto save

dnngll:
Hello @JunkyardCat,

thanks for your input. I don't quite get your point though: We want different things, and one doesn't negate the other (however, @gsantner already stated that he's not planning to implement #1303). Or did I misunderstand?

In https://github.com/gsantner/markor/issues/1419#issuecomment-904016195 I was just referring to @gsantner's comment because I thought, that https://github.com/gsantner/markor/issues/1419#issuecomment-904013447 wasn't valid.

gsantner:
hello, there is already #1419 about this topic. Please have this discussion there.

And let's see what future brings.

gsantner:
One pretty simple thought: What if you get a button at the top toolbar - `Close file without saving`. Clicking it closes the file and the original file remains as is. No settings option or alike.

Manual save button is available as well, since quite some time now.

Would this solve the problem?

JunkyardCat:
@gsantner 

Can i ask what is it that wont work out? can you give more concrete example? is it the viewing real time of markdown file in preview mode? is it the auto parsing of todo?

For me I think itll work out. On top of my head i think the one thing thatll not work as expected is the preview of the markdown when clicking the preview button since its not yet saved and also i guess the todo wont be able to auto apply its thing(i dont really know whats it called. its the one where you put in carriage return and it automatically parses the entry before accordingly).
i can live without those things.

tengucrow:
[...]

PR #1516, opened 2021-12-02
File change state indication & general improvements

harshad1:
This PR addresses #1311 

~~`*` is added to the end of the filename if the file is currently dirty~~ Save icon is used to indicate if the file has changed

TODO:
- [x] Add `*` in `MainActivity`
- [x] Fix title alignment


app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -673,21 +685,22 @@ public boolean onBackPressed() {
     // Save the file
     // Only supports java.io.File. TODO: Android Content
     public boolean saveDocument(boolean forceSaveEmpty) {
-        if (isAdded() && _hlEditor != null && _hlEditor.getText() != null) {
+        if (_isTextChanged && isAdded() && _hlEditor != null && _hlEditor.getText() != null) {
```
harshad1, line 688: Only writing if `_isTextChanged == true`. Should address several cases from #1303 etc

app/src/main/java/net/gsantner/markor/model/Document.java
```diff
@@ -192,6 +195,15 @@ public static Document fromArguments(Context context, Bundle arguments) {
         return document;
     }
 
+    private void setHash(final String s) {
+        _lastLength = s.length() ;
+        _lastHash = FileUtils.crc32(s.getBytes());
+    }
+
+    public boolean isDataSame(final String s) {
+        return s.length() == _lastLength && FileUtils.crc32(s.getBytes()) == _lastHash;
```
harshad1, line 204: Using a fast length comparison first

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
```diff
@@ -403,7 +397,9 @@ public void onFsViewerConfig(FilesystemViewerData.Options dopt) {
                 public void onFsViewerDoUiUpdate(FilesystemViewerAdapter adapter) {
                     if (adapter != null && adapter.getCurrentFolder() != null && !TextUtils.isEmpty(adapter.getCurrentFolder().getName())) {
                         _appSettings.setFileBrowserLastBrowsedFolder(adapter.getCurrentFolder());
-                        _toolbar.setTitle(adapter.areItemsSelected() ? "" : getFileBrowserTitle());
+                        if (_viewPager.getCurrentItem() == 0) {
+                            _toolbar.setTitle(adapter.areItemsSelected() ? "" : getFileBrowserTitle());
+                        }
```
harshad1, line 402: Prevent title from being set unless we are in the file browser. Otherwise the title for startup tab is always 'Markor' even when, say, quicknote is used as startup

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
```diff
@@ -363,17 +333,41 @@ public String getFileBrowserTitle() {
...
 
-        if (pos > 0 && pos < 3) {
-            permc.doIfExtStoragePermissionGranted(); // cannot prevent bottom tab selection
+        if (pos != tabIdToPos(R.id.nav_notebook)) {
+            restoreDefaultToolbar();
+        }
+
+        if (pos == tabIdToPos(R.id.nav_quicknote) || pos == tabIdToPos(R.id.nav_todo)) {
+            // cannot prevent bottom tab selection
+            new PermissionChecker(this).doIfExtStoragePermissionGranted();
```
harshad1, line 370: Cleanups. These seem to work fine. Will report after extensive testing.

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
```diff
@@ -312,38 +308,12 @@ public void onBackPressed() {
...
-            case R.id.nav_notebook: {
-                _viewPager.setCurrentItem(0);
-                _toolbar.setTitle(getFileBrowserTitle());
-                return true;
-            }
+        _viewPager.setCurrentItem(tabIdToPos(item.getItemId()));
+        return true;
+    }
 
-            case R.id.nav_todo: {
```
harshad1, line 325: All of this is redundant with `onViewPagerPageSelected`. Removed

gsantner:
did you notice any craah/issue on the latest update?

harshad1:
> did you notice any craah/issue on the latest update?

No, but I am using this branch

app/src/main/java/net/gsantner/markor/model/Document.java
```diff
@@ -192,6 +194,14 @@ public static Document fromArguments(Context context, Bundle arguments) {
         return document;
     }
 
+    private void setContent(final String s) {
+        _lastContent = s;
```
harshad1, line 198: Just storing the content for comparison instead of hash. Simpler and faster. We can try something fancy like storing the hash if content is very large. But in practice, I don't think this will be a problem.

gsantner: what was the reason why we scraped the previous one where we already used & stored hash? just that it's crc32 instead of md5 now

harshad1: I thought about it and I don't think using a hash is any better than just storing the content. 

Options are:
1. crc32 - fast, low chance of collision
2. sha512 - slower, very low change of collision
3. store full text - fastest, 0 chance of collision. But potential extra memory consumption.

Until recently we stored multiple versions of the full content with no issues, so memory isn't a concern in practice. I don't see any downsides to just storing and comparing the entire content.

gsantner:
I've observed that at first app start, after granting the storage/file permission, the todo & quicknote tabs stay empty. restarting the app loads them properly.

also when removing permission via system settings, swiping to quicknote, new permission dialog, accept, and file stays empty.

is that fixed in this PR maybe already?

app/src/main/java/net/gsantner/markor/model/Document.java
gsantner: Generally don't think collision isn't something to think about for Markor's use (i.e. if somebody intentionally crafts a md5/sha1 collision) --> not that important, for any of crc/md5/sha1/shaXXX it will give different value if you just type one different character somewhere.

Though have to admit,  we consume quite excess amount of RAM (especially in highlighters), wouldn't wonder if we have like 5 full copies of text floating around somewhere 😄 . By increasing size of file it gets an issue on low-end cheap devices (running into OOM).

Plus: We can't store a copy of the original text in bundle/activity (1MB limit, where ½-¾ of it is already fixed consumed) - regarding device rotations and stuff. Storing some 64char long hash isn't a problem.

**It's not in scope to support it now & in this PR. You can ignore it, but just for your information how my view is on it**

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
gsantner: What do you think if I add a recurring timer that queries the title lets say all 2 seconds - on all activities?  
Setting title everywhere seems a bit hacky. Maybe just triggering a getActivityTitle in all fragments automatically that just return the current expected string solves the issue?

If you think it makes sense I can work on this on separate PR

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
gsantner: oh, is this maybe related/fix to my separate "observed" comment?

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
gsantner: What do you think if I add a recurring timer that queries the title lets say every second - on all activities?  
Title is set from everywhere, and everywhere a bit of different hacky code 😄 . Maybe just triggering a getActivityTitle in all fragments automatically that just return the current expected string solves the issue?

If you think it makes sense I can work on a concept of this on separate PR

app/src/main/java/net/gsantner/markor/model/Document.java
harshad1: > Plus: We can't store a copy of the original text in bundle/activity

Ah. I forgot about this. I will revert to using the hash.

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
harshad1: I don't think this will really solve the issue. Will make it harder to debug though :).

IMO the only solution is to track down all these cases and clean them up + add proper defensive logic

gsantner:
nevermind the todo/quicknote tab permission comment. Fixed it on master, and it works for both current SDK29 and the upcoming SDK31 (MANAGE_EXTERNAL_STORAGE) usage #1521

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
harshad1: To be clear, none of these changes change the functionality at all. I am just reorganizing the existing code so that things are called once, rather than 5 times. Equivalent already existed in this function in master.

gsantner: Do you like the navigation "4 bottom tabs" how it is? Or would you say, ok top tabs, or something else might fit better?

And yeah, the needed handling is really awful, there are like 5 different ways to get to one of the tabs, or 5 different callbacks when you reached one. :smiley:

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -328,7 +325,36 @@ public boolean onQueryTextChange(String text) {
...
+        if (_undoMenuItem != null && _lastCanUndo != canUndo) {
+            _undoMenuItem.setEnabled(canUndo).getIcon().mutate().setAlpha(canUndo ? 255 : 40);
+            _lastCanUndo = canUndo;
+        }
+
+        final boolean canRedo = _editTextUndoRedoHelper.getCanRedo();
+        if (_redoMenuItem != null && _lastCanRedo != canRedo) {
+            _redoMenuItem.setEnabled(canRedo).getIcon().mutate().setAlpha(canRedo ? 255 : 40);
+            _lastCanRedo = canUndo;
+        }
```
harshad1, line 357: Breaking the undo / redo menu updates out so that we don't call redraw menu _every time_

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -551,29 +578,25 @@ public void onFsViewerConfig(FilesystemViewerData.Options dopt) {
...
+        checkTextChangeState();
+        updateUndoRedoIconStates();
+    }
 
+    public void checkTextChangeState() {
+        if (_isTextChanged == _document.isContentSame(_hlEditor.getText())) {
+            _isTextChanged = !_isTextChanged;
+
+            if (_saveMenuItem != null) {
+                _saveMenuItem.setEnabled(_isTextChanged).getIcon().mutate().setAlpha(_isTextChanged ? 255 : 40);
```
harshad1, line 592: New approach - just enable / disable save icon and grey it out to show save state. Much cleaner than adding a *

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -551,29 +578,25 @@ public void onFsViewerConfig(FilesystemViewerData.Options dopt) {
...
-                    ((AppCompatActivity) activity).supportInvalidateOptionsMenu();
-                }
-            }, HISTORY_DELTA);
-        }
-        Activity activity = getActivity();
-        if (activity != null && activity instanceof AppCompatActivity) {
-            ((AppCompatActivity) activity).supportInvalidateOptionsMenu();
-        }
+        checkTextChangeState();
+        updateUndoRedoIconStates();
```
harshad1, line 584: As our tests and updates are fast (don't need to redraw whole menu etc). No need to use a HISTORY_DELTA delay. Can just call functions directly.

Investigating impact on performance.

gsantner: I think off adding a generic debouncer timer class to Callback.java. It's a common thing that there is a recurring pressure on some things, which inside makes even more pressure (graphics drawing).

you can keep it here without optimizations / stripping code

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
gsantner: I personally prefer this way of showing. Not sure it user is happy with it. But less bloat = better, hope they can agree.

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
harshad1: In general I think it would be useful, yes. `debounce(delay, callback.b0)` or something

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
harshad1: I like the tabs at the bottom. It feels clean and I can both swipe to change as well as click.

I am _really_ bad at UX design though :p. I'll need to think really hard about it and look at how others have solved similar issues before having a strong opinion.

app/src/main/java/net/gsantner/opoc/util/FileUtils.java
```diff
@@ -504,13 +504,16 @@ public static File join(File file, String... childSegments) {
...
-                sb.append(Integer.toString((b & 0xff) + 0x100, 16).substring(1));
+            final MessageDigest alg = MessageDigest.getInstance("SHA-512");
+            final int length = text.length();
+            for (int i = 0; i < length; i++) {
+                final char c = text.charAt(i);
+                alg.update((byte) (c >> 8)); // high part
+                alg.update((byte) (c & 0xFF)); // low part
             }
-            return sb.toString();
+            return new String(alg.digest());
```
harshad1, line 516: I updated this function to use a CharSequence directly. Should prevent unnecessary copies and allocations etc

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
```diff
@@ -136,11 +136,7 @@ protected void onCreate(Bundle savedInstanceState) {
 
         (new ActivityUtils(this)).applySpecialLaunchersVisibility(_appSettings.isSpecialFileLaunchersEnabled());
 
-        _bottomNav.postDelayed(() -> {
-            if (_appSettings.getAppStartupTab() != R.id.nav_notebook) {
-                _bottomNav.setSelectedItemId(_appSettings.getAppStartupTab());
-            }
-        }, 1);
+        _bottomNav.postDelayed(() -> _bottomNav.setSelectedItemId(_appSettings.getAppStartupTab()), 10);
```
harshad1, line 139: Ideally, we would trigger this in onResume of bottomNav or something

harshad1:
image

Issues remaining are 
1. Left align the title (there is a lot of space between the back arrow and the title)
2. Reduce the spacing between the icons.

gsantner:
I thought about removing the title generally, or generally making changes to the toolbar / new layout.  
it should be OK / you don't need to waste time on it now

harshad1:
> I thought about removing the title generally

I don't think this would be a good idea. Besides, we use the title as another button for heading search / todo filter so it is not wasted space

> it should be OK / you don't need to waste time on it now

I am still testing. Will see if I can get it done in a day or so. If I can't we can leave it as is

app/src/main/java/net/gsantner/opoc/util/FileUtils.java
```diff
@@ -504,13 +505,9 @@ public static File join(File file, String... childSegments) {
...
 
-    public static String sha512sum(final byte[] bytes) {
+    public static String sha512sum(final byte[] data) {
         try {
-            final StringBuilder sb = new StringBuilder();
-            for (final byte b : MessageDigest.getInstance("SHA-512").digest(bytes)) {
-                sb.append(Integer.toString((b & 0xff) + 0x100, 16).substring(1));
-            }
-            return sb.toString();
+            return Arrays.toString(MessageDigest.getInstance("SHA-512").digest(data));
```
harshad1, line 510: Idiomatic and faster

app/src/main/res/values-v21/styles.xml
```diff
@@ -8,4 +9,12 @@
         <item name="android:windowSharedElementExitTransition">@android:transition/move</item>
         <item name="android:windowActionModeOverlay">true</item>
     </style>
+
+    <!-- Techically valid for v17 -->
+    <style name="ActionButtonStyle" parent="Widget.AppCompat.ActionButton">
+        <item name="android:minWidth">0dip</item>
+        <item name="android:paddingStart">8dip</item>
+        <item name="android:paddingEnd">8dip</item>
+    </style>
```
harshad1, line 18: Reduce icon spacing

app/src/main/java/net/gsantner/markor/model/Document.java
```diff
@@ -99,6 +105,16 @@ public File getFile() {
         return _file;
     }
 
+    public long lastModified() {
+        try {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                return Files.readAttributes(_file.toPath(), BasicFileAttributes.class).lastModifiedTime().toMillis();
```
harshad1, line 111: This method seems much more consistent than `lastModified`

gsantner: is there also such a method in SAF available?

app/src/main/res/values-v21/styles.xml
gsantner: you can move it to just styles.xml. It will probably tell you about Start/End not being available but you can try with silence it with tools ignore.

app/src/main/java/net/gsantner/opoc/util/FileUtils.java
gsantner: do we also use it? but yeah shorter and better

md5 is just fine as well - speed is way more interesting here than "maximum collision safety"

harshad1: switched back to crc32. It's the fastest and 1 in 2^32 collision chance is good enough.

harshad1:
This is ready for a final look through and merge imo

app/src/main/java/net/gsantner/opoc/util/FileUtils.java
gsantner: OK, hope that's OK for you, I think it's the better option.

btw the hash function takes a alg parameter, but currently uses hardcoded SHA-512

```suggestion
            return Arrays.toString(MessageDigest.getInstance(alg).digest(data));
```

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -135,6 +134,11 @@ public static DocumentEditFragment newInstance(File path, boolean pathIsFolder,
     private MarkorWebViewClient _webViewClient;
     private boolean _nextConvertToPrintMode = false;
     private long _loadModTime = 0;
+    private boolean _isTextChanged = true; // true forces first update
+    private Activity _activity;
+    private MenuItem _saveMenuItem;
+    private MenuItem _undoMenuItem;
+    private MenuItem _redoMenuItem;
```
gsantner, line 141: I had similar needs in the past - to access menu items any time. For that I've added at some point an accessor at GsFragmentBase

grafik

this one can be used

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -135,6 +134,11 @@ public static DocumentEditFragment newInstance(File path, boolean pathIsFolder,
     private MarkorWebViewClient _webViewClient;
     private boolean _nextConvertToPrintMode = false;
     private long _loadModTime = 0;
+    private boolean _isTextChanged = true; // true forces first update
+    private Activity _activity;
```
gsantner, line 138: As much as I hate calling `getActivity()` instead of simply accessing it directly via some var .. I think it's better than storing it as hard member reference. On device rotation there might be also a point where the member reference is outdated, but `getAcitvity()` would already return the right thing. (And activity leak)

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
gsantner: see the GsFragmentBase comment

gsantner:
I could finally get rid of the unused continuous integration job

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
harshad1: I mean, the same logic applies to _webView, _shareUtil, _appSettings etc right? We are assigning `_activity` in `onViewCreated`, which should properly happen _after_ rotation etc.

I can revert it. But if we do, imo we should think about what else to revert to direct access

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
harshad1: I thought of this, but we check view state _every time text is changed_ and calling `findViewById` every time might be a performance issue.

harshad1:
This is ready for a final look through and merge imo

> I could finally get rid of the unused continuous integration job

:100:

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
gsantner: this is correct, but the utilities don't have a way to access activity otherwise than passing it somewhere. at fragments theres the builtin `getActivity` which always returns the right thing 

do you see an issue with it like getActivity being much slower than storing and using member?

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
gsantner: I'd say it's not performance relevant (unlike syntax highlighter). If you want to keep it this way with storing it, I will check if I can slim it a bit still

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
harshad1: > do you see an issue with it like getActivity being much slower than storing and using member?

There is no performance impact, only asthethic. getActivity just returns a member from the base class.

I'll just revert it - reduces the complexity in this PR.

gsantner:
With addition of the additional save button the toolbar is fully filled up. It is always hard to make people understand what  the usual "up as home" button does and means, so I would anyway want to remove it `(DocumentActivity::toolbar::setUpAsHomeEnabled(false)`.

This is how it looks with and without that button

harshad1:
Yeah. All that button does is trigger back. We can definately try it. I'll make the relevant changes.

Dec. 11, 2021 02:56:45 Gregor Santner ***@***.***>:

app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
```diff
@@ -161,11 +161,7 @@ protected void onCreate(Bundle savedInstanceState) {
         }
 
         setSupportActionBar(_toolbar);
-        ActionBar ab = getSupportActionBar();
-        if (ab != null) {
-            ab.setDisplayHomeAsUpEnabled(true);
-            ab.setDisplayShowTitleEnabled(false);
```
harshad1, line 167: No back button in DocumentActivity

app/src/main/java/net/gsantner/markor/format/TextConverter.java
```diff
@@ -84,7 +84,7 @@ public String convertMarkupShowInWebView(Document document, String content, WebV
         }
 
         String baseFolder = new AppSettings(context).getNotebookDirectoryAsStr();
-        if (document.getFile() != null && document.getFile().getParentFile() != null) {
```
harshad1, line 87: document requires file != null. If file == null, construction of document will fail immediately. Therefore these checks are redundant and misleading. Removed document file null checks

gsantner:
do you think about more, or should I merge it now?

app/src/main/java/net/gsantner/markor/model/Document.java
harshad1: I don't see it. But `DocumentFile.lastModified()` may be more reliable?

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -342,12 +358,21 @@ public void loadDocument() {
...
             final CharSequence text = _hlEditor.getText();
             if (text == null || !content.contentEquals(text)) {
+
+                final int[] sel = StringUtils.getSelection(_hlEditor);
+                sel[0] = Math.min(sel[0], content.length());
+                sel[1] = Math.min(sel[1], content.length());
+
                 _hlEditor.setText(content);
+
+                _hlEditor.setSelection(sel[0], sel[1]); // hleditor can handle invalid selections
```
harshad1, line 368: Maintaining selection

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -653,17 +666,7 @@ public String getFragmentTag() {
...
-                getActivity().getIntent().getBooleanExtra(DocumentActivity.EXTRA_DO_PREVIEW, false)
-                        || _appSettings.getDocumentPreviewState(getPath())
-                        || _document.getFile().getName().startsWith("index."));
-        if (_isPreviewVisible && !preview) {
-            setDocumentViewVisibility(false);
-            return true;
-        } else if (!_isPreviewVisible && preview) {
-            setDocumentViewVisibility(true);
-            return true;
-        } else if (_menuSearchViewForViewMode != null && !_menuSearchViewForViewMode.isIconified()) {
```
harshad1, line 666: Now that home button is gone, back button directly exits even when in preview mode

Did this so that a user can directly exit from preview state. Without this change user is forced to go through edit more making remembering preview state useless.

harshad1:
@gsantner I am still testing + updating workflows without the back button and other edge cases. Lets wait a few days before merging,

gsantner:
OK ready? Or should I wait with testing again?

gsantner:
OK ready, or should I wait with testing?

harshad1:
Ready imo. I have been using this without issues for the last couple days.


Dec. 18, 2021 11:48:18 Gregor Santner ***@***.***>:

gsantner:
OK then I think about merging and releasing an update somewhen in holidays

harshad1:
May be better to merge soon to give it a gap between the merge and the release. That will give us a few more days to test...

My concern is that some parts may not have been tested fully as my normal workflow does not use them.

Dec. 20, 2021 14:28:50 Gregor Santner ***@***.***>:

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -281,16 +282,12 @@ public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
...
 
         // Undo / Redo / Save (keep visible, but deactivated and tinted grey if not executable)
-        Drawable drawable;
-        drawable = menu.findItem(R.id.action_undo).setEnabled(canUndo).setVisible(!_isPreviewVisible).getIcon();
-        drawable.mutate().setAlpha(canUndo ? 255 : 40);
-        drawable = menu.findItem(R.id.action_redo).setEnabled(canRedo).setVisible(!_isPreviewVisible).getIcon();
-        drawable.mutate().setAlpha(canRedo ? 255 : 40);
+        _undoMenuItem = menu.findItem(R.id.action_undo).setVisible(!_isPreviewVisible);
+        _redoMenuItem = menu.findItem(R.id.action_redo).setVisible(!_isPreviewVisible);
+        _saveMenuItem = menu.findItem(R.id.action_save).setVisible(!_isPreviewVisible);
```
gsantner, line 290: So it is hidden now as I see it, but later below there is still the code `setEnabled => AlphaTint`

image

Can I remove tinting there then?

app/src/main/java/net/gsantner/markor/model/Document.java
gsantner: I'd not further delay this PR. But for future I'd like try it and use it if it works 😄

gsantner (commented):
Happy holidays & christmas

app/src/main/java/net/gsantner/markor/model/Document.java
gsantner: I'd not further delay this PR. But for future I'd like try it and use it if it works smile

PR #1535, opened 2021-12-21
filebrowser: Change notebook folder menu icon to home

harshad1:


harshad1:
@gsantner This change makes sense to me. Please merge if you agree

gsantner:
My personal opinion is that is's fine to have all "Go to folder X" at same point, and adding more buttons/options isn't that much required.

Unless you have a strong preference, we can ask the people in the discussion what they prefere (thumbs up/down)

gsantner:
My personal opinion is that is's fine to have all "Go to folder X" at same point, and adding more buttons/options isn't that much required.

Unless you have a strong preference, we can ask the people in the discussion what they prefer (thumbs up split out/down keep it grouped)

I leave this here for reference how it looks and what still fits depending on amount of things in top toolbar.

image

harshad1:
Note that the change in this PR only affects the file browser view in main activity

i.e. 
image

harshad1:
@gsantner do you have any further opinions about this PR?

gsantner:
To be honest, I don't really want that in. The toolbar is getting bloated and with each new item, less space for the title

harshad1:
I've reverted the icon position. Only changed the icon

Issue #1517, opened 2021-12-02
Android Launcher-specific white flash when starting Markor (Markor icon in external splashsceen)


garret:
## General information
* App version: 2.8.3 (from F-Droid)
* Android version: 12
* Device: Pixel 6

### Description
Screen flashes to white (even though the app is set in dark mode). This occurs especially when opening the app for first time and the markor logo just flashes in a white background. This is a bit annoying,especially when opening the app at night. I think that it is a solvable fix as also recently the app Lemmur had the same issue and was fixed in the latest update. 

References: 
- https://github.com/krawieck/lemmur/releases/tag/v0.7.0
- https://github.com/krawieck/lemmur/pull/257
- https://github.com/krawieck/lemmur/pull/256

gsantner:
search issues.

garret:
@gsantner what do you mean with "search issues, use the existing one"?

gsantner:
The template for issues contains this, it's there for a reason:

```
## General information
* App version: 
* Android version:  
* Device: 
* I understand that this issue will be closed without further notice in case I didn't describe the problem including general information, or didn't [search](https://github.com/gsantner/markor/issues?q=) for similar issues & [requests](https://github.com/gsantner/markor/discussions?discussions_q=).
```

garret:
@gsantner thanks for explaining. Could you please point me to the issue that would be an exact duplicate of this? I searched for "white" in the issues box and could only find another issue that seems similar (I even commented there) https://github.com/gsantner/markor/issues/568 but it is not the same. For this reason, I opened a new one. But maybe I am not good in searching as you write :)

gsantner:
568 is the same cause

garret:
Thank you very much for solving this! I had told you it could be fixed independently from the android launcher! :)

PR #1543, opened 2021-12-25
Day / night colors

harshad1:
This PR attempts to address #1517 

In this PR the light and dark themes have been unified into Day and Night themes as specified [here](https://developer.android.com/guide/topics/ui/look-and-feel/darktheme)

This will allow the correct `windowBackground` to be selected for the splash  as well as simplify a _lot_ of code. 

The downside is that I will have to reorganize Markor's colors and this will be challenging to do correctly.


app/src/main/res/values/colors.xml
```diff
@@ -24,22 +24,15 @@
...
 
-    <color name="light__background">#eeeeee</color>
-    <color name="light__background_2">#f9f9f9</color>
-    <color name="light__primary_text">#212121</color>
-    <color name="light__secondary_text">#757575</color>
-
-    <color name="dark__background">#222222</color>
-    <color name="dark__background_2">#424242</color>
-    <color name="dark__primary_text">@color/white</color>
-    <color name="dark__secondary_text">#9e9e9e</color>
```
gsantner, line 42: You can keep all the color code here in original

app/src/main/res/values-night/colors.xml
```diff
@@ -0,0 +1,14 @@
...
+
+    <color name="colorPrimary">@color/primary_dark</color>
+    <color name="textActionBarColor">@color/lighter_grey</color>
+
+    <color name="background">#222222</color>
+    <color name="background_2">#424242</color>
+    <color name="primary_text">@color/white</color>
+    <color name="secondary_text">#9e9e9e</color>
+
+    <color name="primary__greyish">#282828</color>
```
gsantner, line 12: and reference them  here by `@color/name`

harshad1: Hmmm, yeah, this makes sense, thanks

app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
```diff
@@ -136,12 +136,14 @@ public static void askUserIfWantsToOpenFileInThisApp(final Activity activity, fi
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
+        _appSettings = new AppSettings(this);
+        _appSettings.applyAppTheme();
```
gsantner, line 140: MarkorBaseActivity.java ?

app/src/main/java/net/gsantner/markor/App.java
```diff
@@ -10,27 +10,44 @@
...
     @Override
     public void onCreate() {
         super.onCreate();
         _app = this;
+        res =  getResources();
+        instantiateFakeWebviewColorFix();
+    }
+
+    private void instantiateFakeWebviewColorFix() {
+        // Per https://stackoverflow.com/a/54191884/4717438
```
harshad1, line 47: I don't know how anyone would do this without Stack Overflow

app/src/main/java/net/gsantner/markor/ui/hleditor/Highlighter.java
```diff
@@ -168,79 +168,24 @@ protected boolean isFirstHighlighting() {
...
-     * Create Span for isMatching in paragraph's. Note that this will highlight the full matched pattern
-     * (including optionals) if no group parameters are given.
-     *
-     * @param spannable     Text spannable
-     * @param pattern       The pattern to match
-     * @param creator       A ParcelableSpanCreator for ParcelableSpan
-     * @param groupsToMatch (optional) groups to be matched, indexes start at 1.
-     */
-    protected static void createSpanForMatchesP(final Spannable spannable, final Pattern pattern, final SpanCreator.ParagraphStyleCreator creator, int... groupsToMatch) {
+    protected static <SpanType> void createSpanForMatches(final Spannable spannable, final Pattern pattern, final SpanCreator<SpanType> creator, int... groupsToMatch) {
```
harshad1, line 175: This change is unrelated to day / night colors per-se. It is just  a cleanup I been meaning to make.

app/thirdparty/java/other/writeily/widget/WrMarkorWidgetProvider.java
```diff
@@ -121,14 +119,4 @@ public void onUpdate(Context context, AppWidgetManager appWidgetManager, int[] a
         }
         super.onUpdate(context, appWidgetManager, appWidgetIds);
     }
-
-    public static void handleWidgetScheme(Context context, RemoteViews remoteViews, Boolean enabled) {
-        if (enabled) {
```
harshad1, line 126: @gsantner I have updated the widget scheme so that it follows the _system_ theme instead of the app theme. This simplifies a lot of code and makes the widget break less. If this is not acceptable, we can revert this change.

app/thirdparty/java/other/writeily/ui/WrRenameDialog.java
```diff
@@ -107,9 +107,7 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
     private AlertDialog.Builder setUpDialog(final File file, LayoutInflater inflater) {
         View root;
         AlertDialog.Builder dialogBuilder;
-        boolean darkTheme = AppSettings.get().isDarkThemeEnabled();
-        dialogBuilder = new AlertDialog.Builder(getActivity(), darkTheme ?
-                R.style.Theme_AppCompat_Dialog : R.style.Theme_AppCompat_Light_Dialog);
+        dialogBuilder = new AlertDialog.Builder(getActivity(), AppSettings.get().getDialogLayout());
```
harshad1, line 110: TODO: replace function with night-qualified resource

harshad1:
@gsantner This PR is getting closer to done. If you're able, could you try this branch and check if the colors are broken anywhere?

app/thirdparty/java/other/writeily/widget/WrMarkorWidgetProvider.java
gsantner: Thats good

app/src/main/java/net/gsantner/markor/App.java
gsantner: you can move it to onCreate when only used once and here

gsantner: is there another point where we can add it (or call once before switching to view mode?). I know there are a lot of devices which have issues with webview, but they can use Markor normally for everything except view mode. With that change in Application-Object it may happen that it SIGSEV's immediatly on those devices

app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -556,50 +555,16 @@ public boolean isEditorHistoryEnabled() {
...
-                return rcolor(R.color.solarized_fg);
-            case 3:
-                return rcolor(R.color.solarized_fg);
-        }*/
-
-
-        boolean darkMode = isDarkThemeEnabled();
-        int defval = rcolor(darkMode ? R.color.white : R.color.dark_grey);
-        return getInt(darkMode ? R.string.pref_key__editor_basic_color_scheme__fg_dark : R.string.pref_key__editor_basic_color_scheme__fg_light, defval);
+        return getInt(R.string.pref_key__editor_basic_color_scheme__fg, rcolor(R.color.primary_text));
```
gsantner, line 559: this looks like it reduces user configurable colors to only one (for either day/night, instead of one of day and one for night)

app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -556,50 +555,16 @@ public boolean isEditorHistoryEnabled() {
...
-                return rcolor(darkMode ? R.color.solarized_bg_dark : R.color.solarized_bg_light);
-            case 3:
-                return rcolor(darkMode ? R.color.solarized_bg_dark : R.color.solarized_bg_light);
-        }*/
-
-        boolean darkMode = isDarkThemeEnabled();
-        int defval = rcolor(darkMode ? R.color.dark_grey : R.color.light__background);
-        return getInt(darkMode ? R.string.pref_key__editor_basic_color_scheme__bg_dark : R.string.pref_key__editor_basic_color_scheme__bg_light, defval);
-
+        return getInt(R.string.pref_key__editor_basic_color_scheme__bg, rcolor(R.color.background));
```
gsantner, line 563: this looks like it reduces user configurable colors to only one (for either day/night, instead of one of day and one for night)

app/src/main/res/values-night/string-not_translatable.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="pref_key__editor_basic_color_scheme__bg" translatable="false">@string/pref_key__editor_basic_color_scheme__bg_dark</string>
+    <string name="pref_key__editor_basic_color_scheme__fg" translatable="false">@string/pref_key__editor_basic_color_scheme__fg_dark</string>
+</resources>
```
harshad1, line 5: @gsantner We still have dark and light versions. It is just that the decision is made by the system, not in code

app/src/main/java/net/gsantner/markor/App.java
harshad1: It may be possible. I will have to investigate. We should also just test and check if this actually causes anyone issues before spending too much effort.

app/src/main/java/net/gsantner/markor/ui/FilesystemViewerCreator.java
```diff
@@ -56,10 +54,10 @@
...
         opts.accentColor = R.color.accent;
         opts.primaryColor = R.color.primary;
-        opts.primaryTextColor = darkTheme ? R.color.dark__primary_text : R.color.light__primary_text;
-        opts.secondaryTextColor = darkTheme ? R.color.dark__secondary_text : R.color.light__secondary_text;
-        opts.backgroundColor = darkTheme ? R.color.dark__background : R.color.light__background;
-        opts.titleTextColor = titleLight ? R.color.dark__primary_text : R.color.light__primary_text;
+        opts.primaryTextColor = R.color.primary_text;
+        opts.secondaryTextColor = R.color.secondary_text;
+        opts.backgroundColor = R.color.background;
+        opts.titleTextColor = R.color.primary_text;
```
harshad1, line 60: Do you want to remove these options and just move them into the layouts?

app/src/main/java/net/gsantner/markor/App.java
harshad1: Actually, I am 90% sure this will not cause an issue - we instantiate (but leave hidden) a webview when we inflate `DocumentEditFragment`. So any users who see a crash on _instantiating_ a webview will be crashing anyway

app/src/main/java/net/gsantner/markor/App.java
```diff
@@ -10,27 +10,41 @@
...
 
+    public static Resources getResouces() {
+        return res;
+    }
+
     @Override
     public void onCreate() {
         super.onCreate();
         _app = this;
+        res =  getResources();
```
gsantner, line 42: Whats the reason/use for res? It's private and all it does is

```
func getRes() {
return this.res;}

func onCreate() {
  this.res = getRes(); 
   -->  this.res = this.res
   --> this.res is unused and just null assigned multiple times 
}

harshad1:
I initially added it as a way to get resources in a static context.

Will be removing it as it turned out to be unnecessary.

Dec. 28, 2021 12:49:32 Gregor Santner ***@***.***>:

harshad1:
@gsantner Is it possible to get more people to test this on more devices. My primary device is a google pixel, which represents an 'ideal' scenario.

harshad1:
@el416 As part of this PR, the widget theme has been updated so that it will follow the _system_ setting, not the app setting. Any opinions?

gsantner:
I will try it again for sure, for other people, the best place to ask is in the general github discussion, and on matrix

app/src/main/java/net/gsantner/markor/ui/FilesystemViewerCreator.java
gsantner: The named colors from overall palette could be moved to theme yes.

grafik

Preferably in a second PR, to not overlook something here

gsantner:
something left on your side?

harshad1:
Still tweaking a few things here and there. Please review / merge in the shortcut PR first

gsantner:
I added another theme option, black. With the new state it was not too hard to add, and it's one of the very popular things. (People should really use issue search :cry: )

harshad1:
IMO this is ready. Haven't found any issues in several days.

gsantner:
@harshad1  
I did another round of bigger improvements, checkout master! Feels way better & faster now to me.

gsantner:
@harshad1  
I did another round of bigger improvements, checkout master! Feels way better & faster now to me.  

A bigger thing that was missing in the PR was that the entrance point might not be MainActivity. So if you came in by share-into/widget/shortcut, you always get white theme.

harshad1:
Nice changes! And thanks for catching that color bug. I didn't realize as I use dark theme in both markor and Android itself.

PR #1544, opened 2021-12-25
BUGFIX: null document

harshad1:
Document can be null in this context - specifically when launching Action Order Activity

### DIFF
--- a/app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
+++ b/app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
@@ -63,14 +63,14 @@ public abstract class TextActions {
     private static final String ORDER_SUFFIX = "_order";
     private static final String DISABLED_SUFFIX = "_disabled";
 
-    public TextActions(Activity activity, Document document) {
+    public TextActions(final Activity activity, final Document document) {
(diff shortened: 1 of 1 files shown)


gsantner:
@harshad1   
noticed new bug on master: When you create a new (empty file) with Markor, enter text, and go out --- you will see content is not saved. Autosaving seems broken when file initially was empty, only manual save press works. As soon there is content, autosaving also works.

gsantner:
ah,, seems to be the new MIN_LENGTH of 5.

grafik

I would revert that to 1 or 2, what do you think?

harshad1:
1 is good, I think. Least surprising.

gsantner:
there seems to be also  an issue with saving the cursor position. 

1) open file
2) go to somewhere in document
3) go back to mainactivity / close app
4) open document again, and the position is again where it was at 1)

harshad1:
Hmm. I see it. Will look into it.

Dec. 31, 2021 09:34:28 Gregor Santner ***@***.***>:

> there seems to be also an issue with saving the cursor position.
> 
1. > open file
2. > go to somewhere in document
3. > go back to mainactivity / close app
4. > open document again, and the position is again where it was at 1)

Issue #1393, opened 2021-07-17
In math mode, the percent sign is not shown


parabolapersuader:
## General information
* App version: 2.7.0
* Android version: 10
* Device: Phone
* I understand that this issue will be closed without further notice in case I didn't describe the problem including general information, or didn't [search](https://github.com/gsantner/markor/issues?q=) for similar issues & [requests](https://github.com/gsantner/markor/discussions?discussions_q=).

### Description

Entering something simple like `$22\%$` will show 22<space> instead of the percent sign.

Issue #1389, opened 2021-07-11
Errorneous TeX rendering where multiple `_` or `^` exist in math expressions


Gennadiyev:
## General information
* App version: 2.7.0
* Android version: 10 
* Device: HUAWEI Matepad Pro 10.8

## Description

An attempt was made to render the following formula:

```latex
$\delta ^ {(n)} (at) = \dfrac{1}{|a|} \cdot \dfrac{1}{a ^ n} \delta ^ {(n)} (t)$
```

But it shows bad results:

![Screenshot_20210711_145641.jpg](https://i.loli.net/2021/07/11/gd1MeatlQWEYDJz.jpg)

Similarly, the following equation will also fail to render properly:

```latex
$\left.\mathrm{d}x_1\right|_{x_1=a}$
```

But has a workaround by adding spaces before and after each underscore, to prevent markdown-like syntax from being rendered first.

## References

#952 and #825 are two similar issues about rendering TeX, where a suggestion is provided to use `<div>` to wrap around the markdown block. However, it is a bad practice for markdown writers to use HTML tags within their files. Meanwhile, this issue technically only exist in markor. Because markor is not the only markdown editor I'm currently using, I don't feel necessary to implement this workaround.

## Guessings

I haven't run into the codes yet, but it's quite apparent that markor is attempting to **parse what's wrapped inside the `$...$`, which corrupts the expression before it's sent to KaTeX**.

I fully understand that it may be part of KaTeX's fault, but I'm a MathJax and KaTeX user, regularly calling them in my own projects without such an error.

> I'm not native to English, please forgive for ambiguity.

Gennadiyev:
Quick note: #1393 will also be resolved if KaTeX does not parse beforehands.

gsantner:
feel free to improve math support at Markor, and/or Flexmark (Markors markdown parser/converter).

Markor is not a Markdown parser/converter.

PyTran:
$w^{12}$ is very big $w=0$ is nothing

$w^{12}$ is very big $w^0=0$ is nothing

These is also a related problem when there are two superscripts in the same paragraph. In the above the top line renders correctly but the second line does not.

fgtham:
The easiest way would be to just disable the flexmark Superscript extension:

https://github.com/gsantner/markor/blob/79317e33503a02662ffd7e467a77c343247544e8/app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java#L111

[...]

PR #1575, opened 2022-02-02
Add support for mhchem (KaTeX extension)

fgtham:
KaTeX provides the [mhchem](https://github.com/KaTeX/KaTeX/tree/main/contrib/mhchem)  extension for typesetting chemical formulas. This PR adds the JS asset for KaTeX v011.1 and loads the file right after `katex.min.js`. I hope that's all there is to it :)


gsantner:
Thx. I happily merge. However I don't use KaTex on my own.

There are some issues open about KaTex at Markor and nobody has stepped up to do something about it so far. As you seem to have some idea about it, it would be very cool if you can also check those? <https://github.com/gsantner/markor/issues/1389> <https://github.com/gsantner/markor/issues/1393>

fgtham:
> There are some issues open about KaTex at Markor and nobody has stepped up to do something about it so far. As you seem to have some idea about it, it would be very cool if you can also check those? #1389 #1393

Sure, I will have a look.

> I see there are other KaTex extensions as well. Do you see any other of them as something you would use regular?

There seem to be two other extensions: [Copy-tex](https://github.com/KaTeX/KaTeX/tree/main/contrib/copy-tex) and [`math/tex` Custom Script Type](https://github.com/KaTeX/KaTeX/tree/main/contrib/mathtex-script-type). I am not familiar with either one.

PR #1579, opened 2022-02-04
Simplify document handling

harshad1:
Clean up document handling in `DocumentActivity` and related locations


gsantner:
I just left from my job and get a visit now. Won't have much time anyway :D

gsantner:
I just left from work and get a visit now. Won't have much time anyway :D

gsantner:
Btw, what I think off doing maybe somewhen in future is making the editor+webview+buttonbar combo more like a Android View. Instead of fragment.  

And then just have DocumentActivity containing the cool view. Own View in ShareIntoActvitity for the top thing. Own View in MainActivity.

Not 100% sure if, but thats somewhere in my mind. Fragments are just horrible to work with to be honest :smile: .

harshad1:
This would be ideal, I think. Having to deal with 2 slightly different lifecycles is really annoying.

Feb. 4, 2022 09:04:01 Gregor Santner ***@***.***>:

harshad1:
I am satisfied with my testing. This is ready.

gsantner (approved):

PR #1582, opened 2022-02-04
Remove EXTRA_PATH_IS_FOLDER intent parameter everywhere

harshad1:
We put EXTRA_PATH_IS_FOLDER in all our intents, but this is not really necessary anywhere... Removing this and seeing what the effects are.


harshad1:
Haven't found any issues with this in my testing. Ready to merge, IMO

gsantner (approved):

PR #1584, opened 2022-02-06
Add support for flexmark admonition extension

fgtham:
The [Admonition Extension](https://github.com/vsch/flexmark-java/wiki/Admonition-Extension) allows for easy creation of beautiful colored boxes:

```
!!! note "Note"
    This is the note body. Full markdown content is allowed.
```

I think it's quite nice to have, as I often work with side notes and boxed content. What do you think?


gsantner:
what parts are from you and which is css/js by the flexmark extension developers? Or did you fully write the js/css on your own?

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
```diff
@@ -122,7 +126,8 @@
             YamlFrontMatterExtension.create(),
             TypographicExtension.create(),        // https://github.com/vsch/flexmark-java/wiki/Typographic-Extension
             GitLabExtension.create(),             // https://github.com/vsch/flexmark-java/wiki/Extensions#gitlab-flavoured-markdown
-            FootnoteExtension.create()            // https://github.com/vsch/flexmark-java/wiki/Footnotes-Extension#overview
+            FootnoteExtension.create(),           // https://github.com/vsch/flexmark-java/wiki/Footnotes-Extension#overview
+            AdmonitionExtension.create()
```
gsantner, line 130: suggestion: insert above footnote (=1 change), and add the link to the extension documentation

gsantner:
what parts are from you and which is css/js by the flexmark extension developers? Or did you fully write the js/css on your own?

I just see you tweaking the css/js a bit. If you had the original files, and applied tweaks to it...it's better to keep the original and apply overrides in a separate file. because when doing maintainence and updating the css from upstream ... somebody needs take time again to find out what you did custom, and how to restore it

fgtham:
Good point. I reset `admonition.css` to the original upstream contents and added my tweaks in a second file where needed.

I don't know what to do with the javascript, though. I needed to add an event listener to the otherwise unmodified JS for it to work.

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
fgtham: applied

gsantner:
so far I went with including custom CSS/JS directly as oneliner string in java. Depending on how big the customizations are it might or might not sense to do it same here

fgtham:
The changes to CSS are a bit too big (53 LoC), I would not add them as oneliner to Java. Instead I moved them to `admonition-custom.css`.

I also reset `admonition.js` to the upstream version. Unfortunately two minor modifications to the file are still neccessary. I made them in place and commented them accordingly to facilitate future maintenance. I hope that's ok.

fgtham:
I see you have some *really* long CSS onliners in Java :) I also moved the custom CSS into Java and removed `admonition-custom.css`.

gsantner:
ok thats exactly what I wante to know ... thats fine then^^

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
```diff
@@ -101,6 +102,10 @@
...
 
+    public static final String HTML_ADMONITION_INCLUDE = "<link rel='stylesheet'  type='text/css' href='file:///android_asset/flexmark/admonition.css'>" +
+            "<script src='file:///android_asset/flexmark/admonition.js'></script>";
+    public static final String CSS_ADMONITION = CSS_S + ".adm-block { width: initial; font-size: 90%; text-indent: 0em; } .adm-heading { height: auto; padding-top: 0.4em; padding-left: 2.2em; padding-bottom: 0.4em; } .adm-body { padding-top: 0.25em; padding-bottom: 0.25em; margin-left: 0.5em; margin-right: 0.5em; } .adm-heading > span { color: initial; } .adm-icon { position: absolute; top: 50%; left: 0.5em; transform: translateY(-50%); } .adm-block > .adm-heading { position: relative; cursor: pointer; } .adm-block.adm-collapsed > .adm-heading { margin-bottom: 0; } .adm-block.adm-collapsed .adm-body { display: none !important; } .adm-block.adm-open > .adm-heading:after, .adm-block.adm-collapsed > .adm-heading:after { top: 50%; transform: translateY(-50%); content: '▼'; } .adm-block.adm-collapsed > .adm-heading:after { content: '◀'; } pre + div.adm-block, div.adm-block + pre { margin-top: 1.75em; }" + CSS_E;
```
gsantner, line 107: if there something where you say, OK this is really special and maybe not so important... it can be also slimmed down. Lesser custom rules the better :D

fgtham: I had three problems with the default CSS:

* Depending on font and font size, content in `div.adm-heading` is not always vertically centered.
* The collapse indicators `▼` and `◀` were not aligned and jumped around, which is irritating and doesn't look nice.
* There was no vertical space between an admonition box and an immediately preceeding or following code block.

The changes in `CSS_ADMONITION` address these issues. I can understand that you prefer less customization, but in this case I think it's neccessary.

gsantner: I'm not about generally removing it. I just want to know if theres something in the the lots custom rules which _is not_ necessary.

fgtham: I cleaned up the CSS a bit and removed unneccessary rules. I would really like to keep the font size and paddings, because it uses space mor efficiently on small displays.

gsantner: It is fine to keep what is neccassary, I really just talked about the unneccassary. Which was good to ask - otherwise nothing would be removed :D

gsantner:
something remaining?

fgtham:
Not that I know of. I've been using it so far without obvious problems.

gsantner:
> What do you think?

That I likely won't ever use it, and most users neither. But it's OK^^.

gsantner (dismissed):

gsantner:
screenshot-2022-02-19__14-45-33

gsantner (approved):
screenshot-2022-02-19__14-45-33

PR #1597, opened 2022-02-13
Support and display YAML front-matter block contents

fgtham:
Most of my documents contain a leading YAML front-matter block with title, creation date and tags attributes. I'd *really* like that information to be displayed on top of the note in view mode. This PR adds support for it. It must be switched on in Settings / Markdown / Show YAML-Block.

These points apply:
* keys must consist of alpha-numeric characters, no spaces, punctuation, etc.
* attribute values are either strings or lists
* the `tags` attribute is special: if it's a string of comma-separated values, it is split into a real list, so these two `tags` are equivalent:
    ```yaml
    tags:
      - tag1
      - tag2
    # equivalent to:
    tags: tag1, tag2
    ```
* single or double quotes surrounding a string are removed

Assume this block:

```yaml
---
foo: "bar baz"
tags: tag1, tag2, tag3
alist:
  - an item
  - another item
---
```

The resulting HTML is:

```html
<div class='yaml-front-matter-container'>
  <div class='yaml-front-matter-item yaml-foo-container'>
    <span class='yaml-foo-item'>
      bar baz
    </span>
  </div>
  <div class='yaml-front-matter-item yaml-tags-container'>
    <span class='yaml-tags-item'>
      tag1
    </span>
    <span class='yaml-tags-item'>
      tag2
    </span>
    <span class='yaml-tags-item'>
      tag3
    </span>
  </div>
  <div class='yaml-front-matter-item yaml-alist-container'>
    <span class='yaml-alist-item'>
      an item
    </span>
    <span class='yaml-alist-item'>
      another item
    </span>
  </div>
</div>
```

The following block:

```yaml
---
title: Markor Markdown Reference
date: 10.02.2022 06:17:35
title: markdown, reference
---
```

Will be displayed like that:

Screenshot_20220213-222258


app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
```diff
@@ -169,6 +185,35 @@ public String convertMarkup(String markup, Context context, boolean isExportInLi
...
         }
 
+        // Extract YAML Front Matter
+        if (!enablePresentationBeamer) {
+            if (appSettings.isMarkdownYamlDisplayEnabled() && markup.startsWith("---")) {
+                Map<String, List<String>> yamlFrontMatterMap = Collections.EMPTY_MAP;
+                yamlFrontMatterMap = extractYamlFrontMatter(markup);
+
+                if (!yamlFrontMatterMap.isEmpty()) {
+                    for (Map.Entry<String, List<String>> entry : yamlFrontMatterMap.entrySet()) {
```
gsantner, line 195: watch out for Zim Wiki (as it's transpiled to Markdown), do the files work fine?

maybe some additional unit-test(s) that compares input and output could be helpful generally

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
```diff
@@ -169,6 +185,35 @@ public String convertMarkup(String markup, Context context, boolean isExportInLi
...
+                        String key = entry.getKey();
+                        List<String> valueList = entry.getValue();
+                        String value = "";
+
+                        if (key.equals("tags")) {
```
gsantner, line 200: I'd say it's fine to leave sorting up to the user. if they add "math" as first tag before "chemistry", they probably do it intentionally so math appears more on top. if not they still can reorder on own.

thats basically also what I do on my jekyll blog posts - most important tags and categories are at top :smile: .

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
```diff
@@ -169,6 +185,35 @@ public String convertMarkup(String markup, Context context, boolean isExportInLi
             head += CSS_PRESENTATION_BEAMER;
         }
 
+        // Extract YAML Front Matter
+        if (!enablePresentationBeamer) {
+            if (appSettings.isMarkdownYamlDisplayEnabled() && markup.startsWith("---")) {
```
gsantner, line 190: `appsettings && nopresentations && startwith`

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
```diff
@@ -195,6 +240,8 @@ public String convertMarkup(String markup, Context context, boolean isExportInLi
                     .set(TocExtension.BLANK_LINE_SPACER, false);
         }
 
+
+
```
gsantner, line 244: unnecessary

app/src/main/res/values-de/strings.xml
```diff
@@ -242,6 +242,7 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="description">Beschreibung</string>
     <string name="edit_picture">Bild bearbeiten</string>
     <string name="math">Mathematik</string>
+    <string name="yaml">YAML-Block anzeigen</string>
```
gsantner, line 245: translations only via translation service. it's not happy about conflicts ;D

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
```diff
@@ -131,10 +138,19 @@
     //## Methods
     //########################
 
+    private Map<String, List<String>> extractYamlFrontMatter(String markup) {
+        Set<Extension> yamlParserExtension = Collections.singleton(YamlFrontMatterExtension.create());
+        Parser yamlParser = Parser.builder().extensions(yamlParserExtension).build();
```
gsantner, line 143: final Parser yamlFrontmatterParser = Parser.builder().exectensions(Collections.singleton...).build()

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
```diff
@@ -169,6 +185,35 @@ public String convertMarkup(String markup, Context context, boolean isExportInLi
...
+                            if (valueList.size() == 1) {
+                                // It's not a real tag list, but rather a string of comma-separated strings.
+                                valueList = Arrays.asList(valueList.get(0).split("(?:,\\s*)"));
+                            }
+                            Collections.sort(valueList);
+                        }
+                        for (String v : valueList) {
+                            v = v.replaceFirst("^(['\"])(.*)\\1", "$2");
+                            value += "<span class='yaml-" + key + "-item'>" + v + "</span>";
+                        }
```
gsantner, line 210: key = "date", "categories", "tags" etc right?

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
```diff
@@ -169,6 +185,35 @@ public String convertMarkup(String markup, Context context, boolean isExportInLi
...
+                        if (key.equals("tags")) {
+                            if (valueList.size() == 1) {
+                                // It's not a real tag list, but rather a string of comma-separated strings.
+                                valueList = Arrays.asList(valueList.get(0).split("(?:,\\s*)"));
+                            }
+                            Collections.sort(valueList);
+                        }
+                        for (String v : valueList) {
+                            v = v.replaceFirst("^(['\"])(.*)\\1", "$2");
+                            value += "<span class='yaml-" + key + "-item'>" + v + "</span>";
```
gsantner, line 209: + v +  --> html safe? If I have `title: "Q & A"`, is `Q & A`(=broken) here or `Q &amp; A` ?

gsantner (changes requested):
My suggestion is to get rid of the `+ v +` part that adds the content.

I suggest instead to place a token there like `{{ post.<<key>> }}`, which is later replaced by the value. 

I did quite a similar thing in another project .... generate the proper html escaped strings, and replace post.key, page.key & site.key with that value.

This means automatically, in case the user has somewhere `{{ page.title }}` in the file, it gets the frontmatter title :smile: 

grafik

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
fgtham: Thanks, I wasn't aware of Zim Wiki. I will have a look.

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
fgtham: fixed

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
fgtham: Yes. The idea is to give the user the possibility to apply custom CSS rules.

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
fgtham: You're right, its broken. Thanks for pointing it out, I will think about it.

fgtham:
I like the idea of using tokens instead, will see if I get it working.

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
fgtham: Should be fixed now.

gsantner:
should be something like that -->

```
for (String k : fronmattermap) {
   final String v = TextUtils.htmlEncode(frontmattermap.at(k));
   for (String scope : "page,post,site".split(",")) {
       content.replace("{{ " + scope + "." + k + " }}", v);
   }
}
```

gsantner:
should be something like that -->

```
for (String k : fronmattermap) {
   final String v = TextUtils.htmlEncode(frontmattermap.at(k).join(", "));
   for (String scope : "page,post,site".split(",")) {
       content = content.replace("{{ " + scope + "." + k + " }}", v);
   }
}
```

gsantner:
let me know when I should take a look again

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
fgtham: I had a look at some Zim files but could not find any issues. Do you have anything in mind which could lead to problems?

fgtham:
It would be great if you could have a look again.

In Settings / Markdown, you can choose which attributes to display as part of the front matter, if any:

Screenshot_20220219-111902
Screenshot_20220219-111914

All attributes are available as tokens `{{ <scope>.<attributeName> }}` throughout the document. Valid scopes are 'page', 'post', 'site' and 'front-matter'. To avoid replacing a token with the correspondig value, escape it with a leading backslash, e.g. `\{{ page.title }}`.

The attributes can be styled by adding CSS rules for `span.<scope>-item-<attributeName>`. In scope 'front-matter', each attribute value is wrapped in a `div.front-matter-container-<attributeName>`, the 'title' and 'tags' attributes already have CSS rules associated. 

Attribute values that are lists are joined into a comma-separated string. The list delimiter can be overwritten with `span.<scope>-delimiter-<attributeName>::before { content: '...'; }`.

gsantner:
TextUtils.join

image

gsantner:
don't make it more complex (also performance wise) than it is. simple search + replace is enough.  
`{{ + scope + . + attrname + }}`

image

gsantner:
I like it that it currently supports markdown. However I have to say - it's incompatible with everything else. I use frontmatter on various things, and it always just pastes the text without doing markdown converting inside that as well.

Text is often quite unpredictable and hard to know in advance all cases - but if anything, at Markor we've seen that even the most advanced regex fits all cases. What I'd keep is the things like html encoding and quotes. But I'd remove code, em, katex and b. If the user wants to have something bold, they can also use css for that.

image

gsantner:
get rid of wrapInDiv - if you want the wrap the stuff in a div it can be done at the template. which is by either the user, or in the predefined case - in the generating code.

image

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
```diff
@@ -323,4 +377,62 @@ public boolean isFileOutOfThisFormat(String filepath) {
...
+            List<String> attrValueOut = new ArrayList<>();
+
+            if (attrName.equals("tags") && attrValue.size() == 1) {
+                // It's not a real tag list, but rather a string of comma-separated strings
+                attrValue = Arrays.asList(attrValue.get(0).split(",\\s*"));
+            }
+
+            if (wrapInDiv) {
+                attrVal_S = HTML_TOKEN_ITEM_CONTAINER_S;
+                attrVal_E = HTML_TOKEN_ITEM_CONTAINER_E + "\n";
```
gsantner, line 406: get rid of wrapInDiv - if you want the wrap the stuff in a div it can be done at the template. which is by either the user, or in the predefined case - in the generating code.

image

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
```diff
@@ -323,4 +377,62 @@ public boolean isFileOutOfThisFormat(String filepath) {
...
+            for (String aValue : attrValue) {
+                // Strip surrounding single or double quotes
+                aValue = aValue.replaceFirst("^(['\"])(.*)\\1", "$2");
+                aValue = TextUtils.htmlEncode(aValue);
+                aValue = aValue.replaceAll("`(.*?)`", "<code>$1</code>");
+                aValue = aValue.replaceAll("_(.*?)_", "<em>$1</em>");
+                aValue = aValue.replaceAll("\\*(.*?)\\*", "<b>$1</b>");
+                aValue = aValue.replaceAll("(?<!-)---(?!-)", "&mdash;");
+                aValue = aValue.replaceAll("(?<!-)--(?!-)", "&ndash;");
+                aValue = aValue.replaceAll("\\$(.*?)\\$", "<span class='katex'>$1</span>");
```
gsantner, line 420: I like it that it currently supports markdown. However I have to say - it's incompatible with everything else. I use frontmatter on various things, and it always just pastes the text without doing markdown converting inside that as well.

Text is often quite unpredictable and hard to know in advance all cases - but if anything, at Markor we've seen that even the most advanced regex fits all cases. What I'd keep is the things like html encoding and quotes. But I'd remove code, em, katex and b. If the user wants to have something bold, they can also use css for that.

image

Preferably I'd like to not introduce something that doesn't work on any other markdown software / converter.

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
```diff
@@ -323,4 +377,62 @@ public boolean isFileOutOfThisFormat(String filepath) {
...
+                attrValueOut.add(HTML_TOKEN_ITEM_S + aValue + HTML_TOKEN_ITEM_E);
+            }
+
+            // Replace "{{ <scope>>.<key> }}" tokens in note body, if they are not escaped with a preceeding backslash
+            for (String scope : scopes.split(",\\s*")) {
+                String tokenPattern = "(?<!\\\\)\\{\\{ " + scope + "\\." + attrName + " \\}\\}";
```
gsantner, line 426: don't make it more complex (also performance wise) than it is. simple search + replace is enough.  
`{{ + scope + . + attrname + }}`

image

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
```diff
@@ -323,4 +377,62 @@ public boolean isFileOutOfThisFormat(String filepath) {
...
+                String delimiter = "<span class='" + scope + "-delimiter-" + attrName + " delimiter'></span>";
+                String replacement = attrVal_S + String.join(delimiter, attrValueOut) + attrVal_E;
```
gsantner, line 428: TextUtils.join

image

gsantner (changes requested):

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
gsantner: no. i dont use zim either.

@fredericjacob

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownTextConverter.java
fgtham: The idea was to prevent tokens from being replaced: `{{ page.title }}` would expand, while `\{{ page.title }}` would be left untouched.

As an alternative I can move the backslash into the braces, e.g. `{{ \page.title }}`. Then I could get rid of the `tokenPattern` regex and use simple string replacement instead.

gsantner: no, don't add special escape code here.

if post.title is "hello"

and somewhere it is .... \{{ post.title }} it's fine to get \5

fgtham: Sometimes maybe not. Consider the following example. It would be impossible to type this:

``````
---
title: My Jekyll Notes
---

Here's how to use tags in Jekyll:

```
<ul>
  {% for post in site.posts %}
    <li>
      <a href="{{ post.url }}">{{ post.title }}</a>
    </li>
  {% endfor %}
</ul>
```
``````

It would render as:

``````
Here's how to use tags in Jekyll:

```
<ul>
  {% for post in site.posts %}
    <li>
      <a href="{{ post.url }}">My Jekyll Notes</a>
    </li>
  {% endfor %}
</ul>
```
``````

gsantner: this is correct - but it's not what we are goaling for. markor doesn't have any CMS blog software or Liquid builtin. What you post there is mostly about writing i.e. pagination internals for a blog.

what however more the goal is here - that you can write articles with a yaml ----title: "something"--- and you also get rendered that title in the file where it's references.

again - goal isn't a full blown static site generator. keep it simple :smile:

fgtham: No, I mean the literal code example. Assume I'm learning to use Jekyll and make notes of it using Markor. I would write `I can refer to the current document's title using {{ post.title }}.` It would not render correctly.

gsantner: Ah yeah well there is a pretty simple way here - don't specify title in yaml then. Or use invisible space character somewhere there. I prefer to not lose performance by regex look arounds to this very special case - where you want to just describe the use case.

fgtham: Using invisible space is a nice idea, I like it :smile: Will drop the escape stuff.

fgtham:
Does it make sense to have three scopes 'page', 'post' and 'site'? If we drop two and only use 'post', the code will be even simpler. No need to iterate over the scope list for every attribute any longer, cutting down the mumber `replace()` calls.

gsantner:
I'm OK with it if it is just `post.`

gsantner:
I'm OK with it if it is just `post.` Though I would keep the code as is, but just remove site & page from the array. So a loop with just one entry then.

gsantner:
I took care of it, you don't need to make change for this.

gsantner:
do you have something left, otherwise I merge it now

fgtham:
Thanks! I have nothing left, have been using it for a while with no apparent issues.

fgtham:
In app/src/main/res/values/strings.xml line 261 shouldn't this read 'post.title' instead of 'page.title'?

gsantner:
now yes, before reducing to just post .. it would have worked : - )

gsantner (dismissed):

PR #1622, opened 2022-02-22
Fixes for spacing and save

harshad1:
This is a follow up PR for the reduction in spacing. Due to the way padding and minWidth interacted, the extra icon spacing would not take effect until the extra icon spacing was > 8. In this PR, I have fixed it so spacing is calculated and applied correctly.


app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -663,15 +663,14 @@ public String getFragmentTag() {
     // Save the file
     // Only supports java.io.File. TODO: Android Content
     public boolean saveDocument(final boolean forceSaveEmpty) {
-        if (!isAdded()) {
+        if (!isAdded() || _hlEditor == null || _document == null) {
```
harshad1, line 666: I've never seen this fail on my phone, but I noticed it occasionally failing in the emulator, so I added some defensive code.

app/src/main/res/layout/quick_keyboard_button.xml
```diff
@@ -11,17 +11,15 @@
...
-    android:layout_height="wrap_content"
+    android:layout_width="24dp"
+    android:layout_height="24dp"
     android:background="@drawable/keyboard_shortcut_button"
     android:gravity="center_vertical|center_horizontal"
-    android:minWidth="40dp"
-    android:minHeight="48dp"
-    android:paddingStart="12dp"
-    android:paddingLeft="12dp"
+    android:paddingLeft="8dp"
```
harshad1, line 18: 24dp icon + (8 * 2 = 16) dp padding = 40dp min width

app/src/main/res/layout/quick_keyboard_button.xml
```diff
@@ -11,17 +11,15 @@
...
+    android:paddingRight="8dp"
     android:paddingTop="8dp"
-    android:paddingEnd="12dp"
-    android:paddingRight="12dp"
     android:paddingBottom="8dp"
+    android:scaleType="fitCenter"
     android:tint="@color/textActionIconColor"
+    android:src="@drawable/ic_add_white_24dp"
```
harshad1, line 24: Default drawable makes it easier to debug the layout. will not show up in practice.

gsantner: tools:src ?

harshad1: Anything, honestly. Just so that there is a default icon I can see in the layout :)

gsantner: then tools: instead android: is exactly whats needed

gsantner (approved):

PR #1623, opened 2022-02-22
Set HTML lang attribute

fgtham:
After conversion from markdown to html, the document is wrapped in a `div` with the `lang` attribute set to the app language. For a specific document, the language can be overwritten by specifying the `lang` attribute in the YAML header.

This change alone will not do anything visible to the user, but now she can easily enable hyphenation, if desired, by injecting the appropriate CSS:

```css
html {
    word-break: break-word;
    hyphens: auto;
}
```

This might use display space more efficiently on small devices.


gsantner:
i think the lang attribute should be at the `<html>` tag, not? At least thats the only place I have seen it yet in files

fgtham:
It's common to set the language on the `html` element, but  [not required](https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes). As global attribute it can be used on any element.

I have not found a simple way to access the `html` element from MarkdownTextConverter.Java, that's why I used the `div` tags instead. It works 🙂

gsantner:
I recommend to modify the template in the base convert class, and put language in the html attribute. The language can be applied there, no additional base parameter needed. This means free language attribute for everything converting to html.

`HTML001_HEAD_WITH_BASESTYLE`

fgtham:
Changed it to set `<html lang='...'>` to the current language. It's not possible to overwrite the attribute value from within a document, but probably that's not needed anyway :)

app/src/main/java/net/gsantner/markor/format/TextConverter.java
```diff
@@ -43,7 +44,7 @@
     protected static final String TOKEN_POST_TODAY_DATE = "{{ post.date_today }}";
 
     protected static final String HTML_DOCTYPE = "<!DOCTYPE html>";
-    protected static final String HTML001_HEAD_WITH_BASESTYLE = "<html><head>" + CSS_S + "html,body{padding:4px 8px 4px 8px;font-family:'" + TOKEN_FONT + "';}h1,h2,h3,h4,h5,h6{font-family:'sans-serif-condensed';}a{color: " + TOKEN_LINK_COLOR + ";text-decoration:underline;}img{height:auto;width:325px;margin:auto;}" + CSS_E;
+    protected static final String HTML001_HEAD_WITH_BASESTYLE = "<html lang='" + Locale.getDefault().getLanguage() + "'><head>" + CSS_S + "html,body{padding:4px 8px 4px 8px;font-family:'" + TOKEN_FONT + "';}h1,h2,h3,h4,h5,h6{font-family:'sans-serif-condensed';}a{color: " + TOKEN_LINK_COLOR + ";text-decoration:underline;}img{height:auto;width:325px;margin:auto;}" + CSS_E;
```
gsantner, line 47: does not respect the app's set language. it's static and thus will use the system language. you need to set the value in the convert method, you can't do it with static var properly

gsantner (changes requested):

app/src/main/java/net/gsantner/markor/format/TextConverter.java
fgtham: Doesn't `Locale.getDefault().getLanguage()` return the app language?

gsantner: * start app, use system settings to change language, see wrong language
* start app, change language inside app's settings, see wrong language

fgtham: Should be working now.

gsantner (approved):

PR #1624, opened 2022-02-23
Snippets

harshad1:
This is ~~an experimental~~ a snippets PR

- Snippets directory is `<notebook>/.app/snippets`
- Each file in this directory will represent 1 snippet
- Snippets will be interpolated with date-time formatting
    - Regions in the snippet delimited by backticks will be interpolated with SimpleDateTime formatting 
- ~~Snippets are file-type specific. Each type has a corresponding snippet extension~~
    - ~~Markdown => `.md`~~
    - ~~Todo.txt => `.todo`~~
    - ~~Zim => `.zim`~~
    - ~~Plain text => `.txt`~~
- Snippets are inserted an action.
- The action will insert the snippet at the current cursor location

- [x] Fix datetime interpolation
- [x] Long press to insert last snippet?


harshad1:
@gsantner I am getting pretty happy with this approach. It is simple, unobtrusive and quite useful. Opinions would be appreciated.

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
```diff
@@ -640,6 +645,54 @@ public static void showSetPasswordDialog(final Activity activity) {
...
+            case (TextFormat.FORMAT_MARKDOWN):
+                return ".md";
+            case (TextFormat.FORMAT_PLAIN):
+                return ".txt";
+            case (TextFormat.FORMAT_TODOTXT):
+                return ".todo";
+            case (TextFormat.FORMAT_ZIMWIKI):
+                return ".zim";
+        }
+        return "";
```
harshad1, line 659: Each file format has it's own snippet extension

harshad1:
@gsantner This is ready for discussion and possible merging.

app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
```diff
@@ -618,6 +620,20 @@ protected boolean runCommonTextAction(final @StringRes int action) {
                 runRenumberOrderedListIfRequired();
                 return true;
             }
+            case R.string.tmaid_common_insert_snippet: {
```
gsantner, line 623: suggestion: Long press button shows dialog, copying the selected portion of text (or all if no selection) into new snippet. aka  "create new snippet"

app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
```diff
@@ -618,6 +620,20 @@ protected boolean runCommonTextAction(final @StringRes int action) {
...
+                SearchOrCustomTextDialogCreator.showInsertSnippetDialog(_activity, _document.getFormat(), (snip) -> {
+                    _hlEditor.replaceCurrentSelection(StringUtils.interpolateEscapedDateTime(snip));
+                    _appSettings.setLastFormatSnip(_document.getFormat(), snip);
+                });
+                return true;
+            }
+            case R.string.tmaid_common_insert_recent_snippet: {
```
gsantner, line 630: I would keep it simple, just one button.

app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -352,6 +352,15 @@ public void toggleFavouriteFile(File file) {
...
     private static final String PREF_PREFIX_VIEW_SCROLL_Y = "PREF_PREFIX_VIEW_SCROLL_Y";
+    private static final String PREF_PREFIX_FORMAT_LAST_SNIPPET = "PREF_PREFIX_FORMAT_LAST_SNIPPET";
+
+    public void setLastFormatSnip(final @StringRes int formatId, final String text) {
+        setString(PREF_PREFIX_FORMAT_LAST_SNIPPET + _context.getString(formatId), text);
+    }
+
+    public String getLastFormatSnip(final @StringRes int formatId) {
+        return getString(PREF_PREFIX_FORMAT_LAST_SNIPPET + _context.getString(formatId), "");
+    }
```
gsantner, line 363: > I would keep it simple, just one button.

which also means this is gone

app/src/main/java/net/gsantner/opoc/util/StringUtils.java
```diff
@@ -345,4 +347,35 @@ public static void selectLines(final EditText edit, final List<Integer> position
             edit.setSelection(selStart, selEnd);
         }
     }
+
+    // Search for matching pairs of backticks
+    // interpolate contents of backtick pair as SimpleDateFormat
+    public static String interpolateEscapedDateTime(final String snip) {
```
gsantner, line 353: I suggest this: `date: "<<parsed format>>"`

(Somewhat in sense of <https://learn.cloudcannon.com/jekyll/date-formatting#date>)

app/src/main/java/net/gsantner/markor/ui/SearchOrCustomTextDialogCreator.java
gsantner: I'd keep it simple, just text files which content can be inserted. no format/filetype specifics

specifically also - html for example has no specific markor format assigned yet. But user might want to insert html block independent to what the current file is written in.

app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
harshad1: I definitely want to keep the long - press for the last snippet. For example, I have `<input type="checkbox">`  which I would like to rapidly insert per-line.

Instead, we can do this via the neutral button in the dialog.

app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
harshad1: this is just used to differentiate the long vs short press.

app/src/main/java/net/gsantner/opoc/util/StringUtils.java
harshad1: Hmmm, I was going with something inspired by textmate snippets

app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
```diff
@@ -619,6 +622,19 @@ protected boolean runCommonTextAction(final @StringRes int action) {
                 runRenumberOrderedListIfRequired();
                 return true;
             }
+            case R.string.tmaid_common_insert_snippet: {
+                SearchOrCustomTextDialogCreator.showInsertSnippetDialog(_activity, (snip) -> {
+                    _hlEditor.replaceCurrentSelection(StringUtils.interpolateEscapedDateTime(snip));
+                    _lastSnip = snip;
```
harshad1, line 628: Storing the last snippet in memory in each TextFormat instance. Not stored in appsettings this way. Doesn't matter if data is lost on rotation etc

app/src/main/java/net/gsantner/opoc/util/StringUtils.java
harshad1: Using backticks is easier to parse and escape IMO. We can use a type prefix: `\`date: HHmm\`` to differentiate interpolation types if we want multiple types

app/src/main/java/net/gsantner/opoc/util/StringUtils.java
```diff
@@ -345,4 +347,35 @@ public static void selectLines(final EditText edit, final List<Integer> position
...
+        final StringBuilder interpolated = new StringBuilder();
+        final StringBuilder temp = new StringBuilder();
+        boolean isEscaped = false;
+        boolean inDate = false;
+        for (int i = 0; i < snip.length(); i++) {
+            final char c = snip.charAt(i);
+            if (c == '\\' && !isEscaped) {
+                isEscaped = true;
```
harshad1, line 361: TODO: Don't escape inside an interpolation region

app/src/main/java/net/gsantner/opoc/util/StringUtils.java
```diff
@@ -345,4 +347,35 @@ public static void selectLines(final EditText edit, final List<Integer> position
...
+            } else if (isEscaped) {
+                isEscaped = false;
+                temp.append(c);
+            } else if (c == '`' && inDate) { // Ending a date region
+                inDate = false;
+                interpolated.append(ShareUtil.formatDateTime((Locale) null, temp.toString(), System.currentTimeMillis()));
```
harshad1, line 367: TODO: Delegate to interpolate function

harshad1:
@gsantner Are we good to go here?

app/src/main/java/net/gsantner/markor/ui/hleditor/TextActions.java
gsantner: OK this also better than storting to settings, because if you use multiple windows you can have different "last actions" this way

gsantner:
OK so when the discussions resolved, gonna merge. Though, I don't feel confident enough to tell all users yet that it's the big snippets thing they waited for. Especially by the snippet directory and how templating/date things work.    
Expectations for the snippets seem pretty high to what I saw, I expect many people to ask for many things (Which doesn't mean it needs to do everything, everybody asks for). So let's see it as a first basic implementation that is there to copy-paste small things, which just works without big buzz :-).

app/src/main/java/net/gsantner/markor/format/plaintext/PlaintextTextActions.java
```diff
@@ -102,10 +103,11 @@ public boolean onLongClick(View v) {
                     runRenumberOrderedListIfRequired(true);
                     return true;
                 }
-                default: {
-                    return runAction(_action, true, null);
+                case R.string.tmaid_common_insert_snippet: {
```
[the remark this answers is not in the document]
harshad1: Yes. `runAction` just delegates to `runCommonTextAction` which triggers the short press action. i.e. if we left the default case, long press in PlainText would trigger a short press if a long press action were not defined, instead of showing the tool tip as expected.

gsantner:
Suggestion: To append the list of snippets also to the list at newfiledialog "template spinner". I think the feature is very nice as it is and should fulfill the most needs for the popular templates feature as well.

PR #1653, opened 2022-03-28
WIP: Improvements to show selection after search / heading etc

harshad1:
This PR attempts to scroll to the selection position after search or navigation to heading etc

i.e. make sure the selection is properly centered on the screen.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -60,7 +60,7 @@
             android:exported="true"
             android:label="@string/app_name"
             android:launchMode="singleTop"
-            android:windowSoftInputMode="stateUnchanged|adjustResize"
+            android:windowSoftInputMode="stateAlwaysHidden|adjustResize"
             android:taskAffinity=".activity.MainActivity">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
@@ -171,7 +171,7 @@
             android:parentActivityName=".activity.MainActivity"
             android:taskAffinity=".activity.DocumentActivity"
             android:theme="@style/AppTheme.Unified"
-            android:windowSoftInputMode="stateUnchanged|adjustResize"
+            android:windowSoftInputMode="stateAlwaysHidden|adjustResize"
             tools:targetApi="lollipop">
             <meta-data
                 android:name="android.support.PARENT_ACTIVITY"
--- a/app/src/main/java/net/gsantner/opoc/util/StringUtils.java
+++ b/app/src/main/java/net/gsantner/opoc/util/StringUtils.java
@@ -9,9 +9,18 @@
 #########################################################*/
 package net.gsantner.opoc.util;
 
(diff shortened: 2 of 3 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -60,7 +60,7 @@
             android:exported="true"
             android:label="@string/app_name"
             android:launchMode="singleTop"
-            android:windowSoftInputMode="stateUnchanged|adjustResize"
+            android:windowSoftInputMode="stateAlwaysHidden|adjustResize"
```
harshad1, line 63: This makes it so the keyboard is hidden after the dialog is dismissed.

app/src/main/java/net/gsantner/opoc/util/StringUtils.java
```diff
@@ -345,6 +350,53 @@ public static void selectLines(final EditText edit, final List<Integer> position
...
+            if (parent2 instanceof ScrollView) {
+                vScroll = (View) parent2;
+            }
+        }
+
+        // Find position to set in Y
+        final int[] sel = StringUtils.getSelection(text);
+        final int lineStart = StringUtils.getLineStart(text.getText(), sel[0]);
+
+        final int viewHeight = vScroll.getHeight();
```
harshad1, line 382: `getHeight` gets the height in the current state (i.e. reduced when keyboard is open). We need the height in the keyboard closed state.

harshad1:
Turns out TextView does some pretty complicated stuff to bring the cursor into view. Next step is to just overwrite this to perform the requisite action

app/src/main/java/net/gsantner/opoc/util/StringUtils.java
harshad1: Disregard this, found simpler solution

harshad1:
This is good to merge imo

app/src/main/AndroidManifest.xml
gsantner: doesnt this make it virtual keyboard always invisible (i.e. not popping up) on mainactivity - so breaks todo/quicknote?

especially, when closing virtual keyboard somehow, and you click again in textfield, does it pop up again?

harshad1: The keyboard will not auto-show on launching DocumentActivity or MainActivity. Returning to DocumentActivity or MainActivity from a dialog, for example will also cause the keyboard to be hidden. A tap will show the keyboard.

Works well on my device. Will test `stateHidden` too and report back

gsantner: Thank you

harshad1: `stateAlwaysHidden` is the behavior I think is best - always hide the keyboard when the editor takes focus. i.e. when opened or when returning from a dialog. User can tap to show keyboard.

gsantner (approved):

PR #1658, opened 2022-04-09
Several tweaks to showing selection and related functions

harshad1:
~~Minor fix (not even really a bug). Just reduces chances of functions being called in the wrong order~~

I've ended up making more extensive tweaks and changes. Marking this PR WIP until I am satisfied.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/ui/hleditor/HighlightingEditor.java
+++ b/app/src/main/java/net/gsantner/markor/ui/hleditor/HighlightingEditor.java
@@ -334,8 +334,11 @@ public class HighlightingEditor extends AppCompatEditText {
         if (!hasFocus()) {
             requestFocus();
         }
-        post(() -> StringUtils.showSelection(this, index, index));
-        post(() ->setSelection(index));
+
+        post(() -> {
+            StringUtils.showSelection(this, index, index);
+            postDelayed(() -> setSelection(index), 200); // Run _after_ showSelection
+        });
     }
 
     public void setAccessibilityEnabled(final boolean enabled) {
--- a/app/src/main/java/net/gsantner/markor/ui/hleditor/HighlightingEditor.java
+++ b/app/src/main/java/net/gsantner/markor/ui/hleditor/HighlightingEditor.java
@@ -26,7 +26,7 @@ import net.gsantner.markor.activity.MainActivity;
 import net.gsantner.markor.model.Document;
 import net.gsantner.markor.util.AppSettings;
 import net.gsantner.opoc.util.Callback;
-import net.gsantner.opoc.util.TextUtils;
+import net.gsantner.opoc.util.StringUtils;
 
 import java.io.File;
 import java.util.HashSet;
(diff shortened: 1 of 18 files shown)


app/src/main/java/net/gsantner/markor/ui/hleditor/HighlightingEditor.java
```diff
@@ -334,8 +334,11 @@ public void setCursor(final int index) {
         if (!hasFocus()) {
             requestFocus();
         }
-        post(() -> StringUtils.showSelection(this, index, index));
-        post(() ->setSelection(index));
+
+        post(() -> {
+            StringUtils.showSelection(this, index, index);
+            postDelayed(() -> setSelection(index), 200); // Run _after_ showSelection
```
gsantner, line 340: make sure this runs before autoformat is enabled again (on startup). otherwise, if you have a todo.txt file with a trailing line, or markdown with a 1./2./3. before last line, you get unintenionally something inserted.  
This possibly currently goes around that because selection is applied async later.

<https://github.com/gsantner/markor/commit/6014327f346af269533e7742397aad5fe52445d6>

harshad1: This shouldn't trigger any change in the text. It just shows the selection. I am pretty sure the bug with auto format and insertion was from the `setText`, not the `setSelection`. Will test for a few more days to confirm.

Also note that I added `withAutoFormatDisabled(final Callback.a0 callback)` in `HighlightingEditor`

app/src/main/java/net/gsantner/opoc/util/TextUtils.java
```diff
@@ -417,6 +417,20 @@ public static void showSelection(final TextView text, final int start, final int
...
+            edit.post(() -> {
+                if (!edit.hasFocus()) {
+                    edit.requestFocus();
+                }
+
+                edit.setSelection(start, _end);
+                edit.post(() -> showSelection(edit, start, _end));
+            });
+        }
+    }
```
harshad1, line 432: Made this function static and moved it here as it is general.

app/src/main/java/net/gsantner/opoc/util/TextUtils.java
```diff
@@ -325,30 +327,28 @@ public static void selectLines(final EditText edit, final List<Integer> position
...
                 for (int i = 0; i < lines.length; i++) {
                     (pSet.contains(i) ? sel : unsel).add(lines[i]);
                 }
                 sel.addAll(unsel);
-                final String newText = TextUtils.join("\n", sel);
+                final String newText = android.text.TextUtils.join("\n", sel);
                 edit.setText(newText);
                 selStart = 0;
-                selEnd = StringUtils.getIndexFromLineOffset(newText, positions.size() - 1, 0);
+                selEnd = TextUtils.getIndexFromLineOffset(newText, positions.size() - 1, 0);
```
harshad1, line 349: Changed the name from `StringUtils` -> `TextUtils` to better reflect what it contains.

gsantner:
something really bad happened?

harshad1:
Hmm no. I think I closed this accidentally.

harshad1:
I think this is good to merge.

- Some tweaks to show selection heuristic
- Some tweaks to how functions are applied
- File name change

PR #1669, opened 2022-04-23
Fix: Template is not applied when creating new file.

tifish:
It seems the condition has some defects.


gsantner:
Makes sense, thanks

gsantner (approved):
Makes sense, thanks

PR #1678, opened 2022-04-24
[edu] add more fonts in app

YG-yyds:
add fonts that from the Internet.

do not merge. We just use the font that can not be used in business


gsantner:
every font is bloating app, they need to be carefully chosen. Only open source/font license is acceptable, with license added. So if it really need to be a asia specific font, rather just one with wide language & glyph support (like Noto).

YG-yyds:
Thanks, and I understand it.

gsantner:
see #1679

PR #1685, opened 2022-04-24
[edu] change location

Shallow-D12:
"add_file" button'location can be changed like this.


harshad1:
@Shallow-D12 Would you be able to post a screenshot of this change?

Shallow-D12:
> @Shallow-D12 Would you be able to post a screenshot of this change?

button

harshad1:
IMO this increases the chance that the add button will block some of the last item in the explorer view.

Gmail, Fairmail etc all use a compose button to the right (like markor does currently) and I think it works better this way.

Shallow-D12:
> IMO this increases the chance that the add button will block some of the last item in the explorer view.
> 
> Gmail, Fairmail etc all use a compose button to the right (like markor does currently) and I think it works better this way.

But for some left-handers, this layout makes the button hard to press.

gsantner:
see #1679

PR #1688, opened 2022-04-27
Renumber list performance improvement

harshad1:
This PR improves performance when renumbering a list.

Renumbering is currently done in place. This adds an undo and triggers a check for whether the text has changed multiple times in rapid succession.

In this PR the renumber has been modified to work on a _copy_ of the text. The original editable is updated in one step by diffing against the updated copy.

Have also improved the way accessibility is disabled for some operations


harshad1:
This is ready

gsantner (approved):
thank you

PR #1689, opened 2022-04-29
Keep in-files search dialog open, closes #1663, by @harshad1

harshad1:
This PR adds the following features to the in-files search results:

1. The dialog will stay open when an item is clicked. Can return to main via back or recent apps and open another file.
2. The file names (group names) have been made bold
3. Long clicking the file name will take you to the top of the file


app/src/main/java/net/gsantner/markor/ui/fsearch/FileSearchResultSelectorDialog.java
```diff
@@ -194,7 +204,7 @@ public View getGroupView(final int groupPosition, final boolean isExpanded, fina
                 textView = (TextView) inflater.inflate(R.layout.expandable_list_group_item, null);
                 textView.setClickable(false);
             }
-            textView.setText(groupInfo.toString());
+            textView.setText(Html.fromHtml(String.format("<b>%s</b>", groupInfo.toString())));
```
gsantner, line 207: %s string needs to be XML/HTML encoded otherwise might render wrong

gsantner (changes requested):

gsantner (approved):
thanks works good

remove the bold though..it was just everything bold - too much. But generally the dialog and it's contents could be a bit face-lifted wording improved / better styled :smile: .

Adirlearn-dev12 (commented):
Me sorprendió la app

Issue #189, opened 2018-02-08
Suggestion: "two spaces" on the toolbar


jerry1970:
#### General information
* **App version:** 0.3.1

#### Description
I regularly type in lyrics. To prevent the whole verse reading as a single paragraph, I have to type in two spaces at the end of each sentence.

But my keyboard app (SwiftKey) turns this into a period and a space unless I keep some time between typing the two spaces.

Might it be useful also for other people to have a button on the toolbar to add two spaces at the end of the current line? Maybe the cursor doesn't have to be at the end of the line. And maybe the app can check whether there are already two spaces to prevent extra spaces being added.

gsantner:
The app won't restrict to use more than two spaces, and won't restrict where to place them (even if some special button is used).

Other than that, it may be possible but, I'm yet unsure about it, because the toolbar has already quite much items and we cannot just put everything in, it that one can think of.

PR #1693, opened 2022-04-30
Supports UTF-8 file with BOM.

tifish:
The code may be a bit rough:
- Does it break any convension?
- Does it cover all cases?
- The icon may be inappropriate. I just pick one from Android Studio.


gsantner:
As of icon, that is OK - thats just what everybody else did so far here :D

app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
```diff
@@ -171,6 +171,11 @@ private AlertDialog.Builder makeDialog(final File basedir, final boolean allowCr
                     final byte[] templateContents = getTemplateContent(templateSpinner, basedir, f.getName(), encryptCheckbox.isChecked());
                     shareUtil.writeFile(f, false, (arg_ok, arg_fos) -> {
                         try {
+                            if (appSettings.isEditorAddBomToNewFile()) {
```
gsantner, line 174: Maybe just make it a toggle on the newfile dialog, not a general settings option. so when you create a new file you tick that on or off if you want to.

what do you think?

app/src/main/java/net/gsantner/opoc/util/FileUtils.java
```diff
@@ -60,6 +61,37 @@ public static String readTextFileFast(final File file) {
         return "";
     }
 
+    public static Pair<String, Boolean> readTextFileFastWithBOM(final File file) {
```
gsantner, line 64: preferably we don't stack read methods onTop onTop onTop with different features.   
We already had that and the current focus on simplifying and debloating.

So what I rather suggest is, have one such function, and have a Pair<String, Map> as result. And in that Map any metadata can be put. With the first case here being BOM information.

What do you think.

app/src/main/java/net/gsantner/opoc/util/FileUtils.java
```diff
@@ -177,8 +209,13 @@ public static byte[] readCloseBinaryStream(final InputStream stream) {
         return baos.toByteArray();
     }
 
-    public static boolean writeFile(final File file, final byte[] data) {
+    public static boolean writeFile(final File file, final byte[] data, boolean withBOM) {
```
gsantner, line 212: when you agree with the metadata comment, you could pass it in here again and so withBom information gets auto extracted from that `Map`.

gsantner (changes requested):
^

app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
tifish: That's a good idea.

app/src/main/java/net/gsantner/opoc/util/FileUtils.java
tifish: I'm not familiar with Java. Thanks for your advice.

tifish:
I committed all changes. See if I missed something.

gsantner:
OK should I try it on the weekend?

app/src/main/java/net/gsantner/markor/model/Document.java
```diff
@@ -257,14 +262,21 @@ public synchronized boolean saveContent(final Context context, final String cont
...
                         fos.write(contentAsBytes);
                         fos.flush();
                     } catch (Exception ignored) {
                     }
                 });
                 success = true;
             } else {
-                success = FileUtils.writeFile(_file, contentAsBytes);
+                final Map<String, Object> options = new HashMap<>(1);
+                options.put("withBom", _withBom);
```
gsantner, line 278: plz make the "withBom" key a `final static` constant

tifish:
Yeah, tell me if anything I can do.

app/src/main/java/net/gsantner/markor/ui/NewFileDialog.java
```diff
@@ -171,6 +177,11 @@ private AlertDialog.Builder makeDialog(final File basedir, final boolean allowCr
...
                     shareUtil.writeFile(f, false, (arg_ok, arg_fos) -> {
                         try {
+                            if (appSettings.getNewFileDialogLastUsedUtf8Bom()) {
+                                arg_fos.write(0xEF);
+                                arg_fos.write(0xBB);
+                                arg_fos.write(0xBF);
+                            }
                             if (templateContents != null && (!f.exists() || f.length() < ShareUtil.MIN_OVERWRITE_LENGTH)) {
                                 arg_fos.write(templateContents);
                             }
```
gsantner, line 187: We should get wrid of own file writing implementation here and instead use FileUtils, what is used at editor as well.

But that is a topic of it's own.

gsantner (approved):

harshad1:
@tifish FYI, The bom test in Document #189 is causing a crash for me on start. The bug does not appear in debug mode, just release mode.

Also, why are we returning a map from readTextFileFast insead of a simple boolean?

app/src/main/java/net/gsantner/opoc/util/FileUtils.java
```diff
@@ -44,20 +47,41 @@ public class FileUtils {
     // Used on methods like copyFile(src, dst)
     private static final int BUFFER_SIZE = 4096;
 
-    public static String readTextFileFast(final File file) {
+    public final static String WITH_BOM = "withBom";
+
+    public static Pair<String, Map<String, Object>> readTextFileFast(final File file) {
+        Map<String, Object> info = new HashMap<>(1);
```
harshad1, line 53: This, why does it need to be a hashmap? From my reading, a boolean will do, right?

app/src/main/java/net/gsantner/opoc/util/FileUtils.java
harshad1: Ah, I see why we chose the map approach :)

PR #1694, opened 2022-05-03
Increase performance on editables, more chunked operations, by @harshad1

harshad1:
In this PR, the chunked operation via copy has been made more generic. It has also been applied to the regexReplaceAction (which is used for a _lot_ of things). Should improve apparent performance and undo behavior. 

TODO:

- [ ] ~~Apply to other common update functions~~
- [ ] ~~Apply to groups of calls (i.e. indent + renumber)~~


gsantner (approved):
Thanks, to my feeling it improves performance quite a bit

Issue #1671, opened 2022-04-24
Saved QuickNote changes disappear sometimes


MaxGyver83:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**. Not a question or feature request.
- [X] The issue is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=) or [discussions](https://github.com/gsantner/markor/discussions?discussions_q=). _(I've searched it)_.
- [X] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.

### Description

I usually use only the QuickNote file. Since some changes to that files got lost, I always press the "save" icon after updating this file. But it still happens that when I return to Markor, my last edit is gone. This is similar to issue #1537 but the difference is that I actively save my changes.

### Steps to reproduce

1. Open QuickNote file in Markor
2. Edit text
3. Save
4. Go to home screen and use another app
5. Return to QuickNote in Markor: Sometimes the last edit is lost.

### Information

Android version: 11 (Oxygen  OS V11.3)
Device: OnePlus Nord 2 5G
App Version: v2.8.6 (130)

### Source

F-Droid

### Format / File type

No

### Additional info / Log

```shell
-
```

gsantner:
try a apk from the GitHub Actions tab and check if it's better

MaxGyver83:
OK, I just installed the latest master build from GitHub Actions (commit 2102dc6). I'll need to test it for some days because this issue did occur only sometimes.

MaxGyver83:
The new version worked fine for some days. But yesterday it happened again. I added some text and when I opened the app again (about 2.5 hours later), the last change was gone.

Is there any log file I could upload here?

gsantner:
There is no log I can think about you could add. But also I couldn't reproduce the issue at any time and I use Markor every day / hour.

Sure that this is not some kind of synchronization issue like Syncthing alike overwriting the file after your edit?

MaxGyver83:
I only use Internxt but it doesn't backup the `QuickNote.md`.
Maybe it's just a problem with OnePlus/Oxygen OS!?

harshad1:
Is the file on an SD card or something?

MaxGyver83:
No. (This device doesn't have an SD card slot.)

harshad1:
[...]

PR #1703, opened 2022-05-16
WIP: More deterministic saving

harshad1:
In this PR I am exploring some improvements to loading / saving logic.


app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -559,17 +579,11 @@ public void onFsViewerConfig(FilesystemViewerData.Options dopt) {
         }
     }
 
-    @OnTextChanged(value = R.id.document__fragment__edit__highlighting_editor, callback = OnTextChanged.Callback.AFTER_TEXT_CHANGED)
-    public void onContentEditValueChanged(CharSequence text) {
-        checkTextChangeState();
-        updateUndoRedoIconStates();
-    }
```
harshad1, line 566: Moved into a debounced callback in the onViewCreated routine

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -216,44 +218,62 @@ public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
...
-                return;
+    public void pause() {
+        saveDocument(false);
+        if (_appSettings != null && _document != null) {
+            _appSettings.addRecentDocument(_document.getFile());
+            _appSettings.setDocumentPreviewState(_document.getPath(), _isPreviewVisible);
+            if (_hlEditor != null) {
+                _appSettings.setLastEditPosition(_document.getPath(), _hlEditor.getSelectionStart());
             }
-            _textSdWarning.setVisibility(permok ? View.GONE : View.VISIBLE);
```
harshad1, line 251: Moved the permission testing into the load and save calls

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
```diff
@@ -381,10 +383,16 @@ public void onViewPagerPageSelected(int pos) {
...
+            ((DocumentEditFragment) _viewPagerAdapter.getItem(_prevPos)).pause();
+        }
+
         if (pos == tabIdToPos(R.id.nav_quicknote) || pos == tabIdToPos(R.id.nav_todo)) {
-            // cannot prevent bottom tab selection
             new PermissionChecker(this).doIfExtStoragePermissionGranted();
+            ((DocumentEditFragment) _viewPagerAdapter.getItem(pos)).resume();
         }
+
+        _prevPos = pos;
```
harshad1, line 395: Explicitly pause and resume when we switch. Do not rely on the visibility hint

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -216,44 +218,62 @@ public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
...
 
+        final Runnable after = StringUtils.makeDebounced(_hlEditor, 500, () -> {
+            checkTextChangeState();
+            updateUndoRedoIconStates();
+        });
+        _hlEditor.addTextChangedListener(TextWatcherDummy.after((s) -> after.run()));
+    }
+
+    @Override
+    public void onFragmentFirstTimeVisible() {
```
harshad1, line 229: Move initial position code into `onFragmentFirstTimeVisible` as we want it to run after the fragment is _displayed_ to the user.

app/src/main/java/net/gsantner/markor/model/Document.java
```diff
@@ -229,18 +230,22 @@ public boolean saveContent(final Context context, final String content) {
...
         }
 
         if (!testCreateParent()) {
             return false;
         }
-        shareUtil = shareUtil != null ? shareUtil : new ShareUtil(context);
 
         // Don't write same content if base file not changed
-        if (isContentSame(content) && _modTime >= lastModified()) {
+        if (_modTime >= lastModified() && isContentSame(content)) {
```
harshad1, line 248: Do cheap test first

app/src/main/java/net/gsantner/markor/model/Document.java
```diff
@@ -274,7 +281,11 @@ public synchronized boolean saveContent(final Context context, final String cont
 
         if (success) {
             setContentHash(content);
-            _modTime = lastModified();
+            final long curModTime = lastModified();
+            if (_modTime >= curModTime) {
+                Log.w("MARKOR_DOCUMENT", "File modification time unchanged after write");
```
harshad1, line 286: To aid debugging. I think we should leave this or something similar in there

app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
```diff
@@ -268,7 +268,7 @@ public static void showMultiChoiceDialogWithSearchFilterUI(final Activity activi
         // =========================================================================================
 
         // Ok button action
-        if ((dopt.callback != null) || (dopt.isMultiSelectEnabled)) {
+        if ((dopt.isSearchEnabled && dopt.callback != null) || (dopt.isMultiSelectEnabled && dopt.positionCallback != null)) {
```
harshad1, line 271: Unrelated  fix for crash in special keys dialog

app/src/main/java/net/gsantner/opoc/util/FileUtils.java
```diff
@@ -522,6 +524,18 @@ public static String sha512(final byte[] data) {
...
+        final CRC32 alg = new CRC32();
+        final int length = data.length();
+        for (int i = 0; i < length; i++) {
+            final char c = data.charAt(i);
+            // Upper and lower bytes
+            alg.update((byte) (c & 0xff));
+            alg.update((byte) (c >> 8));
+        }
+        return alg.getValue();
+    }
```
harshad1, line 537: Crc32 on charsequence. Avoids 2 copies to string and bytes[].

app/src/main/java/net/gsantner/opoc/util/ShareUtil.java
```diff
@@ -1093,10 +1093,10 @@ private Activity greedyGetActivity(final Activity... activity) {
      * Requires storage access framework permission for external storage (SD)
      *
      * @param file  The file object (file/folder)
-     * @param isDir Wether or not the given file parameter is a directory
      * @return Wether or not the file can be written
      */
-    public boolean canWriteFile(final File file, final boolean isDir) {
+    public boolean canWriteFile(final File file) {
+        final boolean isDir = file.isDirectory();
```
harshad1, line 1099: I think this change is ok

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
gsantner: Generally .. do you have issues with the "yellow warning"? Many people talk about getting it wrongly, but it just works fine to me.

app/src/main/java/net/gsantner/markor/model/Document.java
gsantner: hmm..might with some luck also resolve #1671

app/src/main/java/net/gsantner/opoc/util/ShareUtil.java
gsantner: hmm, it can produce wrong result. when previously isDir=true is passed, then it would be checked if the parentdirectory is writable.   
Generally this function has been written quite some time ago, might need some improvements/fixes - hope adding this not breaks additionally

app/src/main/java/net/gsantner/markor/model/Document.java
gsantner: OK to me to add more logs (albeit it won't be exporable by some in-app button, which is OK).  
Please use getClass().getName for the first parameter

app/src/main/java/net/gsantner/opoc/util/ShareUtil.java
harshad1: No, you're right. I will revert this change. No need to possibly introduce additional bugs if it isn't needed.

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -654,12 +667,28 @@ public String getFragmentTag() {
...
+
+            if (_shareUtil.isUnderStorageAccessFolder(file) && _shareUtil.getStorageAccessFrameworkTreeUri() == null) {
+                _shareUtil.showMountSdDialog(getActivity());
+            }
+
+            final boolean permok =  _document.testCreateParent() && _shareUtil.canWriteFile(file);
+            _textSdWarning.setVisibility(permok ? View.GONE : View.VISIBLE);
+            return permok;
+        }
+        return false;
```
harshad1, line 683: In this past (and in this PR too) we are checking permissions all the time instead of just once on the initial load. I guess the reason is that the filesytem could change while the document is open. Do you think it makes sense to keep this? Or is it sufficient to just display a warning to the user on  start and then not check?

harshad1, line 683: In this past (and in this PR too) we are checking permissions all the time instead of just once on the initial load. I guess the reason is that the filesytem could change while the document is open. Do you think it makes sense to keep this? Or is it sufficient to just display a warning to the user on  start and then not check?

harshad1:
Closing as this PR will be merged in smaller, more resonable chunks.

harshad1:
I've never seen it in practice. But I only edit files within my notebook and have a pretty standard device (pixel 3)

May 18, 2022 10:17:07 Gregor Santner ***@***.***>:

PR #1707, opened 2022-05-18
Save as New File, closes #760

### LINKED ISSUE #760: Feature: Save as New File
#### General information

* **App version:** v101-2.2.1 from github
* **System:** Android 6.0.1 Samsung

#### Description

Sometimes I need to _save_ a new **Quicknote{¿date?}.md** vice **appending** the older/current Quicknote.md but there's no easy way to accomplish that.    
So suggest you add an entry to the menu-pull-down as in screenshot:    
● Save As _(newfile)_ & also for **_translations_** be shure to distinguish it from    
● Save _(overwrite)_

#### Screenshot

Markor-SaveAs

2catycm:
Hello, this is a solution for Issue #760.
I added save as by using new file dialog.
Here is the running screenshot on my phone.
476A3135E949B004200C9DA6388BBA82
DC85121AC8EED9E2D31539522A7236E6

D6AEF06F53E2CE9B744F76D1CA9080F7

When you long click the save as button, it will tell you that it creates new file, but not save the current file, as @ChiefMikeK suggested. 

The previous PR has problem that localization of Chinese is not up to date. Now it is fixed.


app/src/main/res/values-zh-rCN/strings.xml
```diff
@@ -104,7 +104,14 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="history">历史</string>
     <string name="sync">同步</string>
     <string name="search_or_custom">搜索 / 自定义</string>
-    <string name="save">保存</string>
+
```
gsantner, line 107: translations are managed in Crowdin <https://crowdin.com/project/markor/invite>. Please don't add translated strings in pull requests. Instead add the (chinese) translation then when the PR is merged ... at Crowdin.

--> remove the both changes to zh strings.xml

thanks

harshad1:
Maybe just use 'confirm' and 'cancel' for the button strings?

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -671,6 +702,83 @@ public boolean saveDocument(final boolean forceSaveEmpty) {
...
+     * 3. When the user gets back to from the new document, he/she can continue to edit the old document in this fragment.
+     */
+    public boolean saveAsNewDocument() {
+        File currentFolder = new File(this._document.getPath()).getParentFile();
+        Log.d(getFragmentTag(), "saveAsNewDocument: find currentFolder is " + currentFolder);
+        if (currentFolder == null) {
+            return false;
+        }
+        NewFileDialog dialog =
+            NewFileDialog.newInstance(currentFolder.getAbsoluteFile(), false, (ok, newFile) -> {
```
harshad1, line 718: Can we just use SearchOrCustomTextDialog here? Will simplify things IMO

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
```diff
@@ -462,10 +462,10 @@ public Fragment getItem(int pos) {
                     return FilesystemViewerFragment.newInstance(getFilesystemFragmentOptions(null));
                 }
                 case R.id.nav_quicknote: {
-                    return DocumentEditFragment.newInstance(_appSettings.getQuickNoteFile(), -1);
+                    return DocumentEditFragment.newInstance(_appSettings.getQuickNoteFile(), -1, MainActivity.this);
                 }
                 case R.id.nav_todo: {
-                    return DocumentEditFragment.newInstance(_appSettings.getTodoFile(), -1);
+                    return DocumentEditFragment.newInstance(_appSettings.getTodoFile(), -1, MainActivity.this);
```
harshad1, line 468: Can't we just use getActivity() for the parent? We already use `getActivity() instanceof MainActivity` etc

app/src/main/res/values-zh-rCN/strings.xml
2catycm: ok

2catycm:
> Maybe just use 'confirm' and 'cancel' for the button strings?

you mean the save dialog before save as. Yes, it is much shorter.

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
2catycm: So I need to add a new method in the class SearchOrCustomTextDialogCreator to create a dialog for user to specify the name, right?

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
2catycm: > Can't we just use getActivity() for the parent? We already use `getActivity() instanceof MainActivity` etc

Thanks. I didn't know android has this method for Fragment before.
You are right, no need for a new field.

2catycm:
> Maybe just use 'confirm' and 'cancel' for the button strings?

I don't think `confirm` and `cancel` express well,  because we are doing `save as`, and if I click `cancel`, I would think that I am cancelling the `save as` since additional `save` was not scheduled by myself, but by the program. So in my opinion `save` and `don't save` may be better.

gsantner:
Its fine with cancel and  confirm/save/OK.

gsantner:
Its ok to have two words here.
But keep in mind, it will be translated. So rather shorter the better.

Ja, Ich möchte es speichern

2catycm:
> My general input is though, do we really need a own "keep old" dialog? Every editor I know keeps the old file "as is" and just switches the editor to the new (copied) file.

Yes, you mean  the new file opened has the new content, while the old file has the old content as it were not saved.

For example, the old file is 1.txt with content "123", 

we edit it to be "1234" which is unsaved.

If we then save it as 2.txt, 

most editors let 2.txt="1234", while 1.txt="123"

However, we have a onPause(), which will automatically save the 1.txt to be "1234"

gsantner:
In this case, just keep the old file "as is" (i.e. don't save/overwrite it with the current text). --> 1.txt stays with 123 and old timestamp. and 2.txt will be whatever is now in the editor, be it 123 (when user didn't change something), or 12345 because user added some new (unsaved yet) text.

2catycm:
Too many commits in this PR, I plans to open a new branch by git branch and wants to made that a PR.
https://github.com/Two-Cats-Software-Organization/markor/tree/Save_as_New_File

2catycm:
> In this case, just keep the old file "as is" (i.e. don't save/overwrite it with the current text). --> 1.txt stays with 123 and old timestamp. and 2.txt will be whatever is now in the editor, be it 123 (when user didn't change something), or 12345 because user added some new (unsaved yet) text.
> 
> --> So the code behind "No, Thanks" action I guess.

You are right, that is what most editors do. 
- The "No thanks" currently does an opposite operation, not as you descriped.
- The problem which stuckes me from implement that is I cannot figure out how to **open a new document in place**.
Why need to **open new document in place**
- You see, my current implementation invokes a new DocumentEditActivity on top of the old one, with new document in it.
- This makes the old activity paused, and it invokes onPause(), by which the old document will be save.
- What comes with this is that the old file is definitely saved, i.e., cannot be choose by user, unless modifying onPause()
- What "No thanks" do is however make the new file (the saved as file) to be "as is"(same as the old file content without new change). 
If I implement the open new document action in place, that is to say, same activity, leaving old file unsaved, open new file.   
I cannot figure out how to do that, 
- is there a way for `Android Activiy` to restart itself(with different fields, or, say, different intent)? 
- And the restart should not call onPause(). 
- or is there a way for just `DocumentActivity` `DocumentEditFragment` to load another file?

2catycm:
Alternatively, I came up with a new idea for the UI logic: 
- Why not just copy the content, but not open the newly created document?

In #760, @ChiefMikeK 's user story was that he wants to use QuickNote. One day he found that his quick note was very useful, so he wanted to save it as "My interesting note 1.md". What he would do next, I think, was to clear the contents in QuickNote, and write his next note. When he finished, he was to save the new quicknote as "My interesting note 2.md". 

So by saving as, maybe we don't need to open it. We just want to back it up, but works on the old file.

2catycm:
> Alternatively, I came up with a new idea for the UI logic:
> 
> * Why not just copy the content, but not open the newly created document?
> 
> In #760, @ChiefMikeK 's user story was that he wants to use QuickNote. One day he found that his quick note was very useful, so he wanted to save it as "My interesting note 1.md". What he would do next, I think, was to clear the contents in QuickNote, and write his next note. When he finished, he was to save the new quicknote as "My interesting note 2.md".
> 
> So by saving as, maybe we don't need to open it. We just want to back it up, but works on the old file.

I mean, if he were to edit the new file, he would probably not use Quicknote, but addes new file directly instead. So the `save as` in Quick note intents to save the note as file, while also keep editing quick note.

gsantner:
there is a close menu action, which is for the case to close without save. maybe that is helpful?

gsantner:
Closing due inactive, reopen when updated/ ready to check again.

2catycm:
got it.
By the way, how to add a new setting for markor. I think we can remember the choice made by the user.



---Original---
From: "Gregor ***@***.***&gt;
Date: Sat, May 21, 2022 16:30 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [gsantner/markor] Save as New File, closes #760 (PR #1707)




 
Its ok to have two words here.
 But keep in mind, it will be translated. So rather shorter the better.
 
Ja, Ich möchte es speichern

harshad1:
Note that I recently updated the reload button so it always fetches changes from disk (i.e. no checking timestamp etc)

So if you think you have changes in the current open document which should not be saved reload will revert all of them.

Of course, you need to reload before the changes are saved.

Issue #1566, opened 2022-01-17
per-file-settings: incl favorite state when renaming/moving a file


mbomb007:
## General information
* App version: 2.8.4
* Android version: 8.1
* Device: Samsung j3
* I understand that this issue will be closed without further notice in case I didn't describe the problem including general information, or didn't [search](https://github.com/gsantner/markor/issues?q=) for similar issues & [requests](https://github.com/gsantner/markor/discussions?discussions_q=).

### Description

Star a txt file
Rename it
The file icon is no longer yellow, and it needs to be restarred

harshad1:
Yeah, we store starred files by path. You'll notice that the file will become starred again if you rename it back.

The solution is to update the path in starred files when a file is renamed. Should probably update all per-file-option paths too.

Obviously this will not work if the file is updated outside Markor.

gsantner:
Yeah it's not only about starring. Any file, be it renamed outside markor or inside, is treated as new file, with new own settings. which also includes starred.

Twilight-Shuxin:
Hi @gsantner  I'd like to volunteer for this work if the change is preferred. Is it still open?

gsantner:
yes. but be said, it's a bit advanced topic.

Twilight-Shuxin:
Okay thanks, it would also be great if there could be any guide or suggestions. I'd try my best to understand the file system and solve this issue.

harshad1:
The filesystem is just a basic Android filesystem. We store a mapping of path -> setting for each file in a flat map.

What needs to be done is
1. Identify all functions where a file is changed (move, copy, delete)
2. Add code to delete or change the mapping key as appropriate.

More advanced things to look at (in the future)
- [ ] Option to reset settings for all files
- [ ] Delete settings for file if it is not present (can test all saved settings at shutdown or something)

harshad1:
Favourites are simply a list of paths which is parsed / updated whenever necessary.

Look at `AppSettings.java` for all of this

Twilight-Shuxin:
Thanks so much for your help, I think I'm getting to understand it now.

Twilight-Shuxin:
Hi, here's my understanding of this problem and I wonder if I had got it right: 

All favourite files are stored in a favourite file list maintained in AppSettings.java, and each file entry is identified by their absolute paths(which includes the file name itself). 
So to solve the issue, what needs to be done is to find all places where the absolute path of a file is to be changed and to update the favourite file list accordingly. 

[...]

PR #1708, opened 2022-05-20
Favourite tags to follow file changes #1566

Twilight-Shuxin:
This pull request includes modification for favourite tags to follow file changes(renaming, copying, deleting and moving). I have tested several scenarios on my own mobile phone as bellow:

- Rename a favourite file / folder, the new file / folder has same favourite status as before
- Move a favourite file / folder to another folder, the new file / folder has same favourite status as before
- Delete a favourite file / folder, later create file / folder with same name(would not show up as favourite)
- Copy a favourite file / folder to another folder, the new file / folder has same favourite status as before
(Also multiple selecting are put to test)

For these tests I recorded videos containing the testing process, but because tapping doesn't always show my finger movement so they look a bit confusing and I decided not to paste them here. 

I have two commits in this pull request, with the first commit I forgot to reformat the code(1476c1f), and with the second commit it seems to be changing many other un-modified files' format along(That's why there is 249 changed files.). I'm not sure whether it should be in this way, if any thing needs to be improved please inform me.


Twilight-Shuxin:
#1566

Twilight-Shuxin:
I think I understand where the problem is now... I reformatted xml files along and they shouldn't be. Sorry and I'll fix this issue.

gsantner:
reopened in #1709

PR #1709, opened 2022-05-20
Favourite tags to follow file changes #1566

Twilight-Shuxin:
This pull request includes modification for favourite tags to follow file changes(renaming, copying, deleting and moving).  #1566  I have tested several scenarios on my own mobile phone as bellow:

- Rename a favourite file / folder, the new file / folder has same favourite status as before
- Move a favourite file / folder to another folder, the new file / folder has same favourite status as before
- Delete a favourite file / folder, later create file / folder with same name(would not show up as favourite)
- Copy a favourite file / folder to another folder, the new file / folder has same favourite status as before
(Also multiple selecting are put to test)

For these tests I recorded videos containing the testing process, but because tapping doesn't always show my finger movement so they look a bit confusing and I decided not to paste them here. If any thing needs to be improved or explained please contact me.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/util/AppSettings.java
+++ b/app/src/main/java/net/gsantner/markor/util/AppSettings.java
@@ -350,7 +350,7 @@ public class AppSettings extends SharedPreferencesPropertyBackend {
 
     public void deleteFavouriteFile(File file) {
         List<String> list = new ArrayList<>();
-        List<File> favourites = getFavouriteFilesAll();
+        List<File> favourites = getFavouriteFiles();
         String abs = file.getAbsolutePath();
         for (File f : favourites) {
             list.add(f.getAbsolutePath());
@@ -364,7 +364,7 @@ public class AppSettings extends SharedPreferencesPropertyBackend {
     public void renameFavoriteFile(File file, File newFile) {
         List<String> listBefore = new ArrayList<>();
         List<String> listAfter = new ArrayList<>();
-        List<File> favourites = getFavouriteFilesAll();
+        List<File> favourites = getFavouriteFiles();
         String abs = file.getAbsolutePath();
         for (File f : favourites) {
             listBefore.add(f.getAbsolutePath());
@@ -592,17 +592,6 @@ public class AppSettings extends SharedPreferencesPropertyBackend {
     }
 
     public ArrayList<File> getFavouriteFiles() {
-        ArrayList<File> list = new ArrayList<>();
-        for (String fp : getStringList(R.string.pref_key__favourite_files)) {
-            File f = new File(fp);
-            if (f.exists() || FilesystemViewerAdapter.isVirtualStorage(f)) {
-                list.add(f);
-            }
-        }
-        return list;
-    }
-
-    public ArrayList<File> getFavouriteFilesAll() {
         ArrayList<File> list = new ArrayList<>();
         for (String fp : getStringList(R.string.pref_key__favourite_files)) {
             File f = new File(fp);
--- a/app/src/main/java/net/gsantner/markor/util/AppSettings.java
+++ b/app/src/main/java/net/gsantner/markor/util/AppSettings.java
@@ -325,10 +325,69 @@ public class AppSettings extends SharedPreferencesPropertyBackend {
         ShortcutUtils.setShortcuts(_context);
     }
 
+    public void copyFavouriteFile(File fileBefore, File fileAfter) {
+        List<String> list = new ArrayList<>();
+        List<File> favourites = getFavouriteFiles();
+        for (File f : favourites) {
+            // Only existing files or file which is 'virtual storage' is considered
+            if (f != null && (f.exists() || FilesystemViewerAdapter.isVirtualStorage(f))) {
+                list.add(f.getAbsolutePath());
+            }
+        }
+        String abs = fileBefore.getAbsolutePath();
+        boolean starred = false;
+        if (list.contains(abs)) {
+            starred = true;
+        }
+        if (starred) {
+            abs = fileAfter.getAbsolutePath();
+            if (!list.contains(abs)) {
+                list.add(abs);
+            }
+        }
+        setStringList(R.string.pref_key__favourite_files, list);
+    }
+
+    public void deleteFavouriteFile(File file) {
+        List<String> list = new ArrayList<>();
+        List<File> favourites = getFavouriteFilesAll();
+        String abs = file.getAbsolutePath();
+        for (File f : favourites) {
+            list.add(f.getAbsolutePath());
+        }
+        if (list.contains(abs)) {
+            list.remove(abs);
+        }
+        setStringList(R.string.pref_key__favourite_files, list);
+    }
+
+    public void renameFavoriteFile(File file, File newFile) {
+        List<String> listBefore = new ArrayList<>();
+        List<String> listAfter = new ArrayList<>();
+        List<File> favourites = getFavouriteFilesAll();
+        String abs = file.getAbsolutePath();
+        for (File f : favourites) {
+            listBefore.add(f.getAbsolutePath());
+            listAfter.add(f.getAbsolutePath());
+        }
+        String fileAddressBefore = file.getAbsolutePath();
+        String fileAddressAfter = newFile.getAbsolutePath();
+        for (String fileAddress : listBefore) {
+            if (fileAddress.startsWith(fileAddressBefore)) {
+                String newFileAddress = fileAddressAfter + fileAddress.substring(fileAddressBefore.length());
+                listAfter.remove(fileAddress);
+                listAfter.add(newFileAddress);
+            }
+        }
+        setStringList(R.string.pref_key__favourite_files, listAfter);
+    }
+
+    // getFavouriteFiles: contains all files marked as favourite
     public void toggleFavouriteFile(File file) {
         List<String> list = new ArrayList<>();
         List<File> favourites = getFavouriteFiles();
         for (File f : favourites) {
+            // Only existing files or file which is 'virtual storage' is considered
             if (f != null && (f.exists() || FilesystemViewerAdapter.isVirtualStorage(f))) {
                 list.add(f.getAbsolutePath());
             }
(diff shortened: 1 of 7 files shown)


app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -325,10 +325,69 @@ public void addRecentDocument(File file) {
...
+            if (f != null && (f.exists() || FilesystemViewerAdapter.isVirtualStorage(f))) {
+                list.add(f.getAbsolutePath());
+            }
+        }
+        String abs = fileBefore.getAbsolutePath();
+        boolean starred = false;
+        if (list.contains(abs)) {
+            starred = true;
+        }
+        if (starred) {
```
gsantner, line 342:
```suggestion
        if (list.contains(abs)) {
```

app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -325,10 +325,69 @@ public void addRecentDocument(File file) {
...
+            abs = fileAfter.getAbsolutePath();
+            if (!list.contains(abs)) {
+                list.add(abs);
+            }
+        }
+        setStringList(R.string.pref_key__favourite_files, list);
+    }
+
+    public void deleteFavouriteFile(File file) {
```
gsantner, line 351: the get favourites function automatically scans all favorites for existance (as files can be deleted also in other apps without Markor). So we don't need a function for this

app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -325,10 +325,69 @@ public void addRecentDocument(File file) {
...
+        for (File f : favourites) {
+            list.add(f.getAbsolutePath());
+        }
+        if (list.contains(abs)) {
+            list.remove(abs);
+        }
+        setStringList(R.string.pref_key__favourite_files, list);
+    }
+
+    public void renameFavoriteFile(File file, File newFile) {
```
gsantner, line 364: rename is a combination of copy (new file) and delete. So I think it's not needed to have a function for that

Right?

app/src/main/java/net/gsantner/markor/util/AppSettings.java
```diff
@@ -543,6 +602,15 @@ public ArrayList<File> getFavouriteFiles() {
         return list;
     }
 
+    public ArrayList<File> getFavouriteFilesAll() {
```
gsantner, line 605: add a bool parameter to the existing favourites  -->

`if (checkExists && (exist || virtual))`

app/src/main/java/net/gsantner/markor/util/AppSettings.java
Twilight-Shuxin: The part dealing with rename is quite different from copy & delete. When copying & deleting, the implementation in original code would search down all sub-directories, while with renaming some times only a single method  provided by andorid library .renameTo is called upon. So with copying & deleting, each time a file's change is applied the function is called upon and that single file's favourite status is changed, but with renaming each time the renameFavourite function is called upon it might be dealing with a set of files(including all files in subdirectory which were starred before).

app/src/main/java/net/gsantner/markor/util/AppSettings.java
Twilight-Shuxin: The reason I wrote this method is I wished only to modify status after files are successfully deleted(since deletions culd fail and in such cases I think it might be better to have the favourites un-modified). So after deletion, this method is called upon to deal with those files still recorded as favourite but is already deleted.

app/src/main/java/net/gsantner/markor/util/AppSettings.java
Twilight-Shuxin: This function performs for the purpose of noticing those files which are still recorded but is actually deleted already.

gsantner: yes, but still: It's the same like the normal `getFavouriteFiles(`, except the `if` condition.

Twilight-Shuxin: I merged two into one and deleted the `if` statement, because I think if code runs correctly, when the function is called in other situations it shouldn't contain file paths which does not meet the constraints. Still, if adding a parameter would be better, that also seems pretty good to me.

gsantner: > I think if code runs correctly, when the function is called in other situations it shouldn't contain file paths which does not meet the constraints.


NO! You are wrong.   Users can also delete file outside of Markor. Even automatically, some synchronisation software could delete the file. So just thinking `ok we remove it from the list when we delete it inside Markor` is not enough.

gsantner: > I think if code runs correctly, when the function is called in other situations it shouldn't contain file paths which does not meet the constraints.


NO! You are wrong.   Users can also delete file outside of Markor. Even automatically, some synchronisation software could delete the file. So just thinking `ok we remove it from the list when we delete it inside Markor` is not enough. There MUST be validation check for file existing - at time of loading the favourites. This is what the if condition is about.

Twilight-Shuxin: Oh, I see. I understand it now, I'll do the updates.

Twilight-Shuxin: There are still a few places unchanged, I'll solve them a bit later.

Twilight-Shuxin: > There are still a few places unchanged, I'll solve them a bit later.

Now the files are up to date and several tests are done.

app/thirdparty/java/other/writeily/model/WrMarkorSingleton.java
```diff
@@ -66,60 +67,66 @@ private boolean saneMove(final File file, final File dest) {
...
         }
         return false;
     }
 
-    public boolean deleteFile(final File file, final Context context) {
+    public boolean deleteFile(final File file, final Context context, AppSettings appSettings) {
         if (file.isDirectory()) {
             for (final File childFile : file.listFiles()) {
-                deleteFile(childFile, context);
+                deleteFile(childFile, context, appSettings);
```
harshad1, line 102: I would suggest not coupling move and delete to an instance of appSettings. These are more like general utility functions. I think we should instead call delete -> check success -> remove from favorites from the appropriate call sites.

Twilight-Shuxin: Yes, I also think that coupling it here is not such a good choice. I'll do the updates.

gsantner: as said earlier: you don't need to remove something from favorites at all. we don't need that explicit removal. The getter for favorites automatically does (& has to) filter out all non existing favorites. So a new call to the getter is all you should need. And that once.

Twilight-Shuxin: That's the previous design, and it cause a behavior which seems like a bug to me: favorite a file, delete it, and create that same file again, the file would show up as favorite. Is that a desired behavior?

gsantner: then there is a issue in that function. it should drop that favorite.

though we also have these things all over the place. all per-file settings are tied to specific files, and they are not resetted/dropped upon file removal. so not sure if generally the SharedPreference way with filename as key is the holy grail ^^. (speaking about things like atuo-format, font-size, ..)

harshad1:
We can add a cleanup system on app start. Check if all files exist and delete settings if they don't. To be more conservative we can do something like:

1. Mark settings as being dirty (no corresponding file) with a timestamp
2. If any dirty timestamps are > 1 month old, delete the setting



This will prevent us from loosing a setting if a file is accidentally deleted and restored.



On the other hand, this is a pretty low priority problem, imo 





---- On Tue, 24 May 2022 14:06:29 -0700 Gregor Santner ***@***.***> wrote ----





@gsantner commented on this pull request.


In https://github.com/gsantner/markor/pull/1709#discussion_r880950567:

>          if (file.isDirectory()) {
             for (final File childFile : file.listFiles()) {
-                deleteFile(childFile, context);
+                deleteFile(childFile, context, appSettings);


then there is a issue in that function. it should drop that favorite.

though we also have these things all over the place. all per-file settings are tied to specific files, and they are not resetted/dropped upon file removal. so not sure if generally the SharedPreference way with filename as key is the holy grail ^^. (speaking about things like atuo-format, font-size, ..)

Twilight-Shuxin:
I thinked on this problem, and I think that a better approach to this is to maintain 'keys'(things like generated hash numbers) for file settings instead of binding them with file names. That is, a file name to key map is maintained, and file settings are stored with keys. 

Suppose current file name is `currentFileName`, and corresponding key is `key` and when a file is moved or renamed, simply delete the mapping from `currentFileName - key` and change it to `newFileName - key`, in this case all previous file settings would be binded together with the new file. 

When a file is deleted, do nothing. When a new file is created, create a new binding key and set all settings to default, binded with the new key.

When a file is copied, the new file gets a new key and the settings of the new key inherits all settings of the previous key.

Under this design, only with coping and creating new files we actually need to deal with the settings, which could be binded into a same function with one parameter different (whether with inherit key or not, if not then all goes to default).

Out of time keys with their settings could be deleted in the way @harshad1 suggested.

However, this is only a logical design which I'm not sure whether it might fit in well with the original code system(like if file settings are widely spreaded out, it might be causing tons of coupling). Also it would do some big changes, and it's to be considered whether to take the change or not.

gsantner:
So it's inactive a while, and I will close the merge request. I'm thanking everybody helping here, but have to also admit I don't want to include it in this way.

Specifically, a file can be moved by any other app around, where we don't know about what was the old place. Or if it's just some copy? We don't know, we can't know. While we do know about this when we move/rename/duplicate files in Markor, yes we can act differently here. But I rather prefer to behave same like when you would move externally - treat things like a different file, and don't migrate file specific configuration.

Thanks for understanding.

PR #1719, opened 2022-05-25
Improve file handling in a few places

harshad1:
This PR includes the following changes:

1. Some improvements to the document class and how hashes are computed
2. Changing the file info return from a map to an actual object. IMO Map<String, Object> is basically a python object. May as well use an _actual_ struct and get all the benefits
3. Several cleanups


app/src/main/java/net/gsantner/markor/model/Document.java
```diff
@@ -156,11 +155,11 @@ public boolean isEncrypted() {
 
     private void setContentHash(final CharSequence s) {
         _lastLength = s.length();
-        _lastHash = FileUtils.crc32(s.toString().getBytes());
```
harshad1, line 159: crc32 directly on charsequence. No more creating 2x copies

harshad1:
@tifish You may want to look at these changes too

gsantner:
related #1693

PR #1721, opened 2022-05-26
WIP: Debounced the change listener

harshad1:
Adds a generic denouncing routine and uses it for the is file changed listener.

In the future I want to use this to dramatically simplify the debouncing in HighlightingEditor. But that is a bigger change

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
@@ -53,6 +53,7 @@ import net.gsantner.markor.util.ContextUtils;
 import net.gsantner.markor.util.MarkorWebViewClient;
 import net.gsantner.markor.util.ShareUtil;
 import net.gsantner.opoc.activity.GsFragmentBase;
+import net.gsantner.opoc.android.dummy.TextWatcherDummy;
 import net.gsantner.opoc.preference.FontPreferenceCompat;
 import net.gsantner.opoc.ui.FilesystemViewerData;
 import net.gsantner.opoc.util.ActivityUtils;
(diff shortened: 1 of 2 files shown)


app/src/main/java/net/gsantner/opoc/util/StringUtils.java
```diff
@@ -690,4 +692,20 @@ public CharSequence subSequence(int start, int end) {
             return select().subSequence(start, end);
         }
     }
+
+    public static Runnable makeDebounced(final long delayMs, final Runnable callback) {
```
harshad1, line 696: This should probably not be in `StringUtils`. But I can't think of a better place to put it.

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -658,7 +658,7 @@ public String getFragmentTag() {
     // Only supports java.io.File. TODO: Android Content
     public boolean saveDocument(final boolean forceSaveEmpty) {
         // Document is written iff content has changed
-        if (_isTextChanged && _document != null && _hlEditor != null && isAdded()) {
+        if (_document != null && _hlEditor != null && isAdded()) {
```
harshad1, line 661: Don't rely on external isTextChanged as we don't test this with every change. Save will check this internally anyway.

harshad1:
@gsantner this and the other 2 PRs I have are ready for merging. (There may be a conflict between this one and #1719 ). Will resolve after one is merged.

gsantner:
single pr would have been fine to me, I just didnt have time yet

harshad1:
I can combine them tonight, if that's easier. Broke up the larger PR for easier testing and review.

May 26, 2022 12:12:11 Gregor Santner ***@***.***>:

gsantner:
No no, it's OKm

harshad1:
Found a bug. Marking my PRs WIP until I find and address it

Issue #1247, opened 2021-02-19
Full Screen Display, hide SystemUI and actionBar


liujin009:
#### General information

* **App version:** 
* **System:** 

#### Description
**I want to read in watch mode for Full Screen 
hide SystemUI and actionBar**

gsantner:
toolbar (top menu) and text actions (bottom menu) will stay in any way. but for system UI (status bar, navigation bar with back button) I would like to add an option. At same time get rid of the existing "hide statusbar" settings options, and just apply the fullscreen everywhere at Markor.

zhangliyuSustech:
I have made the Full screen, but I want know that where do you want to set the label. I think I can set the label in the Tools, or somewhere that you think is proper? @gsantner

gsantner:
in the various settings, below the dont turn screen off / disallow screenshots option.

PR #1722, opened 2022-05-28
Fullscreen settings option (Hide SystemUI status/nav) #1247

zly7:
#1247 

A simple PR for full screen, I will continously understand the  file setting, I still can't find a good way to change it. But current scheme is OK.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
@@ -521,11 +521,13 @@ public class DocumentEditFragment extends GsFragmentBase implements TextFormat.T
             }
             case R.id.action_full_screen:{
                 _isFullScreen = ! _isFullScreen;
-                DocumentActivity temp = (DocumentActivity) activity;
-                if(_isFullScreen){
-                    temp.hideSystemUI();
-                }else {
-                    temp.showSystemUI();
+                MarkorBaseActivity temp = (MarkorBaseActivity) activity;
+                if (temp!=null) {
+                    if (_isFullScreen) {
+                        temp.hideSystemUI();
+                    } else {
+                        temp.showSystemUI();
+                    }
                 }
 
             }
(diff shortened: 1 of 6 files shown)


app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -518,6 +519,16 @@ public void onFsViewerConfig(FilesystemViewerData.Options dopt) {
                 CoolExperimentalStuff.showSpeedReadDialog(activity, _hlEditor.getText().toString());
                 return true;
             }
+            case R.id.action_full_screen:{
+                _isFullScreen = ! _isFullScreen;
+                DocumentActivity temp = (DocumentActivity) activity;
```
harshad1, line 524: activity == MainActivity and not DocumentActivity in the quicknote and todo tabs. This will error out if you don't check that here. 

The option should be disabled in main activity.

zhangliyuSustech: Thanks for pointing out that! Is there are any other faults like this, I will revise it.

gsantner:
So this is rather something to manually toggle to fullscreen on/off currently, rather than some setting. What I think people look for is a saved setting -->

* AppSettings.java
* preferences_master.xml
* Activity::onResume

agree  @harshad1 ?

harshad1:
We already have a setting to disable the top bar. Problem is that it doesn't work well on many phones (notch etc).

Would need to investigate / think about this more.

May 28, 2022 10:38:02 Gregor Santner ***@***.***>:

> So this is rather something to manually toggle to fullscreen on/off currently, rather than some setting. What I think people look for is a saved setting -->
> 
* > AppSettings.java
* > preferences_master.xml
* > Activity::onResume

d-yahoo (commented):
Jj

gsantner:
Thanks for the PR. I'm closing it now due both inactivity and have something else in mind: To make fullscreen toggling soley a action in bottom or top bar with a button, but not a settings option. 

Thanks for your time though still!

ryanwwest:
@gsantner Any idea on when the fullscreen toggle idea you have might be added? It would be really nice to have, especially since when typing with the on-screen keyboard the bottom navigation bar hides an additional 4+ lines of otherwise readable text. My use case is that I only ever use Markor on the QuickNote tab, so I'd like a one-click solution from the home screen to open the app in QuickNote tab to quickly jot something down.

(As a side note, when I open the app this way (in edit mode), the cursor always starts at the bottom of my specified quick note no matter if Settings -> Edit Mode -> Start on bottom is enabled or disabled. I wish it could start at the top. But I'll see if there's another bug issue for this already).

PR #1724, opened 2022-05-28
add new feature of H1,H2,H3

zhangliyuSustech:
#1723


gsantner:
Draw by SVG? I don't think you need any graphics here :smile:

harshad1:
Can you explain what you're trying to achieve here?

PR #1727, opened 2022-05-28
[spam] solve QuickNote disappear problem

jayunx:
Maybe because of the problem of memory structure, I think a judgment mechanism can be added before exiting.
Hello author, our pr is a test, so please ignore our pr, we are very sorry for that。


gsantner:
If you want to make dummy pull requests, make your own repositories & or forks. There you can do whatever you want.  
Markor is not a dummy project for dummy PRs. If you really want to improve something, please do that and work on real things. Don't do such meaningless pull requests.  
Communicate that also to your other project team(s).

Thanks.

PR #1728, opened 2022-05-28
Just try to use another colorPicker

Fantasy-Radix:
Hello developer, I tried to add another color picker preference for Android that ”enables selecting colors using sliders or hexadecimal color codes“ from github, here is its URL(https://github.com/peterfajdiga/ColorPickerHex), do you mind using another project in your project?
PS: Maybe this small change is not very useful TAT


gsantner:
Yes another library is a possibility. But just adding the library is not enough, it needs to be used. So dropping the old library and replacing the use with the new library.

gsantner:
reopen when the new color picker is available to try/test/use in the app and ready to for a review 

thanks

PR #1733, opened 2022-05-29
Basic Draft for 'View Mode: Save & Restore scroll position'

CutieDeng:
This is a simple demo for this problem. 
After this step, the remain steps are below: 
- Inbuild a micro database system in the app. 
- Fetch the related value, according to the SHA-1 key or file name key. (If able, the extra line information can be supported. )
- Fix the proper location for the file which had been used in our app. 

--- 

It's the trivial and naive solution for this issue in my view. Can we have some more optimizations?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,4 @@ app/flavor*
 # Avoid ignoring Gradle wrapper jar file (.jar files are usually ignored)
 !gradle-wrapper.jar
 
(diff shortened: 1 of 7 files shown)


CutieDeng:
It's committed for the issue #1025.

gsantner:
reopen when there is something minimal available. currently its only 2 code comments.

PR #1735, opened 2022-05-31
WIP: Performance improvements

harshad1:
This branch has some performance improvements for regexReplace and the chunked change system. 

Objective is to make a actions take < 50ms on a 100k character file with highlighting enabled (my monthly log / journal files regularly reach 70k chars). We aren't quite there, but this has brought us from 120ms to 80ms on my pixel 3

PR #1736, opened 2022-05-31
More deterministic save / resume

harshad1:
Calling save and resume explicitly on tab changes


harshad1:
@gsantner I'll clean up #1735 after this PR and #1719 are merged in

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -214,45 +213,58 @@ public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
...
-                startPos = _hlEditor.length();
             }
         }
 
         StringUtils.setSelectionAndShow(_hlEditor, startPos);
     }
 
+    public void resume() {
+        loadDocument();
+    }
```
gsantner, line 239: I have a plan/idea for things like resume/pause, but OK for now

gsantner:
when working on the Custom templates issue I noticed that UTF-8 with bom doesn't really work on the newfile dialog - always see a empty document created then. when choosing same template with bom checked off it works.

(Not sure if realted to the file write PRs)

PR #1751, opened 2022-07-07
Bugfix - missing ok button in dialogs

harshad1:
This fixes a bug I introduced recently where the ok button would not show up on dialogs simply asking for ok / cancel on a message

### DIFF
--- a/app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
+++ b/app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
@@ -268,7 +268,10 @@ public class SearchOrCustomTextDialog {
         // =========================================================================================
 
         // Ok button only present under these circumstances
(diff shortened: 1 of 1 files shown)


app/src/main/java/net/gsantner/opoc/ui/SearchOrCustomTextDialog.java
```diff
@@ -268,7 +268,10 @@ public static void showMultiChoiceDialogWithSearchFilterUI(final Activity activi
         // =========================================================================================
 
         // Ok button only present under these circumstances
-        if (((dopt.callback != null && dopt.isSearchEnabled) || (dopt.positionCallback != null && dopt.isMultiSelectEnabled))) {
+        final boolean isSearchOk = dopt.callback != null && dopt.isSearchEnabled;
+        final boolean isMultiSelOk = dopt.positionCallback != null && dopt.isMultiSelectEnabled;
+        final boolean isPlainDialog = dopt.callback != null && (dopt.data == null || dopt.data.size() == 0);
```
gsantner, line 273: similar to the filemanager button calculations. very good improvement and to understand when reading it

gsantner (approved):

PR #1752, opened 2022-07-07
Tweaked multi-select neutral button naming

harshad1:
I think this PR helps a little, opinions would be appreciated though :)

Have nuked the translations too.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -444,8 +444,8 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="due_today">Due today</string>
     <string name="due_future">Due in future</string>
     <string name="completed">Completed</string>
-    <string name="match_any">Match Any</string>
-    <string name="match_all">Match All</string>
+    <string name="matching_any">Matching Any</string>
+    <string name="matching_all">Matching All</string>
     <string name="dynamic_notebook_root_links">Dynamic search for notebook root</string>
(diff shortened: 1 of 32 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -444,8 +444,8 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="due_today">Due today</string>
     <string name="due_future">Due in future</string>
     <string name="completed">Completed</string>
-    <string name="match_any">Match Any</string>
-    <string name="match_all">Match All</string>
+    <string name="matching_any">Matching Any</string>
+    <string name="matching_all">Matching All</string>
```
harshad1, line 448: made is clearer that the _current_ state is matching all / any vs the state after you click the button

gsantner: Hm, depending on the language the translation can be even the same at both. For the english original I feel that the active (Match) is better than the passive (Matching), but that's a persona feeling.

harshad1: Hmmm. Perhaps we can use add a unicode check box?

`☐ Match all` vs `🗹 Match all`

harshad1: Ideally it would be an actual checkbutton, but modifying the Alerert dialog to achieve this seems too much work..

gsantner: no keep as string. active prefered. unicode might be a option, but not sure how it works out

harshad1:
Lets leave this for now. It is not really a bug and I need to think more about how / whether we want to do this.

PR #1757, opened 2022-07-11
Partial highlighting

harshad1:
This PR adds partial highlighting support discussed in <https://github.com/gsantner/markor/discussions/1749>

1. When text is changed we recompute the spans for the _whole_ text
2. When we scroll, we apply the highlighting for the visible region +- height

This works rn but lots of work remains to be done

- [x] Documentation / docstrings
- [x] Several spans were disabled - these need to be fixed and re-enabled
- [ ] Lots and lots of testing (Will need help testing Zim, especially, as I don't use the zim format)
- [ ] I removed all the profiling code to reduce clutter. Need to make a call and all it all back.


harshad1:
@gsantner given the scope of this change, I think it may be best to hold off until a future release - I will want to use it daily for several weeks before I am happy.

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownHighlighter.java
```diff
@@ -32,8 +30,8 @@ public class MarkdownHighlighter extends Highlighter {
     private static final int MD_COLOR_QUOTE = 0xff88b04c;
     private static final int MD_COLOR_CODEBLOCK = 0xff8c8c8c;
 
-    public MarkdownHighlighter(HighlightingEditor hlEditor, Document document) {
```
harshad1, line 35: Most of the derived classes have been simplified a _lot_

app/src/main/java/net/gsantner/markor/ui/hleditor/HighlightingEditor.java
```diff
@@ -10,116 +10,116 @@
...
 
-import java.io.File;
 import java.util.HashSet;
 import java.util.Set;
 
 @SuppressWarnings("UnusedReturnValue")
-public class HighlightingEditor extends AppCompatEditText {
+public class HighlightingEditor extends AppCompatEditText implements ViewTreeObserver.OnScrollChangedListener, TextWatcher {
+
+    final static float SIGNIFICANT_SHIFT_FRACTION = 0.2f;
```
gsantner, line 38: is this scroll screen area treshhold multiplier?

harshad1: Not quite. Screen area threshold is set to +-1 right now. This is a threshold s.t. the user has to scroll > 0.2x the currently highlighted region to trigger a reapplication. Prevents reapplication of highlight on small jumps and scrolls as you type.

app/src/main/java/net/gsantner/opoc/util/TextViewUndoRedo.java
```diff
@@ -424,11 +424,12 @@ public void onTextChanged(CharSequence s, int start, int before,
                 return;
             }
 
-            mAfterChange = s.subSequence(start, start + count);
+            mAfterChange = StringUtils.toString(s, start, start + count);
```
harshad1, line 427: Prevent these from storing spans

harshad1:
Editing / scrolling in the main activity is significantly slower than in the document activity. I think this may be related to usage of RelativeLayout - keeps requiring redraws. We should look at alternative ways to getting the same UI.

https://www.androidcode.ninja/solved-android-edittext-lag/

app/src/main/java/net/gsantner/markor/ui/hleditor/HighlightingEditor.java
harshad1: Note: I have updated this now. Still experimenting to find good thresholds. Also need to apply a small file optimization where we don't do the dynamic highlighting for files under, say, 25k characters.

harshad1:
@gsantner On my device, I can now edit a 700kb file with highlighting with no trouble.

gsantner:
> RelativeLayout slow

Suggesting ConstraintLayout, or check something else if it has the same performance hit like RelativeLayout.

app/src/main/java/net/gsantner/markor/ui/hleditor/HighlightingEditor.java
gsantner: How do you feel about +-2  x screensize?

harshad1:
@gsantner Yes. I will make a layout change in a separate PR, if it still seems necessary after I am done with this pr.

gsantner:
I can check as well

harshad1:
Quick update: this effort may not work out after all.

The issues I can't seem to get around satisfactorily are:
1. Adding / removing spans from a SpannableStringBuilder is very slow. On a file with many spans per line (especially todo.txt) scrolling becomes jerky
    - 150ms to add 1000 spans. Each line in a todo.txt file can have 10+ spans.
3. Adding / removing metric affecting spans causes the edit text to jump to the cursor. I have some workarounds, but these are janky,
4. Scrolling with metric affecting spans sometimes causes rendering issues.

On the bright side, this works amazingly well with a very large file with just 1 - 2 spans per line.

gsantner:
interesting issue I found on androidX migration - it can happen that a final initalized member is still null, while it's explicitley set a value in at top. This is because in this case some stuff in constructor already does some calls, while the object isn't fully ready yet, but calling derived methods. 

In this case, overridden method addTextChangedListener method was crashing :smile: .

Screenshot_20220722-232649


Pushed a guard for that on master

harshad1:
Lol I noticed this last night. I just pushed changes to nuke the activelisteners system and to just use a flag :) I will resolve the differences with master.

gsantner:
Oh ok, thought it's somehow related to my androidx changes the behaviour there

gsantner:
One thing you wanted to know is whether androidX brings notable general performance improvement on Editor.

I migrated to androidX and updated all required libs/code on the androidsdk-upgrade branch. You can try that out, or merge both (+partial) together in some new local branch to try it combined :D.

harshad1:
> I can check as well

Please let me know if you are able to reduce the layout nesting (I guess using relativelayout). The bottleneck preventing this pr from working fully is DynamicLayout reflow being slow. I think reducing the layout complexity may help.

gsantner:
Which ones are the important ones? From above I read it's mainly at MainActivity (which is bad, but enough when it works fast at opened document window).

Or is it also DocumentActivity/DocumentFragment?

harshad1:
I think we need to experiment with both (main activity should definately move away from relativelayout though) . I'll clean up the PR tonight so it is in a stable state. Then I will try it with alternative Layouts and see if the performance improves to an acceptable level.

Jul. 23, 2022 10:22:08 Gregor Santner ***@***.***>:

gsantner:
merged some layout improvements, they are on master branch

gsantner:
What I disabled by default at some point was the "bigger headings" feature, because its resource intensive.

With partial highliting, do you think its safe to enable it by default again?

harshad1:
I am still not sure partial highlighting will actually work. Still trying to solve some performance and tearing issues.

If it does work then the bigger headings can probably be default - enabled again. I will test this.

Jul. 26, 2022 16:22:13 Gregor Santner ***@***.***>:

gsantner:
Either way, thank you for the efforts

app/src/main/java/net/gsantner/markor/ui/hleditor/Highlighter.java
```diff
@@ -217,24 +345,134 @@ public void draw(@NonNull Canvas canvas, CharSequence text, int start, int end,
...
+        createSpanForMatches(Patterns.WEB_URL, new HighlightSpan().setForeColor(0xff1ea3fd).setItalic(true).setTextSize(_textSize * 0.85f));
     }
 
-}
+    protected final void createUnderlineHexColorsSpans() {
+        createColoredUnderlineSpanForMatches(HEX_CODE_UNDERLINE_PATTERN, m -> new ColorUnderlineSpan(Color.parseColor(m.group(1)), 3f), 1);
+    }
+
+    // We _do not_ implement UpdateLayout or Parcelable for performance reasons
+    public static class HighlightSpan extends CharacterStyle implements UpdateAppearance, Callback.r1<Object, Matcher> {
```
harshad1, line 383: Between this span class and the layout changes, the performance issues have been solved. Now I need to figure out how to solve the jumping and tearing issues in todo.txt

gsantner:
were my few changes helpful? I guess it didn't make a dramatic difference

app/src/main/java/net/gsantner/markor/ui/hleditor/Highlighter.java
gsantner: guess the jumping is due the border-divider-as-a-span, and them popping up and away?

harshad1: > guess the jumping is due the border-divider-as-a-span, and them popping up and going away? Thats generally a so-far-nice-working-hack on it's own, but don't know other way to insert dividers in a better fashion.

Yes, exactly. It will happen to a lesser degree with larger headings in markdown too. I am working on measuring and restoring scroll position relative to text when we re-highlight.

harshad1:
> were my few changes helpful? I guess it didn't make a dramatic difference

I think so. 10% or so, especially in main activity. Kinda hard to tell as I am making a large number of changes. If they are not disrupting anything else, I think we should keep them.

app/src/main/java/net/gsantner/markor/ui/hleditor/Highlighter.java
gsantner: when you dont get far this way, an alternative is to exclude the dividers from the range/screen-height stuff

harshad1:
Yeah. I tried separating static and dynamic spans at first, but it was too slow. Will try again now that we've improved the performance.

Jul. 30, 2022 02:10:42 Gregor Santner ***@***.***>:

harshad1:
Ok. I think I have solved both the tearing and the jump issues - complexity is getting high though :)

Now this requires much more testing and some more perf improvements before we can merge.

harshad1:
@gsantner This is now done.

What is left is:

- [ ] Add profiling code back if really needed
- [ ] Cleanups
- [ ] Documentation / comments
- [ ] More testing, especially on Zim (which I don't use)

I added an option to disable dynamic highlighting to settings (default enabled)

app/src/main/java/net/gsantner/markor/format/ListHandler.java
```diff
@@ -40,15 +40,13 @@ public void onTextChanged(final CharSequence s, final int start, final int befor
         // Detects if enter pressed on empty list (correctly handles indent) and marks line for deletion.
         if (beforeLineEnd != null && count > 0 && start > -1 && start < s.length() && s.charAt(start) == '\n') {
 
-            final Spannable sSpan = (Spannable) s;
-
             final AutoFormatter.OrderedListLine oMatch = new AutoFormatter.OrderedListLine(s, start, _prefixPatterns);
             final AutoFormatter.UnOrderedOrCheckListLine uMatch = new AutoFormatter.UnOrderedOrCheckListLine(s, start, _prefixPatterns);
 
             if (oMatch.isOrderedList && beforeLineEnd == oMatch.groupEnd) {
-                sSpan.setSpan(this, oMatch.lineStart, oMatch.lineEnd + 1, Spanned.SPAN_COMPOSING);
```
harshad1, line 49: Don't use spans for deletion - too expensive

gsantner:
@harshad1  
So Google Play gave green light with Markor file access now, with the 2.10.1 release, i.e. it passed review --- with the androidsdk-upgrade branch containing AndroidX.  
As consequence I merged that merge request so everything is one code base. (-> some merge conflicts related to androidx renamings in your PR)

In case you encounter issues, please let me know. I just got a mail today saying the new update is very bad and it worked better before...but without saying what is wrong...so I don't know whats up 😄 .

app/thirdparty/java/other/writeily/widget/WrMarkorWidgetProvider.java
```diff
@@ -48,38 +48,43 @@ public void onUpdate(Context context, AppWidgetManager appWidgetManager, int[] a
 
             views.setTextViewText(R.id.widget_header_title, FileUtils.getFilenameWithoutExtension(directoryF));
 
+            int intentFlag = PendingIntent.FLAG_UPDATE_CURRENT;
+            if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M) {
+                intentFlag |= PendingIntent.FLAG_IMMUTABLE;
+            }
```
harshad1, line 54: @gsantner 

Just merged in android X

I had to make these changes to the pendingIntent in order to get the widget to work with androidX. They should be ported to master before this PR is merged in. Without this change we get a crash if a widget is present.

app/src/main/java/net/gsantner/opoc/util/ContextUtils.java
```diff
@@ -447,10 +448,14 @@ public boolean isAppInstalled(String packageName) {
...
-        PendingIntent pendi = PendingIntent.getActivity(_context, 555, intent, PendingIntent.FLAG_CANCEL_CURRENT);
-        AlarmManager mgr = (AlarmManager) _context.getSystemService(Context.ALARM_SERVICE);
+    public void restartApp(final Class classToStart) {
+        final Intent intent = new Intent(_context, classToStart);
+        int flag = PendingIntent.FLAG_CANCEL_CURRENT;
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M) {
+            flag |= PendingIntent.FLAG_IMMUTABLE;
+        }
+        final PendingIntent pendi = PendingIntent.getActivity(_context, 555, intent, flag);
+        final AlarmManager mgr = (AlarmManager) _context.getSystemService(Context.ALARM_SERVICE);
```
harshad1, line 458: Same pendingIntent change here as well

harshad1:
@gsantner Aside from the issues with PendingIntent noted above we need to add a dialog to request all file permissions.

If I upgrade / install over an earlier release I don't get a request for all file access and I just see empty folders with all my files missing. Had to go into app permissions and manually give it all file access

gsantner:
webview fileacess issue (images/font) is what I heard today.
 

regarding permission, manage file permissions should be asked as soon you click the red new file button, or open a file...does it not?

app/thirdparty/java/other/writeily/widget/WrMarkorWidgetProvider.java
gsantner: Can merge/commit/pre-merge anything as needed

harshad1:
> regarding permission, manage file permissions should be asked as soon you click the red new file button, or open a file...does it not?

I get the permission request when I click the new button. But it should check this on startup every time, I think. Otherwise I can see the folder structure, but no files.

gsantner:
wasnt there also some merge request which also removed many `if (checkpermission) { fileaction }` condition guards?

gsantner:
added more permission prompts, and webview file access - pushed to master

app/thirdparty/java/other/writeily/widget/WrMarkorWidgetProvider.java
gsantner: merged

gsantner:
I'm seeking towards a short hotfix release in a hour or so, though not tagged - meaning F-Droid users will keep non-androidx for this one.

harshad1:
Iirc I factored those calls out and reduced the number but didn't remove all of them. We may have to revisit this.

We may also have to revisit temp files etc to mitigate android's shitty FS.

Right now though I am using partial highlighting + androidx and everything is working fine

Aug. 11, 2022 11:41:23 Gregor Santner ***@***.***>:

gsantner:
Good to hear! Got some mail about crashes on visiting certain Markor tabs, or overall random crashes -- but dont have a log and cant reproduce it

gsantner:
I did some great advancements/achievements  with newer android file handling, like opening files from other apps without direct file access.

What I added is kind of a proxy file which automatically writes back/forth to/from the contentresolver Uri stuff....while still everything is java.io.File based.  
That of course is for one file, but thats the main use, to open some specific file from some app. So nothing for file browser, but still great addition for Interoptability with other apps.

See also my comment in the discussion thread. If you want, feel free to give it a try as well.

gsantner:
> Foldersync file manager gives the option to open the file in Markor but I get the 'Markor can only work with local files' message and it doesn't work.

Hmm I installed it new From google play and it worked...interesting..Anyway also found out that foldersync also gives enough information to use direct file access (as in the real filepath). So added a singleline change which now results in foldersync resorting to direct file access if possible 😄 .

gsantner:
@harshad1 

the intent has most flags including the extras and action fields set to null....rather guess the widget issue has todo with SDK upgrade or the pendingintent change?

grafik

gsantner:
when I comment that new line of change...it works again

grafik

gsantner:
@harshad1 

the intent from widget file click has most flags including the extras and action fields set to null....rather guess the widget issue has todo with SDK upgrade or the pendingintent change?

grafik

-----

when I comment that new line of change...it works again

grafik

... or by exchange of _IMMUTABLE with _MUTABLE.

What I found: `Thanks for this, solves my FLAG_MUTABLE requirement on an Intent that loads a URL dynamically from a ListView in a widget.` <https://stackoverflow.com/questions/67045607/how-to-resolve-missing-pendingintent-mutability-flag-lint-warning-in-android-a>

gsantner:
cherry picked that in...also had it quite similary locally already but you were faster pushing xD

gsantner:
Installed the current state and from using it looks good. Only exception, links are not  highlighted (blue/smaller). Which means in a link at todo.txt .. in Mordor 3.5 lines of text, and in Markor 1.5 lines of text.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,6 @@
 #Thu Dec 02 03:48:33 CET 2021
 distributionBase=GRADLE_USER_HOME
 distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-bin.zip
-distributionSha256Sum=f6b8596b10cce501591e92f229816aa4046424f3b24d771751b06779d58c8ec4
```
gsantner, line 4: android studio tells you it might make issues to verify the hashsum, but there is one option (the second one) to keep the sha, and it works just fine in android studio then (at least here)

harshad1: Android studio would not build unless I removed this.

harshad1:
Link issue should be fixed - I'd forgotten to add it to markdown (was active only in PlainText)

gsantner (approved):

gsantner:
How does this file behavr for you? With highlighting disable scrolling is smooth & fluidic. But with highlighting enabled it's quite choppy and jumping by pixels

[2019-09-14-reddit-classicwow-the-loch.md](https://github.com/gsantner/markor/files/9349091/2019-09-14-reddit-classicwow-the-loch.md)


[screenrecord-2022-08-16_10.32.47.webm](https://user-images.githubusercontent.com/6735650/184835488-01085f64-e905-45e3-95f4-055802b65bbc.webm)

harshad1:
Some parts are very choppy. Some are smooth. I'll investigate.

Thanks for the test case.

Aug. 16, 2022 01:37:12 Gregor Santner ***@***.***>:

gsantner:
its because that one really long line is parsed again and again in partial highlighting (=much time). at full highlighting it's just parsed once

harshad1:
Lol. The cause is very long lines. We keep calling line start and end which slows down on long lines. I think I know how to solve it... Will take a few days.

Aug. 16, 2022 01:37:12 Gregor Santner ***@***.***>:

gsantner:
Heres another sample, it also have one long line. Though scrolling is not slowed at this one, somehow. 

[2019-03-27-reddit-zelda-i-made-an-octorock-out-of-socks-socktorock.md](https://github.com/gsantner/markor/files/9350826/2019-03-27-reddit-zelda-i-made-an-octorock-out-of-socks-socktorock.md)

harshad1:
@gsantner Ok the issue was actually trying to reflow a very long line. We reflow to mitigate tearing in todo.txt etc. I have reworked how this is done and hopefully both issues are solved now. Please test when you get a chance. 

Zim is completely untested btw. If we can get a zim user to test this branch, that would be amazing.

gsantner:
Zim looks very well, maybe even better than Markdown hehe. But I don't use it either.

screenshot-2022-08-17__09-39-35
screenshot-2022-08-17__09-39-39

gsantner (approved):

PR #1770, opened 2022-07-21
make images fill 100% width of the screen, closes #1769

n-wissam:
Image width is small and not responsive to to fill 100% of available screen width and you can notice that more in landscape mode and on larger screens.
The max-width property defines the maximum width of an image and it works if the image width is larger than the maximum available width . If the image width is smaller than the maximum available width, the max-width property has no effect.

See also <https://github.com/gsantner/markor/discussions/1769>

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/TextConverter.java
+++ b/app/src/main/java/net/gsantner/markor/format/TextConverter.java
@@ -45,7 +45,7 @@ public abstract class TextConverter {
     protected static final String TOKEN_POST_LANG = "{{ post.lang }}";
 
     protected static final String HTML_DOCTYPE = "<!DOCTYPE html>";
-    protected static final String HTML001_HEAD_WITH_BASESTYLE = "<html lang='" + TOKEN_POST_LANG + "'><head>" + CSS_S + "html,body{padding:4px 8px 4px 8px;font-family:'" + TOKEN_FONT + "';}h1,h2,h3,h4,h5,h6{font-family:'sans-serif-condensed';}a{color: " + TOKEN_LINK_COLOR + ";text-decoration:underline;}img{height:auto;width:325px;margin:auto;}" + CSS_E;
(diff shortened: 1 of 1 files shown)


gsantner:
this change "width:100%" also means that i.e. a small 32px icon gets full width icon. 

Other ideas?

n-wissam:
> this change "width:100%" also means that i.e. a small 32px icon gets full width icon.

IMHO that's totally fine since images are used mostly in markdown, not icons. and using `width:325px;` makes small 32px icon gets the same width as other images.
I have tried multiple markdown editors on desktop and I don't think  one of them supports resizing images depending on if the image is an icon or a relatively larger image

gsantner:
I'm not sure if generally resize to upscale is good. Setting max-width/height is so it not overshoots.

 But generally if you have a camera picture, its already quite big usually, isnt it? (so it rather is downscaled to fit, then upscaled).

Just about behaviour on my device, with 350px as it is now. Fills 3/4 of width roughly in portrait. So about 420px would be filling already. Screen is fullhd. And when I rotate the screen, still I can see the image in full glance without it being chopped of.  That is the intention behind this.

What I rather would try is a combination of max-width, no fixed width, and a max-vh. Assuming a square 500px img, it would already get width downscaled at portrait. Real small images stay as they are. For landscape, vh can be used to make image fit on screen page, but not sure how likable it is by users - depends on content.

n-wissam:
> And when I rotate the screen, still I can see the image in full glance without it being chopped of. That is the intention behind this.

You are right I can see the image but it's small compared to available screen width like you see in the screenshot below:

IMG_20220722_122634_964



> What I rather would try is a combination of max-width, no fixed width, and a max-vh. Assuming a square 500px img, it would already get width downscaled at portrait. Real small images stay as they are. For landscape, vh can be used to make image fit on screen page, but not sure how likable it is by users - depends on content.

I am sorry I didn't get what you meant here completely but if the purpose is just to scale down larger images and leave smaller images as they are why not stop using `width:325px;`  ?

gsantner:
> I am sorry I didn't get what you meant here completely but if the purpose is just to scale down larger images and leave smaller images as they are why not stop using width:325px; ?

This is what I talked about. Removing 325 and having other means to limit the size to be viewable.

gsantner:
I made changes on my own here and it's on master.

Closing this.

gsantner (commented):
I made changes on my own here and it's on master.

Closing this.

PR #1781, opened 2022-08-02
Prevent crash on rotate in main activity

harshad1:
_document can be null

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
@@ -202,6 +202,7 @@ public class DocumentEditFragment extends GsFragmentBase implements TextFormat.T
         final Bundle args = getArguments();
         setViewModeVisibility(args.getBoolean(START_PREVIEW, _appSettings.getDocumentPreviewState(_document.getPath())));
 
+        _loadModTime = 0; // force next load
         loadDocument();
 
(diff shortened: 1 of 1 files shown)


gsantner (approved):

gsantner:
Guess its new Markor update time soon? (not sure with/without partial HL)

harshad1:
> Guess its new Markor update time soon?

I would like to test for a few more days before the next point release, if that's ok. I want to see if there are further ways to mitigate android's poor filesystem behavior. Perhaps once the Androidx changes are ready?

Partial HL will take a long time yet IMO. There are many edge cases which need to be solved.

gsantner:
The AndroidX PR is to be prepared for upcoming times. I don't have a plan to merge it with the next update(s). Just when really needed and it brings notable improvements somewhere.  
So unlike HL performance, it's not something I'm working on ""to get done"".

gsantner:
The AndroidX PR is to be prepared for upcoming times. I don't have a plan to merge it with the next update(s). Just when really needed and it brings notable improvements somewhere.  
So unlike HL performance, it's not something I'm working on ""to get done"".

Generally it's a major change of the app, and it needs some user testing. It may work fine for me, but don't wanna let this one out and see it breaks everywhere at users.

gsantner:
As I know this one is a really bad issue, in the sense text can get lost...want to release a markor update today.

Is there something from you I can merge for this? If you say, hay nothing is totally broken, it's improving something, but there is still something that can be made better...it's ok when it's not perfect yet.

PR #1789, opened 2022-08-14
Mention zim wiki support in F-Droid description

Self-Perfection:


Self-Perfection:
I hope zim wiki support get to F-Droid, Google Play and https://alternativeto.net/software/markor/about/

AFAIK Markor is the only Android app that supports zim wiki syntax. For now Zim wiki can discover Markor mostly by the grapevine, it is not searchable in F-Droid / Google Play by "zim" keyword.

gsantner:
Would you mind to generally update the readme & f-droid drescription? (Which similary goes to website & others). 

It's not just Zim is missing there 😄

gsantner:
Would you mind to generally update/overhaul the readme & f-droid description? (Which similary goes to website & others). 
It's not just Zim is missing there 😄 .

I'm specifically also interested in what users think and value most at Markor, so there might be things more important to you than to me at Markor.

Self-Perfection:
@gsantner I find current description concise and to the point. I mostly use following Markor features:

1. Zim wiki syntax support
2. Easy syncing via syncthing
3. Sharing text to Markor (using it as a clipboard)

Point 1 was missing, point 2 is already described in substantial details. So after re-reading description I spotted only one possible improvement: clipboard feature description was vague. I tried to improve it.

So that's it. I am surely missing something but I do not pay close attention to changelogs.

gsantner (approved):
Did a few additions as well. But still it's only F-Droid not readme that is improved.

Anyway, I want to add that change now.

PR #1801, opened 2022-08-18
Editor performance: disable emoji compat, closes #1794

harshad1:
This seems to solve the perf issues. Also cleaned up a few things

See #1794


gsantner:
Does it require some if (Api > 32)?

If you have something else around, fine to add here.  
Want to make a update right afterwards (the third 2.10.2 by now).

Thanks for checking & improving!

harshad1:
I don't think so. But I have only tested it on my device. Will need to try it on an older device later today.

There are a few more tweaks related to highlighting which I need to add. I will do so here, but it will take a few days.

harshad1:
I think this is good to merge as is. I don't have any changes to make rn

gsantner (approved):

clach04:
Belated thank you to @harshad1 🎉 This had been driving me crazy on an older device that I occasionally use, I'd been assuming it was due to the age of the device 😆

PR #1806, opened 2022-08-20
[Feature] View mode: open Image,Video,Audio files in Markor

### LINKED ISSUE #1200: Open media in the Markor app (WebView compatible formats)
#### General information

* **App version:** 2.5
* **System:** 11

#### Description
Allow to open images (jpg/png/webp) in the app - using capabilties of the View-Mode. The webview supports these files AND comes with builtin zoom controls, pinch-to-zoom etc.

So: No addition of custom UI, ImageDisplay classes etc. Instead just use the view-mode webview with img src.

-------

[List](https://developer.android.com/guide/topics/media/media-formats) of supported formats from Android and so most of them on WebView too:

f

gsantner:
View mode: Feature open Image,Video,Audio files and other WebView supported formats in view mode (allow open jpg/mp4/mp3 etc directly in Markor), by @gsantner, closes #1200

* List of supported filetypes depends on native support by Android OS & WebView, [here](https://developer.android.com/guide/topics/media/media-formats) is a overview of the supported filetypes. This of course includes the most relevant stuff like jpg/png/mp4/mp3/webm.. 
* Also added: WebChromeClient => fullscreen video playback. This means also videos embedded in markdown then also have the fullscreen button enabled.
* Extract generic media information like title/artist/bitrate if possible with builtin Android utility

state 28. august:  
screenshot-2022-08-28__16-59-59-embed-binary

state of earliest version:  video, image, audio .. and also you can zoom in with zoom gesture 😄 . And yes everything shown is in a webview using html5 video/img tags + file://reference.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
@@ -42,7 +42,6 @@ import net.gsantner.markor.BuildConfig;
 import net.gsantner.markor.R;
 import net.gsantner.markor.format.TextConverter;
 import net.gsantner.markor.format.TextFormat;
-import net.gsantner.markor.format.binary.EmbedBinaryConverter;
 import net.gsantner.markor.format.general.DatetimeFormatDialog;
 import net.gsantner.markor.model.Document;
 import net.gsantner.markor.ui.AttachImageOrLinkDialog;
@@ -63,7 +62,6 @@ import net.gsantner.opoc.preference.FontPreferenceCompat;
 import net.gsantner.opoc.ui.FilesystemViewerData;
 import net.gsantner.opoc.util.ActivityUtils;
 import net.gsantner.opoc.util.CoolExperimentalStuff;
-import net.gsantner.opoc.util.FileUtils;
 import net.gsantner.opoc.util.StringUtils;
 import net.gsantner.opoc.util.TextViewUndoRedo;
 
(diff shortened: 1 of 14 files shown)


app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -134,12 +138,16 @@ public void onCreate(final Bundle savedInstanceState) {
         }
     }
 
+    public boolean isViewBinary(){
+        return !(getActivity() instanceof MainActivity);
```
gsantner, line 142: Yes this will move around and have a proper detection, it's just for development currently

app/src/main/java/net/gsantner/markor/format/TextFormat.java
```diff
@@ -61,6 +66,7 @@ public class TextFormat {
             CONVERTER_ZIMWIKI,
             CONVERTER_KEYVALUE,
             CONVERTER_PLAINTEXT,
+            CONVERTER_EMBEDBINARY,
     };
 
     public static boolean isTextFile(final String absolutePath) {
```
gsantner, line 72: Strictily speaking, no of course it's not a textfile. But there won't be any other such case, either it's a binary file and is embedded in webview via file:// reference, or it is plaintext.

gsantner, line 72: Strictily speaking, no of course it's not a textformat/textfile. But there won't be any other such case, either it's a binary file and is embedded in webview via file:// reference, or it is plaintext.

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
gsantner, line 142: Yes this will move around and have a proper detection, it's just for development currently. And it will only be allowed to show the view mode for such files. Editor text stays empty, as well read/write/saveText disabled.

gsantner:
@harshad1   
Some peak on this really nice feature, Really nice previewing of filetypes supported by WebView. Note that it not uses any kind of external dependencies or custom UI - it's just Markor's view mode, using the same way templating/converter is used at todo.txt. (just that editor/file-saving is disabled as it's binary)

[screenrecord-2022-08-21_15.07.35.webm](https://user-images.githubusercontent.com/6735650/185792433-91785c24-abb0-40b8-b582-f799d6402ba6.webm)

I discovered this handy class <https://developer.android.com/reference/android/media/MediaMetadataRetriever> that can give many different kinds of metadata, really well working. It's also capable of extracting music cover art, or video preview/frame with ~2 lines of code.  

What I want to do in future is also to add the outputs of the extraction utility I wrote.... to the existing file-info-dialog. So meaning users get more insights straight from the Markor filebrowser's toolbar info button 😄 

What do you think?

gsantner (commented):
Looked through all files at merge overview looks good

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
gsantner: Similar to encryption hook (that reads/writes files differently than normal text) there is another one now that just returns empty string, or returns true when wanting to write. (=dummy). No relevant other change at fragment or someting thus.

PR #1807, opened 2022-08-21
Defensive data loss prevention

harshad1:
In this PR I add several things to prevent potential data loss

1. Mod time checks are augmented with an internal time database to mitigate issues with android's buggy FS
2. `_document != null` checks are replaced with a check and exit to prevent silent write failures
3. ~~onPause and onResume are triggered in a couple more places~~

I will personally want to test this for a long time before I am comfortable merging it into master. 

Todo:

- [ ] Better messaging etc
- [ ] Logging when we encounter a bad state
- [ ] Purging the time database when it grows too big

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
@@ -123,35 +123,12 @@ public class DocumentEditFragment extends GsFragmentBase implements TextFormat.T
     @Override
     public void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        checkCreateDocument(); // In the normal course of things document should be created here
-    }
-
-    // Using `if (_document != null)` everywhere is dangerous
-    // It may cause reads or writes to _silently fail_
-    // Instead we try to create it, and exit if that isn't possible
-    private boolean checkCreateDocument() {
-        if (_documentErrorExitInProgress) {
-            return false;
+        final Bundle args = getArguments();
+        if (_savedInstanceState != null && _savedInstanceState.containsKey(SAVESTATE_DOCUMENT)) {
+            _document = (Document) _savedInstanceState.getSerializable(SAVESTATE_DOCUMENT);
+        } else if (args != null && args.containsKey(Document.EXTRA_DOCUMENT)) {
+            _document = (Document) args.get(Document.EXTRA_DOCUMENT);
         }
-
-        if (_document == null) {
-            final Bundle args = getArguments();
-            if (_savedInstanceState != null && _savedInstanceState.containsKey(SAVESTATE_DOCUMENT)) {
-                _document = (Document) _savedInstanceState.getSerializable(SAVESTATE_DOCUMENT);
-            } else if (args != null && args.containsKey(Document.EXTRA_DOCUMENT)) {
-                _document = (Document) args.get(Document.EXTRA_DOCUMENT);
-            }
-        }
-
-        // At this stage, if the document is still null, a hard exit is the best
-        if (_document == null) {
-            final Activity activity = getActivity();
-            Toast.makeText(activity, R.string.document_error_exit_message, Toast.LENGTH_LONG).show();
-            _documentErrorExitInProgress = true;
-            activity.finish();
-        }
-
-        return _document != null;
     }
 
     @Override
@@ -165,17 +142,25 @@ public class DocumentEditFragment extends GsFragmentBase implements TextFormat.T
         super.onViewCreated(view, savedInstanceState);
         final Activity activity = getActivity();
         _appSettings = new AppSettings(activity);
-        if (_appSettings.getSetWebViewFulldrawing() && Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-            WebView.enableSlowWholeDocumentDraw();
-        }
 
         _hlEditor = view.findViewById(R.id.document__fragment__edit__highlighting_editor);
         _textActionsBar = view.findViewById(R.id.document__fragment__edit__text_actions_bar);
         _textSdWarning = view.findViewById(R.id.document__fragment__edit__content_editor__permission_warning);
         _webView = view.findViewById(R.id.document__fragment_view_webview);
         _primaryScrollView = view.findViewById(R.id.document__fragment__edit__content_editor__scrolling_parent);
-
         _shareUtil = new ShareUtil(activity);
+
+        // Using `if (_document != null)` everywhere is dangerous
+        // It may cause reads or writes to _silently fail_
+        // Instead we try to create it, and exit if that isn't possible
+        if (_document == null || _hlEditor == null || _appSettings == null) {
+            Toast.makeText(activity, R.string.document_error_exit_message, Toast.LENGTH_LONG).show();
+            activity.finish();
+        }
+
+        if (_appSettings.getSetWebViewFulldrawing() && Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            WebView.enableSlowWholeDocumentDraw();
+        }
 
         _webViewClient = new MarkorWebViewClient(activity);
         _webView.setWebViewClient(_webViewClient);
@@ -265,14 +250,8 @@ public class DocumentEditFragment extends GsFragmentBase implements TextFormat.T
     }
 
     public void resume() {
-        if (!getLifecycle().getCurrentState().isAtLeast(Lifecycle.State.STARTED)) {
-            return;
-        }
-
         loadDocument();
-        if (_hlEditor != null) {
-            _hlEditor.onResume();
-        }
+        _hlEditor.onResume();
     }
 
     @Override
@@ -288,22 +267,11 @@ public class DocumentEditFragment extends GsFragmentBase implements TextFormat.T
     }
 
     public void pause() {
-        if (!getLifecycle().getCurrentState().isAtLeast(Lifecycle.State.RESUMED)) {
-            return;
-        }
-
         saveDocument(false);
-        if (_hlEditor != null) {
-            _hlEditor.onPause();
-        }
-
-        if (checkCreateDocument() && _appSettings != null) {
-            _appSettings.addRecentDocument(_document.getFile());
-            _appSettings.setDocumentPreviewState(_document.getPath(), _isPreviewVisible);
-            if (_hlEditor != null) {
-                _appSettings.setLastEditPosition(_document.getPath(), _hlEditor.getSelectionStart());
-            }
-        }
+        _hlEditor.onPause();
+        _appSettings.addRecentDocument(_document.getFile());
+        _appSettings.setDocumentPreviewState(_document.getPath(), _isPreviewVisible);
+        _appSettings.setLastEditPosition(_document.getPath(), _hlEditor.getSelectionStart());
     }
 
     @Override
@@ -391,7 +359,7 @@ public class DocumentEditFragment extends GsFragmentBase implements TextFormat.T
 
     public boolean loadDocument() {
         //Only trigger the load process if constructing or file updated
-        if (checkCreateDocument() && _document.hasFileChangedSinceLastLoad()) {
+        if (_document.hasFileChangedSinceLastLoad()) {
 
             final String content = _document.loadContent(getContext());
             if (!_document.isContentSame(_hlEditor.getText())) {
@@ -513,11 +481,9 @@ public class DocumentEditFragment extends GsFragmentBase implements TextFormat.T
             case R.string.action_format_todotxt:
             case R.string.action_format_plaintext:
             case R.string.action_format_markdown: {
-                if (checkCreateDocument()) {
-                    _document.setFormat(itemId);
-                    applyTextFormat(itemId);
-                    _appSettings.setDocumentFormat(_document.getPath(), _document.getFormat());
-                }
+                _document.setFormat(itemId);
+                applyTextFormat(itemId);
+                _appSettings.setDocumentFormat(_document.getPath(), _document.getFormat());
                 return true;
             }
             case R.id.action_search: {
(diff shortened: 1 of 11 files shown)


app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -236,7 +254,7 @@ public void onFragmentFirstTimeVisible() {
                 final int lno = args.getInt(Document.EXTRA_FILE_LINE_NUMBER);
                 if (lno >= 0) {
                     startPos = StringUtils.getIndexFromLineOffset(_hlEditor.getText(), lno, 0);
-                } else {
+                } else if (lno == Document.EXTRA_FILE_LINE_NUMBER_LAST){
```
harshad1, line 257: Using a sentinel with a random value (not -1) to explicitly trigger the open file at end case.

Fixes the bug with shortcuts opening files at the end. Will break this out into a separate PR

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -760,20 +771,17 @@ public void onAnimationEnd(Animator animation) {
 
     @Override
     protected boolean onToolbarLongClicked(View v) {
-        if (getUserVisibleHint()) {
```
harshad1, line 763: getUserVisibleHint is depreciated.

app/src/main/java/net/gsantner/markor/model/Document.java
```diff
@@ -100,7 +104,32 @@ public File getFile() {
...
+        initModTimePref();
+        _touchTime = System.currentTimeMillis();
+        _modTimePref.edit().putLong(_file.getAbsolutePath(), _touchTime).apply();
+    }
+
+    public boolean hasFileChangedSinceLastLoad() {
+        return _modTime < 0                            // Never read
+                || _touchTime < 0                      // Never read
+                || fileModTime() > _modTime            // File mod time updated
+                || getGlobalTouchTime() > _touchTime;  // File has been modified by another document
```
harshad1, line 129: If we hit this line and the result is _true_ then the mod time is in a bad state. We should log it

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -124,8 +125,8 @@ public DocumentEditFragment() {
     public void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         final Bundle args = getArguments();
-        if (_savedInstanceState != null && _savedInstanceState.containsKey(SAVESTATE_DOCUMENT)) {
-            _document = (Document) _savedInstanceState.getSerializable(SAVESTATE_DOCUMENT);
+        if (savedInstanceState != null && savedInstanceState.containsKey(SAVESTATE_DOCUMENT)) {
+            _document = (Document) savedInstanceState.getSerializable(SAVESTATE_DOCUMENT);
```
gsantner, line 129: While developing #1809 I noticed that there is an issue with rotation at the viewmode vs editmode state.  
So currently on master when you **enter** as editor, go to viewmode, rotate the device, you are in the editor again.  
Also the opposite when you  **enter** in viewmode, rotating the device will always go back to viewmode.

Guess also needs some savedInstance code?

app/src/main/java/net/gsantner/markor/model/Document.java
gsantner: Fine to add a few `Log.`'s at the central important places

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
gsantner: maybe Fragment::isResumed is relevant?  
 <https://stackoverflow.com/questions/57885849/in-androidx-fragment-app-fragment-setuservisiblehint-is-deprecated-and-not-exec/68987518#68987518>

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -210,8 +209,14 @@ public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
         }
 
         _webView.setBackgroundColor(Color.TRANSPARENT);
+
         final Bundle args = getArguments();
-        setViewModeVisibility(args.getBoolean(START_PREVIEW, _appSettings.getDocumentPreviewState(_document.getPath())));
+        final boolean startInPreview = _appSettings.getDocumentPreviewState(_document.getPath());
+        if (args != null && savedInstanceState == null) { // Use the launch flag on first launch
+            setViewModeVisibility(args.getBoolean(START_PREVIEW, startInPreview));
```
gsantner, line 226: ok so looks like you added something here already on the quick, nice! thanks! I wanted to do the same now, but so don't 😄

gsantner:
I received this:

```
Absolutely fantastic text editor! I've previously tried writing diary in other apps, but nothing has handled the 900 000 characters even close to as well. The load times are basically zero, it has all options one might need and more, the file dialog is so good I use that instead of the built-in file manager, it offers dark theme, setting backup, markdown shotcuts and on top of all it's fully FOSS, meaning it's as good for privacy as possible. I can't think of any possible improvements. Easy 5★! 
```

. Not sure what device that is, but sounds like a very big file of a dimension I didn't think TextView is capable of. Nice o hear of course 😄

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -214,16 +215,31 @@ public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
...
         _webView.setBackgroundColor(Color.TRANSPARENT);
-        final Bundle args = getArguments();
-        setViewModeVisibility(args.getBoolean(START_PREVIEW, _appSettings.getDocumentPreviewState(_document.getPath())));
+    }
 
-        _loadModTime = 0; // force next load
-        loadDocument();
+    // Runs after all state has been restored (i.e. text is restored etc)
+    @Override
+    public void onViewStateRestored(final Bundle savedInstanceState) {
```
harshad1, line 222: Moved several things to `onViewStateRestored` so they happen correctly later in the lifecycle

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -236,17 +252,15 @@ public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
...
         int startPos = _appSettings.getLastEditPosition(_document.getPath(), _hlEditor.length());
-        final Bundle args = getArguments();
 
         // First start - overwrite start position if needed
         if (_savedInstanceState == null) {
-            if (isDisplayedAtMainActivity() || _appSettings.isEditorStartOnBotttom()) {
-                startPos = _hlEditor.length();
-            } else if (args != null && args.containsKey(Document.EXTRA_FILE_LINE_NUMBER)) {
+            final Bundle args = getArguments();
+            if (args != null && args.containsKey(Document.EXTRA_FILE_LINE_NUMBER)) {
```
harshad1, line 259: Fix for files starting at end

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -255,50 +269,27 @@ public void onFragmentFirstTimeVisible() {
...
-            _webView.onResume();
-        }
-    }
-
     @Override
     public void onResume() {
-        resume();
+        loadDocument();
+        _webView.onResume();
         super.onResume();
```
harshad1, line 276: Removed lots of tests. We should not encounter a silent failure to write the file now. (There is a greater chance of npe, looking into how we can catch this and save the file contents)

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -378,12 +369,9 @@ private void updateUndoRedoIconStates() {
...
 
-    public void loadDocument() {
-        //Only trigger the load process if constructing or file updated
-        final long modTime = _document != null ? _document.lastModified() : Long.MIN_VALUE;
-        if (modTime > _loadModTime) {
-
-            _loadModTime = modTime;
+    public boolean loadDocument() {
+        // Only trigger the load process if constructing or file updated
+        if (_document.hasFileChangedSinceLastLoad()) {
```
harshad1, line 374: Moved modified logic into _document for greater correctness and simplicity

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -392,25 +380,34 @@ public void loadDocument() {
...
                 sel[1] = Math.min(sel[1], content.length());
 
+                if (_editTextUndoRedoHelper != null) {
+                    _editTextUndoRedoHelper.disconnect();
+                    _editTextUndoRedoHelper.clearHistory();
+                }
+
                 _hlEditor.withAutoFormatDisabled(() -> _hlEditor.setText(content));
 
+                if (_editTextUndoRedoHelper == null) {
```
harshad1, line 390: create the UndoRedoHelper _after_ first load. Also disable and reset it after a load.

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -470,14 +464,14 @@ public boolean onOptionsItemSelected(final MenuItem item) {
             case R.id.action_share_html_source: {
                 if (saveDocument(false)) {
                     TextConverter converter = TextFormat.getFormat(_document.getFormat(), activity, _document).getConverter();
-                    _shareUtil.shareText(converter.convertMarkup(_hlEditor.getText().toString(), _hlEditor.getContext(), false, _document.getFile()),
```
harshad1, line 473: `_hlEditor.getText().toString()` was generating a bunch of warnings everywhere. Moved it to a single function in order to minimize noise

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -625,36 +616,22 @@ public void applyTextFormat(final int textFormatId) {
         updateMenuToggleStates(textFormatId);
     }
 
-    private void initDocState() {
```
harshad1, line 628: This function is called in 1 place. Just rolled it inline

app/src/main/java/net/gsantner/markor/activity/DocumentEditFragment.java
```diff
@@ -784,7 +760,7 @@ public void onAnimationEnd(Animator animation) {
 
     @Override
     protected boolean onToolbarLongClicked(View v) {
-        if (getUserVisibleHint()) {
+        if (isVisible() && isResumed()) {
```
harshad1, line 763: I think this is working

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
```diff
@@ -20,17 +21,17 @@
...
-import android.view.ViewGroup;
 import android.view.WindowManager;
 
 import androidx.annotation.NonNull;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentManager;
-import androidx.fragment.app.FragmentPagerAdapter;
-import androidx.viewpager.widget.ViewPager;
+import androidx.viewpager2.adapter.FragmentStateAdapter;
+import androidx.viewpager2.widget.ViewPager2;
```
harshad1, line 30: Switched to ViewPager2. Seems to be working well with better child-lifecycle handling

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
```diff
@@ -117,7 +117,7 @@ public void onPageSelected(int position) {
     protected void onNewIntent(final Intent intent) {
         super.onNewIntent(intent);
         final File dir = getIntentDir(intent, null);
-        final FilesystemViewerFragment frag = (FilesystemViewerFragment) _viewPagerAdapter.getFragmentByTag(FilesystemViewerFragment.FRAGMENT_TAG);
+        final FilesystemViewerFragment frag = getNotebook();
```
harshad1, line 120: Made a 1 line function to make this common action cleaner

gsantner:
Hey @harshad1 ,

I know you still want to make some improvements, but generally it looks like a state which currently works fine (or at last better). So I merge this now and feel free to make a new pull request again. This also means, it resets the review a bit as I don't need to check the same changes again 😄 .

I merged my recent additions to master and now master into yours and resolved the conflicts.   
Want to do some refactoring and don't wanna put the burden on you to resolve all the merge conflicts later on, so I see it as the better option to merge this now.

gsantner:
@harshad1  
To me the new ViewPager feels a bit slow/sluggish. Is that also for you? I.e. when I'm at filemanager and click the More tab I get literally a slideshow taking ~2 seconds. At old ViewPager it was nearly instant.

harshad1:
Yeah. I see this too. Didn't really notice it until you pointed it out though.

I am planning on making a fix for the tooltip issues on action button (unless you're working on this already).

I'll look into speeding up the viewpager animation at the same time when I do.

Aug. 29, 2022 15:56:37 Gregor Santner ***@***.***>:

gsantner:
> I am planning on making a fix for the tooltip issues on action button 
(unless you're working on this already).

No did not. Be warned, placing that tooltip is tricky and Android not 
gives any control over that. It can ONLY be shown by giving Android 
noticing a double click. And there is no control over the opacity or 
something.
I also tried with Toast before that, that was a mess, always overlaying 
important stuff (in latest iteration of Toast you can't even control the 
screen position anymore either).

The slight delay at show is intentional added though by me. It's because 
otherwise when a dialog or something pops up or on longclick (i.e. 
DateFormatDialog), you would never get to see any hint what you are doing.
Because Android automatically dismisses the tooltip immediately without 
that, and user won't see anything what it's about.

harshad1:
Got it. If I can't find a way to fix the issues I'm seeing, I'll add an option to disable them completely.

The behaviour (at least on my device) is pretty annoying.

H

Aug. 29, 2022 16:26:38 Gregor Santner ***@***.***>:

gsantner:
I dont want a settings option or something for that. If thats so horrible to you, we can check for other means, or as last resort show nothing...as in the past.

harshad1:
That's nice to hear! My personal log / journal files regularly hit 100k chars and I don't want to upgrade my phone so I've been working on improving textview perf for a while now. I'm glad others are getting something out of this too.

Aug. 23, 2022 13:36:48 Gregor Santner ***@***.***>:

PR #1809, opened 2022-08-21
View-Mode: Action buttons similar to Edit-Mode

gsantner:
Enable the bottom action bar also for view mode and populate some actions. Automatically swaps list of actions which are either available to edit-only, view-only or both.

Example actions implemented:
* jump to top/bottom
* jump to TableOfContents
* open filie in other app
* rotate screen

All of these I remember to being asked at least once and users being confused. So now with that addition users can jump to the table of contents for example just with one click reachable by thumb.

new functionality for opoc utils: rotate screen (automatically jumps in round between default->landscape->portrait->sensor->default)

possible future additions, using these buttons:
* #699 

screenshot-2022-08-21__18-38-26


gsantner:
@harshad1 what addition of actions to view-mode also means: we **could** drop the long-click on title code for view mode. as there is now a dedicated button for that reachable by thumb

harshad1:
@gsantner Noticed a bug today:

If you change the format type in a document, all action buttons are removed (the new format's action buttons are not added back)

gsantner:
ah alright, can reproducec that as well.

PR #1824, opened 2022-08-28
Defensive data loss prevention v2

harshad1:


gsantner:
Oh, wanted to not get into that situation...well. When I'm done I can help you to migrate the changes

harshad1:
Not a problem. Just a few remaining issues I am trying to solve. I'll resolve conflicts when I am done. 

Just named this v2 because didn't think too hard about it

gsantner:
OK. I'm doing some renamings like Highlighter -> SyntaxHighlighter, rename ZimWiki more generically Wikitext (which it is to the biggest part), TextActions -> ActionButtons...and more

The bigger thing I want to do still is merge ContextUtil, ActivityUtil and ShareUtil to one combined thing that accepts activity/context as parameter instead of saved copy.

harshad1:
@gsantner Am done here I think. You can merge it in. If I find any lingering bugs I can make another PR.

harshad1:
@gsantner Couple things I noticed from the last set of changes you made:

1. Renaming the action keys means that action order and enable / disable is reset to default - we should consider adding code to port this on update
2. The action tooltips look nice, but the current behavior is pretty annoying - they stay open and block the text until I scroll or click a different action. They should close immediately (< 1s) and should perhaps be partly transparent imo.

app/src/main/AndroidManifest.xml
```diff
@@ -53,7 +53,7 @@
         android:usesCleartextTraffic="true"
         tools:targetApi="o">
 
-        <!-- <profileable android:shell="true"/> -->
+        <profileable android:shell="true"/>
```
gsantner, line 56: uncommented

PR #1827, opened 2022-08-29
Fixe borked Instance State

harshad1:
This was supposed to go in the previous PR but got missed


app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
```diff
@@ -291,12 +293,6 @@ public void onPause() {
     @Override
     public void onSaveInstanceState(@NonNull Bundle outState) {
         outState.putSerializable(SAVESTATE_DOCUMENT, _document);
-
-        // Don't save contents - will be reloaded from file
-        _hlEditor.setHighlightingEnabled(false);
-        _hlEditor.setAutoFormatEnabled(false);
-        _hlEditor.setText("");
```
harshad1, line 298: onSaveInstance state is called all the time without a corresponding restore. This causes the file contents to be lost basically every time the fragment is paused

gsantner:
@harshad1 

Thanks for the fixes again. I'm done now with refactoring (at least to the 95% part of what I want to do). So don't be afraid about touching files.

Hope you will also like it, and let me know if you run into issues (especially with sharing to/from/opening in other apps).

So many package structure & filename changes, simpler inheritance structures, better base names (i.e. "Highlighter"=>"SyntaxHighlighter")...  

The biggest code wise changes are at Context/Activity/ShareUtils. This is now just one big utility class without 3-4 way inheritance. Context/activity objects are not cached anymore. Instead every method has a first parameter if it needs it, and what it needs. The ActivityUtil-ContextUtil distinction came previously from the background ContextUtil was soley for stuff working completley without activity. Now there is just one big class, accessible anywhere (by `_cu` mostly) and offering many important Android functionalities I built up in the last ~8 years.

gsantner:
@harshad1 

Thanks for the fixes again. I'm done now with refactoring (at least to the 95% part of what I want to do). So don't be afraid about touching files.

Hope you will also like it, and let me know if you run into issues (especially with sharing to/from/opening in other apps).

So many package structure & filename changes, simpler inheritance structures, better base names (i.e. "Highlighter"=>"SyntaxHighlighter")...  

The biggest code wise changes are at Context/Activity/ShareUtils. This is now just one big utility class without 3-4 way inheritance. Context/activity objects are not cached anymore. Instead every method has a first parameter if it needs it, and what it needs. The ActivityUtil-ContextUtil distinction came previously from the background ContextUtil was soley for stuff working completley without activity. Now there is just one big class, accessible anywhere (by `_cu` mostly) and offering many important Android functionalities I built up in the last ~8 years.

And yes, it will reset some settings - that is fine. I want to make the next Markor version v3.0.0 signaling also bigger changes. I had that version jump already in mind in previous 2.10.x versions but wanted to mature AndroidX a bit more.

PR #1834, opened 2022-09-01
Various bug fixes, early sept 2022

harshad1:
In this PR I fix several issues

- [x] Zim autoindent
- [x] App start tab
- [x] Text actions not being set on format change
- [x] Cleanups to patterns etc - removed redundant files and moved the hl patterns into the hl classes
- [x] Removed onViewStateRestored. Back to one function. Hopefully fewer crashes


harshad1:
@gsantner There are several issues in the latest master:

1. Font is not setting correctly. It always uses the default font not matter what we select.
2. Experimental settings are not being shown no matter what the toggle is set to

gsantner:
do we really need onViewStateRestored in documentfragment? The point is, the restored savedInstance stuff is also contained in onViewCreate. It's the same data isn't it? And onCreate&onVirwCreated are also called on rotation.

gsantner:
>  Not sure what the source of these errors is. I will be reverting to an earlier version until they are fixed.

I saw some preferences-ktx hint in build.gradle to upgrade. Not sure if it fixes something as well.

harshad1:
I have a feeling it's related to the context utils changes in AppSettings

Sep. 1, 2022 01:33:16 Gregor Santner ***@***.***>:

gsantner:
> I have a feeling it's related to the context utils changes in AppSettings

Hmm that shouldn't be the case. But I check it as well when I have time.

harshad1:
As an experiment, I reverted the AppSettings init system to just using the constructors as before and the settings issues appear fixed. I think the issue is related to the fact that `SHARED_PREFF_APP` is not used after your changes.

These reverts are in this PR

app/src/flavorAtest/java/net/gsantner/opoc/util/GsAndroidSupportMeWrapper.java
```diff
@@ -56,11 +56,11 @@ public void mainOnResume(final Context context) {
         }
```
harshad1, line 56: Studio thinks this entire class / file is unused. Not sure why, but if so, do we want to keep it?

app/src/main/java/net/gsantner/markor/activity/ActionButtonSettingsActivity.java
```diff
@@ -159,7 +160,7 @@ private Holder(View row) {
         }
 
         private void bindModel(final ActionButtonBase.ActionItem action, final String key, final Set<String> disabled) {
-            final Switch enabled = _row.findViewById(R.id.enabled_switch);
+            final SwitchCompat enabled = _row.findViewById(R.id.enabled_switch);
```
harshad1, line 163: Studio suggested this change

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
```diff
@@ -623,13 +625,14 @@ public void applyTextFormat(final int textFormatId) {
...
-        _textActionsBar.removeAllViews();
         _textFormat = FormatRegistry.getFormat(textFormatId, activity, _document);
         _hlEditor.setHighlighter(_textFormat.getHighlighter());
         _hlEditor.setDynamicHighlightingEnabled(_appSettings.isDynamicHighlightingEnabled());
         _hlEditor.setAutoFormatters(_textFormat.getAutoFormatInputFilter(), _textFormat.getAutoFormatTextWatcher());
         _hlEditor.setAutoFormatEnabled(_appSettings.getDocumentAutoFormatEnabled(_document.getPath()));
-        _textFormat.getTextActions().setUiReferences(activity, _hlEditor, _webView);
+        _textFormat.getTextActions()
+                .setUiReferences(activity, _hlEditor, _webView)
+                .recreateTextActionBarButtons(_textActionsBar, _isPreviewVisible ? ActionButtonBase.ActionItem.DisplayMode.VIEW : ActionButtonBase.ActionItem.DisplayMode.EDIT);
```
harshad1, line 635: Use recreate which calls removeAllViews

app/src/flavorAtest/java/net/gsantner/opoc/util/GsAndroidSupportMeWrapper.java
gsantner: It is currently unused yes. But you can just leave it

gsantner:
Ok guess its just a problem then with what preference file is used by default. Shouldn't be too hard to fix :D.

gsantner:
Ok guess its just a problem then with what preference file is used by default. Anf if it matches with the one from preferencefragment. Shouldn't be too hard to fix :D.

harshad1:
Was there a specific reason we switched from constructors to `init`?

gsantner:
Yesish. I wanted to make constructor private and singleton. But have not done that for the reason when really need another instance we have the option still a other option around. 

Generally said, it should not make any difference. This AppSettings creation and the AppliacationObject is the first of all that gets executed in the application overall.

app/src/main/java/net/gsantner/opoc/model/GsSharedPreferencesPropertyBackend.java
```diff
@@ -71,17 +71,14 @@ public class GsSharedPreferencesPropertyBackend implements GsPropertyBackend<Str
     protected String _prefAppName;
     protected Context _context;
 
-    public GsSharedPreferencesPropertyBackend() {
+    public GsSharedPreferencesPropertyBackend(final Context context) {
```
gsantner, line 74: I understand that you prefer AppSettings to instantiate with Context directly. But this here can be left with non-Context requiring constructor, doesn't it?

gsantner, line 74: I understand that you prefer AppSettings to instantiate with Context directly. But this here can be left with non-Context requiring constructor, doesn't it? You can also call `super.init(context, name)` in AppSettings constructor.

This here so I can decide at any time to drop this way of using AppSettings and not supply context right away at instantiation. But still keep it as a single instance put somewhere. And...then initalize it there with the context later.

harshad1: Got it. I will rework this back, hopefully keeping the fix intact

gsantner: OK thanks, hope you are okay with that. But times can change fast as you know. Wouldn't be suprised if we sooner or later with Android update we cannot trust the context being available at this stage anymore. Such big breaking things unfortunately happen, just like I never thought 5 years ago that Android file system would get crippled in such a way that it is now.

gsantner: > hopefully keeping the fix intact

generally said, it shouldn't make any difference at all in this case, for execution it doesn't matter if it's codeline, return, call init func, codeline, codeline,codeline...... or codeline,codeline,codeline,codeline,return

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
```diff
@@ -517,10 +517,12 @@ public boolean onOptionsItemSelected(@NonNull final MenuItem item) {
             case R.string.action_format_todotxt:
             case R.string.action_format_plaintext:
             case R.string.action_format_markdown: {
-                _document.setFormat(itemId);
-                applyTextFormat(itemId);
-                _appSettings.setDocumentFormat(_document.getPath(), _document.getFormat());
-                return true;
+                if (itemId != _document.getFormat()) {
```
harshad1, line 520: Only change format is we are changing

app/src/main/java/net/gsantner/markor/format/markdown/MarkdownReplacePatternGenerator.java
```diff
@@ -30,7 +31,13 @@ public class MarkdownReplacePatternGenerator {
     public static final Pattern PREFIX_UNORDERED_LIST = Pattern.compile("^(\\s*)((-|\\*|\\+)\\s)");
     public static final Pattern PREFIX_LEADING_SPACE = Pattern.compile("^(\\s*)");
 
-    static final Pattern[] PREFIX_PATTERNS = {
+    public static final AutoTextFormatter.FormatPatterns formatPatterns = new AutoTextFormatter.FormatPatterns(
```
harshad1, line 34: Dramatically simplify the number of files / classes / indirection used for the patterns. We can simplify them further in the future, but this is a good stopping point.

harshad1:
@gsantner I am done here I think - will continue testing and make a follow up if I find yet more bugs.

harshad1:
There is 1 unit test which needs fixing. @gsantner Could you try to fix that please? If not possible, could you comment it out or something - I have will have very limited time to work on markor for the next few weeks and want to spend that on stability + bugfixes.

gsantner:
Thank you for any time you help, doesn't matter how much, any contribution apprciated.

Don't worry about that one unit test.

gsantner:
Thank you for any time you help, doesn't matter how much, any contribution apprciated.

Don't worry about that one unit test.

Unless this PR breaks roation (because removing onViewStateRestored)...and user can confirm no crash...I merged this soon. But feel free for new PR.  This here has important fixes.

harshad1:
I haven't seen an issue with rotation yet. We switched to returning null within hleditor which seems to have solved the issue.

I think it's good to merge.

Sep. 2, 2022 09:34:26 Gregor Santner ***@***.***>:

app/src/main/java/net/gsantner/markor/activity/ActionButtonSettingsActivity.java
```diff
@@ -159,7 +159,7 @@ private Holder(View row) {
...
```
gsantner, line 162: grafik

gsantner (changes requested):

gsantner (approved):

harshad1:
@gsantner do you plan to make another release soon?

gsantner:
yes with #1835 also in

Issue #1833, opened 2022-09-01
onViewStateRestored: 2.10.3+ nightly build crashes on startup


Screwtapello:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**. Not a [question](https://github.com/gsantner/markor/issues/95) or [feature request](https://github.com/gsantner/markor/issues/new/choose).
- [X] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [X] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [X] Is the bug still present in the latest code version (git master)? Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Mordor**. No worries Markor and Mordor appear as completely separate applications, you can install both side-by-side and Markor settings are not touched. In case the issue is resolved there, you don't need to create a bug report - the change will be part of the next Markor update.

### Description

After the first-launch tour and the prompt for storage access permissions, and upon every launch thereafter, the "Mordor" nightly build crashes.

### Steps to reproduce

1. Download the latest nightly APK from the link in this issue template
2. Extract the .zip, and install the APK to the phone with `adb install`
3. Launch the "Mordor" app
4. The first-launch tour appears; tap the "next" button repeatedly to get to the end
5. A dialog appears, saying "Need permission for storage to save documents"
6. Within a second or so, whether or not you try to tap  OK or Cancel on the dialog, the app closes
7. Try to launch the app again, it vanishes again, and after enough attempts you get the Android "Mordor keeps stopping/App info/Close app" dialog

### Information

Android version: 10
Device: moto e7 power
App Version: 2.10.3-2156

### Source

Something else (-> Additional info)

### Format / File type

Not specific

### Additional info / Log

```shell
$ adb logcat -b crash
09-01 13:32:04.319 23628 23628 E AndroidRuntime: FATAL EXCEPTION: main
09-01 13:32:04.319 23628 23628 E AndroidRuntime: Process: net.gsantner.markor_test, PID: 23628
[...]

PR #1835, opened 2022-09-04
Rework onFragmentFirstTimeVisible, use isVisible&isResumed

harshad1:
Using isVisible() and isResumed() now. Better calls with `post`

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/ActionButtonBase.java
+++ b/app/src/main/java/net/gsantner/markor/format/ActionButtonBase.java
@@ -596,7 +596,7 @@ public abstract class ActionButtonBase {
                 return true;
             }
             case R.string.abid_common_ordered_list_renumber: {
-                renumberOrderedList(TextViewUtils.getSelection(_hlEditor)[0]);
+                renumberOrderedList();
                 return true;
             }
             case R.string.abid_common_move_text_one_line_up:
@@ -765,7 +765,7 @@ public abstract class ActionButtonBase {
     }
 
     // Derived classes should override this to implement format-specific renumber logic
-    protected void renumberOrderedList(final int cursorPosition) {
+    protected void renumberOrderedList() {
         // No-op in base class
     }
 
@@ -775,7 +775,7 @@ public abstract class ActionButtonBase {
 
     public final void runRenumberOrderedListIfRequired(final boolean force) {
         if (force || _hlEditor.getAutoFormatEnabled()) {
-            _hlEditor.withAutoFormatDisabled(() -> renumberOrderedList(TextViewUtils.getSelection(_hlEditor)[0]));
+            _hlEditor.withAutoFormatDisabled(this::renumberOrderedList);
         }
     }
 
--- a/app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
+++ b/app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
@@ -41,7 +41,7 @@ public class HighlightingEditor extends AppCompatEditText {
 
     final static int HIGHLIGHT_SHIFT_LINES = 8;              // Lines to scroll before hl updated
     final static float HIGHLIGHT_REGION_SIZE = 0.75f;        // Minimum extra screens to highlight (should be > 0.5 to cover screen)
-    final static long BRING_CURSOR_INTO_VIEW_DELAY_MS = 250; // Block auto-scrolling for time after highlighing (hack)
+    final static long BRING_CURSOR_INTO_VIEW_DELAY_MS = 200; // Block auto-scrolling for time after highlighing (hack)
 
     public final static String PLACE_CURSOR_HERE_TOKEN = "%%PLACE_CURSOR_HERE%%";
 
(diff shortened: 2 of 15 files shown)


app/src/main/java/net/gsantner/opoc/frontend/base/GsFragmentBase.java
```diff
@@ -149,17 +144,20 @@ public String getAppLanguage() {
...
         }
+    }
 
-        attachToolbarClickListenersToFragment();
+    @Override
+    public void onResume() {
+        super.onResume();
+        final View view = getView();
+        if (view != null) {
+            view.postDelayed(this::checkRunFirstTimeVisible, 1);
```
harshad1, line 159: May be good to bump up the delay to something higher. Something like ~ 200ms will be barely perceptible to user, but everything will have had time to draw and layout.

app/src/main/java/net/gsantner/opoc/frontend/base/GsFragmentBase.java
```diff
@@ -70,12 +69,7 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
         view.postDelayed(() -> {
-            synchronized (_fragmentFirstTimeVisibleSync) {
-                if (getUserVisibleHint() && isVisible() && _fragmentFirstTimeVisible) {
-                    _fragmentFirstTimeVisible = false;
-                    onFragmentFirstTimeVisible();
-                }
-            }
+            checkRunFirstTimeVisible();
             attachToolbarClickListenersToFragment();
         }, 1);
```
harshad1, line 74: Should probably only call from onResume - that is when the fragment is fully ready and displayed

app/src/main/java/net/gsantner/opoc/frontend/base/GsFragmentBase.java
```diff
@@ -149,17 +128,33 @@ public String getAppLanguage() {
...
+            onFragmentFirstTimeVisible();
+            attachToolbarClickListenersToFragment();
         }
+    }
 
-        attachToolbarClickListenersToFragment();
+    protected void attachToolbarClickListenersToFragment() {
+        final Toolbar toolbar = getToolbar();
+        if (toolbar != null) {
+            toolbar.setOnLongClickListener(clickView -> isVisible() && isResumed() && onToolbarLongClicked(clickView));
```
gsantner, line 142: suggestion: Callback.b2<bool,view> (islong, view) {

if ( isvisible && isRes.....) {


  if islong { return ontoolbarlongclick() }

  else { ontoolbarclick }
}
return false

---


setOnClivkListener(v -> callback(false, v));

setOnLonglivkListener(v -> callback(true, v));

what do you think?

app/src/main/java/net/gsantner/opoc/frontend/base/GsFragmentBase.java
gsantner: stability over speed nitpickers :-)

harshad1: It's still blazing fast. I have not found an editor which opens and highlights a large file as fast as markor :)

gsantner: Can you remotly reproduce the permoance issue mentioned in #1833 with ~5 seconds? I do not.

Especially with the permission request change it got much quicker on first start and further starts as far I can tell. (unrelated to edior)

app/src/main/java/net/gsantner/opoc/frontend/base/GsFragmentBase.java
harshad1: IMO it's extra indirection for no real benefit.

app/src/main/java/net/gsantner/opoc/frontend/base/GsFragmentBase.java
harshad1: > Can you remotly reproduce the permoance issue mentioned in https://github.com/gsantner/markor/issues/1833 with ~5 seconds? I do not.

My reading was that this is happening with the debug build... Which I expect to be much slower/

app/src/main/java/net/gsantner/opoc/frontend/base/GsFragmentBase.java
gsantner: its easier to debug as you can put a break point on i.e. isVisible line , to add i.e. another condition to both clicks

app/src/main/java/net/gsantner/markor/format/ActionButtonBase.java
```diff
@@ -596,7 +596,7 @@ protected final boolean runCommonAction(final @StringRes int action) {
                 return true;
             }
             case R.string.abid_common_ordered_list_renumber: {
-                renumberOrderedList(TextViewUtils.getSelection(_hlEditor)[0]);
```
harshad1, line 599: So I realized that selection is a property of the editable, not the textview. This means we can simplify the call to renumber and preserve selection

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -116,24 +119,33 @@ private void updateDynamicHighlighting() {
...
 
+    private synchronized void postUpdateDynamicHighlighting() {
+        if (!_hlPostQueued) {
+            _hlPostQueued = true;
+            post(() -> {
+                updateDynamicHighlighting();
+                _hlPostQueued = false;
+            });
+        }
+    }
```
harshad1, line 130: Using the post prevents the selection from being hidden when the keyboard is shown

PR #1838, opened 2022-09-07
AutoTextFormatter: Fix crashes with very long numbers

harshad1:
When testing the most recent changes I discovered that messing with very long numbers can cause crashes - this fixes the 2 issues I found


gsantner:
I would includes this as last change for the update and release it then. Though without buzz, 2.10.4.

gsantner:
I would includes this as last change for the update and release it then. Though without buzz, 2.10.4.

Do you have something else around still?

gsantner:
Thank you! I would includes this as last change for the update and release it then. Though without buzz, 2.10.4.

Do you have some other changes around still?

gsantner:
Also fixes  # 1729? Not I guess

gsantner (approved):

harshad1:
> Do you have some other changes around still?

There is an issue I noticed today where inserting a context / project etc causes the editor to scroll to a different location, I will try to get a fix for this out later this evening my time.

gsantner:
Did various improvements and fixes as well these days. Really hope the 
big refactorings don't lead to too much crashes when released.

harshad1:
> Did various improvements and fixes as well these days. Really hope the big refactorings don't lead to too much crashes when released.

I usually run master on my phone for daily use. So far no issues. I think the refactoring was needed - markor feels even better now. Almost like I have a new phone!

PR #1839, opened 2022-09-10
Minimize unnecessary cursor shifting at editor, various fixes & improvements

harshad1:
Am seeing this weird effect - If you put your cursor on the last line in todo.txt and open a dialog (say search) and then the dialog is dismissed (by cancel etc), the textview scrolls away from the line.

Doesn't seem to happen if we select a line which is not the last.


harshad1:
@gsantner Not sure I will get this PR done soon. If it isn't ready by the weekend, please don't wait to release.

This is a minor issue and I will be on vacation for a week :)

app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
```diff
@@ -30,7 +30,7 @@
 import net.gsantner.markor.R;
 import net.gsantner.markor.format.FormatRegistry;
 import net.gsantner.markor.format.plaintext.PlaintextSyntaxHighlighter;
-import net.gsantner.markor.format.todotxt.TodoTxtParser;
```
harshad1, line 33: The TodoTxtParser / TodoTxtTask class is used to both parse a task as well as hold an instance of a parsed task. IMO naming it TodoTxtTask is better.

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtFilter.java
```diff
@@ -37,31 +37,36 @@ public class TodoTxtFilter {
...
-    public static GsCallback.b1<TodoTxtParser> taskSelector(final Collection<String> keys, final GsCallback.r1<List<String>, TodoTxtParser> keyGetter, final boolean isAnd) {
+    public static GsCallback.b1<TodoTxtTask> taskSelector(
+            final Collection<String> keys,
+            final GsCallback.r1<List<String>, TodoTxtTask> keyGetter,
+            final boolean isAnd) {
 
-        final boolean noneIncluded = keys.remove(null);
-        final Set<String> searchSet = (keys instanceof HashSet) ? (HashSet) keys : new HashSet<>(keys);
+        final Set<String> searchSet = new HashSet<>(keys);
+        final boolean noneIncluded = searchSet.remove(null);
```
harshad1, line 69: Don't mutate the collection passed in, always use a new collection and not one which can be modified outside

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtFilter.java
```diff
@@ -180,13 +185,15 @@ public static List<Group> loadSavedFilters(final Context context) {
...
-                    gp.keys.add(keysArray.getString(j));
+                    final String key = keysArray.getString(i);
+                    gp.keys.add(NULL_SENTINEL.equals(key) ? null : key);
                 }
                 loadedViews.add(gp);
             }
             return loadedViews;
         } catch (JSONException e) {
             e.printStackTrace();
+            pref.edit().remove(SAVED_TODO_VIEWS).apply();
```
harshad1, line 196: Nuke saved filters if there is a parsing problem. Otherwise a problem can lead to saved filters never working (always returns empty list).

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -109,8 +109,8 @@ public void afterTextChanged(final Editable s) {
     // ---------------------------------------------------------------------------------------------
 
     private boolean isScrollSignificant() {
-        return Math.abs(_hlRect.top - _olhHlRect.top) > _hlShiftThreshold ||
-                Math.abs(_hlRect.bottom - _olhHlRect.bottom) > _hlShiftThreshold;
+        return (_oldHlRect.top - _hlRect.top) > _hlShiftThreshold ||
+                (_hlRect.bottom - _oldHlRect.bottom) > _hlShiftThreshold;
```
harshad1, line 113: Minimize unnecessary highlighting

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -141,9 +141,12 @@ private void updateHighlighting(final boolean recompute) {
...
                 // we compute the resulting shift and scroll the view to compensate in order to make
-                // the experience smooth for the user
-                final int shiftTestLine = layout.getLineForVertical(_hlRect.centerY());
-                final int oldOffset = layout.getLineBaseline(shiftTestLine);
+                // the experience smooth for the user.
+                int shiftTestLine = -1, oldOffset = -1;
+                if (_scrollView != null && _hlRect.height() == _oldHlRect.height()) {
+                    shiftTestLine = layout.getLineForVertical(_hlRect.centerY());
+                    oldOffset = layout.getLineBaseline(shiftTestLine);
+                }
```
harshad1, line 149: Don't try to correct scroll position when we are changing view size (i.e. keyboard shown)

app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java
gsantner: My opinion: A todo.txt task = 1 line of text, or being model of 1 line.

What this however does is parse, convert, transform stuff so I thought Parser to be fitting.

But at the end, I don't care. If you want to rename it and feel it's bettery it is OK^^.

harshad1: > My opinion: A todo.txt task = 1 line of text, or being model of 1 line.

This is exactly what TodoTxtTask is. It's just that the parsing logic is encapsulated into the same class.

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
```diff
@@ -150,7 +150,7 @@ public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
         // It may cause reads or writes to _silently fail_
         // Instead we try to create it, and exit if that isn't possible
         if (isStateBad()) {
-            Toast.makeText(activity, R.string.document_error_exit, Toast.LENGTH_LONG).show();
+            Toast.makeText(activity, R.string.document_error, Toast.LENGTH_LONG).show();
```
harshad1, line 153: Changed name, not the actual string

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
```diff
@@ -362,15 +362,24 @@ private void updateUndoRedoIconStates() {
...
-        // Only trigger the load process if constructing or file updated
-        if (_document.hasFileChangedSinceLastLoad()) {
+        // Only trigger the load process if constructing or file updated or force reload
+        if (forceReload || _document.hasFileChangedSinceLastLoad()) {
 
             final String content = _document.loadContent(getContext());
+            if (content == null) {
+                errorClipText();
+                return false;
+            }
```
harshad1, line 381: Handle error case

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
```diff
@@ -362,15 +362,24 @@ private void updateUndoRedoIconStates() {
...
+    public boolean loadDocument(final boolean forceReload) {
         if (isSdStatusBad() || isStateBad()) {
             errorClipText();
             return false;
         }
 
-        // Only trigger the load process if constructing or file updated
-        if (_document.hasFileChangedSinceLastLoad()) {
+        // Only trigger the load process if constructing or file updated or force reload
+        if (forceReload || _document.hasFileChangedSinceLastLoad()) {
```
harshad1, line 375: Reload always fetches from disk

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
```diff
@@ -655,6 +664,9 @@ private void setHorizontalScrollMode(final boolean wrap) {
 
             final int[] sel = TextViewUtils.getSelection(_hlEditor);
 
+            final boolean hlEnabled = _hlEditor.getHighlightingEnabled();
+            _hlEditor.setHighlightingEnabled(false);
```
harshad1, line 668: Remove highlighting when changing wrap - improves perf / crash for larger files

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtFilter.java
```diff
@@ -180,13 +185,15 @@ public static List<Group> loadSavedFilters(final Context context) {
                 gp.keys = new ArrayList<>();
                 final JSONArray keysArray = obj.getJSONArray(KEYS);
                 for (int j = 0; j < keysArray.length(); j++) {
-                    gp.keys.add(keysArray.getString(j));
+                    final String key = keysArray.getString(j);
+                    gp.keys.add(NULL_SENTINEL.equals(key) ? null : key);
```
harshad1, line 189: We use a null sentinel to store nulls in the json. Without this, 'no priority', 'no context' etc etc filters weren't working

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -37,7 +38,7 @@
 import net.gsantner.opoc.wrapper.GsTextWatcherAdapter;
 
 @SuppressWarnings("UnusedReturnValue")
-public class HighlightingEditor extends AppCompatEditText {
+public class HighlightingEditor extends AppCompatMultiAutoCompleteTextView {
```
harshad1, line 41: `AppCompatMultiAutoCompleteTextView` Appears to have some performance enhancements. We don't actually use autocompletion or popups

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -139,33 +140,42 @@ private void updateHighlighting(final boolean recompute) {
...
+                int shiftTestLine = -1, oldOffset = -1;
+                if (_scrollView != null && heightSame) {
+                    shiftTestLine = layout.getLineForVertical(_hlRect.centerY());
+                    oldOffset = layout.getLineBaseline(shiftTestLine);
+                }
+
+                if (heightSame) {
+                    // Hack to block bring point into view
+                    // We don't call this when height is changing
+                    blockBringPointIntoView();
```
harshad1, line 157: Refined how the blocking is done. Doing this too much is what lead to the original bug where the cursor would scroll away.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -189,7 +199,7 @@ public void setHighlighter(final SyntaxHighlighterBase newHighlighter) {
 
         if (_hl != null) {
             initHighlighter();
-            _hlDebounced = TextViewUtils.makeDebounced(_hl.getHighlightingDelay(), () -> updateHighlighting(true));
+            _hlDebounced = TextViewUtils.makeDebounced(getHandler(), _hl.getHighlightingDelay(), () -> updateHighlighting(true));
```
harshad1, line 202: Use the HighlightingEditor's handler

app/src/main/java/net/gsantner/markor/frontend/textview/TextViewUtils.java
```diff
@@ -327,23 +330,94 @@ public static void showSelection(final TextView text, final int start, final int
...
+
+    // Due to dynamic highlighting we can't just requestRectangleOnScreen as
+    // when we scroll to the rect, the rect may no longer represent the ROI (layout changes).
+    // So we adopt an iterative scheme - scroll to the rect, check if rect is good and then
+    // repeat until we are satisfied
+    private static class SelectionShower implements ViewTreeObserver.OnScrollChangedListener {
+        final @NonNull TextView _text;
+        final ViewTreeObserver _observer;
+        final int _start, _end;
+        int _iterCount;
```
harshad1, line 347: We iterate a max of iterCount times (using 5 now )

app/src/main/java/net/gsantner/markor/model/Document.java
```diff
@@ -227,7 +238,11 @@ public synchronized String loadContent(final Context context) {
                 content = "";
             }
         } else {
-            final Pair<String, GsFileUtils.FileInfo> result = GsFileUtils.readTextFileFast(_file);
+            // We try to load 2x. If both times fail, we return null
+            Pair<String, GsFileUtils.FileInfo> result = GsFileUtils.readTextFileFast(_file);
+            if (result.second.ioError) {
+                result = GsFileUtils.readTextFileFast(_file);
```
harshad1, line 244: I get this bug once every couple weeks where only the file will only partially load. This is an attempt to mitigate that.

app/thirdparty/java/other/writeily/widget/WrFilesWidgetFactory.java
```diff
@@ -62,11 +63,11 @@ private void updateFiles() {
         } else if (dir.equals(GsFileBrowserListAdapter.VIRTUAL_STORAGE_FAVOURITE)) {
             _widgetFilesList.addAll(ApplicationObject.settings().getFavouriteFiles());
         } else if (dir.exists() && dir.canRead()) {
-            final File[] all = dir.listFiles(file -> true);
+            final boolean showDot = as.isFileBrowserFilterShowDotFiles();
+            final File[] all = dir.listFiles(file -> showDot || !file.getName().startsWith("."));
```
harshad1, line 67: Respect show dotfiles in widget as well

app/src/main/java/net/gsantner/markor/model/Document.java
```diff
@@ -324,13 +347,17 @@ public synchronized boolean saveContent(final Activity context, final CharSequen
...
+                    } catch (Exception e) {
+                        Log.i(Document.class.toString(), e.getMessage());
                     }
                 });
                 success = true;
             } else {
                 success = GsFileUtils.writeFile(_file, contentAsBytes, _fileInfo);
             }
+
+            success &= fileBytes() >= contentAsBytes.length;
```
harshad1, line 359: File size on disk must be greater than content size for success - another defensive check

app/src/main/java/net/gsantner/opoc/format/GsTextUtils.java
```diff
@@ -163,17 +164,6 @@ public static int tryParseInt(final String value, int defaultValue) {
...
 
-    public static <T> ArrayList<T> toArrayList(T... array) {
-        ArrayList<T> list = new ArrayList<>();
-        Collections.addAll(list, array);
-        return list;
-    }
-
-    // Not null, not empty, not spaces only
-    public static boolean ne(final String str) {
-        return str != null && !str.trim().isEmpty();
```
harshad1, line 174: Redundant with `TextViewUtils.isNullOrWhitespace`

app/src/main/java/net/gsantner/opoc/frontend/base/GsFragmentBase.java
```diff
@@ -154,13 +159,26 @@ protected void attachToolbarClickListenersToFragment() {
         }
     }
 
+    public void post(final Runnable action) {
```
harshad1, line 162: Experimental: Ability to post to fragments. Very useful as child fragments are resumed _after_ parent activities.

app/src/main/java/net/gsantner/opoc/format/GsTextUtils.java
gsantner: some issues may come out of simple exchange of it. ne is the opposite.

gsantner: some issues may come out of simple exchange of it. ne is the opposite.

Like this change here: if (badMimeType && file.exists)

Screenshot_20220928-202916

harshad1:
Thanks, yes, I noticed. I am trying to make sure the behaviour matches what is desired. I think the behavior as implemented now in this PR is actually correct. But I am continuing to test.

Sep. 28, 2022 11:29:14 Gregor Santner ***@***.***>:

gsantner:
Can I push something to this PR now, is that a problem?

gsantner:
I push something to master, should be not that hard to merge. My point is, the TextViewUtil method has a rather different use/background then whats needed there. So while at edittext-text saving we need something to check if there is really text, or only whitespace. But for other generic "text not null, text not empty" checks the "not contains any whitespace character" is rather something different, so lets keep it separatly

app/src/main/java/net/gsantner/opoc/util/GsFileUtils.java
```diff
@@ -446,23 +445,62 @@ public static String relativePath(File src, File dest) {
...
+        if (!GsTextUtils.isNullOrEmpty(t = URLConnection.guessContentTypeFromName(file.getName().replace(".jenc", "")))) {
+            return t;
+        }
+
+        // Try extracting by running shell file -b on the file - for textfiles this very often results in "ASCII text"
+        try {
+            final String cmd = "file -b " + file.getAbsolutePath().replace(" ", "\\ ").replace(";", "").replace("&", "").replace("$", "");
+            final Process p = Runtime.getRuntime().exec(cmd);
+            BufferedReader reader = new BufferedReader(new InputStreamReader(p.getInputStream()));
+            t = reader.readLine().trim().toLowerCase(Locale.ROOT);
```
gsantner, line 472: @harshad1 

I made the changes on master and merged it into your branch. I'm done for now.

What is probably the newest best addition is this section here. As even with all the many things Android & Java has to other, it still is by ~10 means somehow incapable to detect files if they are not uber-popular like .jpg or .txt is. 

What I found is that the normal shell `file` utility however works quite well for many things, where the java & android utils fail. So for many plaintext files it just echoes "ASCII text", or  "language script" (like python script).

Now utilizing this info at Markor as well and so hopefully textfiles are way better detected now. I.e. tried it on ".taskpaper" which is also a plain textfile & format like todo.txt is -- everything fails to detect this, but this new one works!

gsantner:
I got a mail that there is for some user a problem with the Undo-Redo buttons and that the app crashes. Didn't run in to it yet, but now the first time saw it.

It's this one here:

Screenshot_20220928-235251

I would just add a line _helper = (helper != null ? helper :  new TextViewUndoRedo() above that line.

Do you think thats good enough?

app/src/main/java/net/gsantner/opoc/util/GsFileUtils.java
harshad1: Very cool. However, we are relying on the `file` shipped with whatever version of android. IMO at this stage we need to consider shipping file or including a library

gsantner: Thats right, but I rather not include some C++/NDKK prebuilt native stuff on my own here 😄 . Lets see for a while how it works out. Looks to me promising that it works well / better then the others (for non-popular filetypes) so far.

harshad1: https://github.com/overview/mime-types should be investigated imo. We have certainly spent a lot of time on mime types

gsantner: it doesn't have a mapping for lets say asciidoc. so even when, we would need some manual mapping still. Just like for `.md.txt` or `.jenc` stuff.

grafik

At the end, the interesting info for Markor is: Is it one of our ~4 "well supported" formats like Markdown, is it some other textfile (=plaintext format), or is it not text and thus some binary. And to current feeling that new file stat call seems to fill that textfile detection gap better then the rest so far :smiley: .

app/thirdparty/java/other/writeily/widget/WrFilesWidgetFactory.java
gsantner: This means the dot filter only applies to dotFiles in real, existing folders. But when you have i.e. a dotFile in the recents or favorites it is still shown.

Thats actually how I like it the most, but just noting that. This means i.e. a favorited `.gitignore` shows up.

harshad1: I agree. Though now that I am looking, is there actually a way to create widgets for favorites etc?

gsantner: yes if you know what you do :D. you can press up->up-> till you reach /storage, and favorites is ""virtual mounted"" at /storage/favorites for example.

harshad1:
Wow lol. Maybe we should make this more discoverable 





---- On Wed, 05 Oct 2022 12:36:39 -0700 Gregor Santner ***@***.***> wrote ---





@gsantner commented on this pull request.


In https://github.com/gsantner/markor/pull/1839#discussion_r988277349:

> @@ -62,11 +63,11 @@ private void updateFiles() {
         } else if (dir.equals(GsFileBrowserListAdapter.VIRTUAL_STORAGE_FAVOURITE)) {
             _widgetFilesList.addAll(ApplicationObject.settings().getFavouriteFiles());
         } else if (dir.exists() && dir.canRead()) {
-            final File[] all = dir.listFiles(file -> true);
+            final boolean showDot = as.isFileBrowserFilterShowDotFiles();
+            final File[] all = dir.listFiles(file -> showDot || !file.getName().startsWith("."));


yes if you know what you do :D. you can press up->up-> till you reach /storage, and favorites is ""virtual mounted"" at /storage/favorites for example.

gsantner:
You need to know, that also the top toolbar button for favorites, at the end does nothing more then set current directory to that virtual path. Ans whenever that path is detected, everything using filebrowser loads from appsettings instead of from listFiles.

Am 5. Oktober 2022 23:19:14 MESZ schrieb Harshad Vedartham ***@***.***>:
>Wow lol. Maybe we should make this more discoverable >

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -109,66 +101,25 @@ public void afterTextChanged(final Editable s) {
...
-                // Addition of spans which require reflow can shift text on re-application of spans
-                // we compute the resulting shift and scroll the view to compensate in order to make
-                // the experience smooth for the user
-                final int shiftTestLine = layout.getLineForVertical(_hlRect.centerY());
-                final int oldOffset = layout.getLineBaseline(shiftTestLine);
-
                 final int[] newHlRegion = hlRegion(_hlRect); // Compute this _before_ clear
-                try {
-                    beginBatchEdit();
-                    blockBringPointIntoView(); // Hack to block bring point into view
```
harshad1, line 151: @gsantner 

After trying to tweak and optimize this for weeks (months?) now, I have given up. Dynamically adding `UpdateLayout` spans causes a million problems in `TextViews` and I would rather work on actually useful features.

I have therefore reverted to a suggestion you made when dynamic highlighting was initially being developed - separate the UpdateLayout spans and apply them statically (i.e. to the whole file) and apply only spans which do not update layout dynamically.

In practice this means that
- TodoTxt (and other formats, but mostly todotxt) will be much less janky
- Larger TodoTxt files will become slower to edit.

On my pixel 3, a 1000 line todo.txt file is definitely slow to edit, 500 lines works fine. I think this is ok for 99.9% of users. (Markdown etc files are essentially unaffected.)

As a side affect, the code is dramatically simpler

gsantner: Hm just some feedback, on git master it seems to work quite well. Not sure how much different way you tried here exactly, but that one looks pretty good to me :-).

harshad1: Yeah. It works well enough 80% of the time (after so many layers of hacks and tweaking). But at least once per editing session I will have the cursor scroll off the screen or the editor will not scroll correctly to the correct line etc etc. It makes working with todo.txt in markor very annoying and breaks my flow when I am trying to plan things.

In practice, I think this new (old) approach is a better compromise - much simpler system and stable system with somewhat lower performance.

gsantner: do you need some testfiles again or do you have it all around?

harshad1:
I have the file with the long line and my test files. All work satisfactorily.

H

Oct. 7, 2022 11:25:03 Gregor Santner ***@***.***>:

harshad1:
Imo this is ready to go

app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
```diff
@@ -982,9 +983,16 @@ public String getMimeType(final Context context, String uri) {
             mimeType = GsFileUtils.getMimeType(new File(uri));
         }
 
+        // Detect extensions which open in this app at plain text
+        // Done only if all other tests fail
+        if (GsTextUtils.isNullOrEmpty(mimeType) && new AppSettings().init(context).isExtOpenWithThisApp(ext)) {
```
gsantner, line 988: Markor AppSettings shouldn't be in GsContextUtils, I will check what I can do here.

app/src/main/java/net/gsantner/markor/model/Document.java
```diff
@@ -196,15 +207,16 @@ public boolean isEncrypted() {
     }
 
     private void setContentHash(final CharSequence s) {
-        _lastLength = s.length();
-        _lastHash = GsFileUtils.crc32(s);
+        _lastLength = s != null ? s.length() : 0;
```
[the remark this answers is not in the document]
harshad1: No. This is for setting the hash based on the actual text. The length is included to make the test very fast in most cases.

app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
harshad1: I guess we can create an `isPlainText` file function somewhere in markor. But I think we have too many mime type / isTextFile functions rn, each with different semantics.

gsantner: How did this check get here? I recall the ext check to be explicitley there where we do also the other extension-to-type association. At converters...and this one being in PlaintextConverter.

harshad1: I added it to make sure snippets and templates showed all possible files.

We could probably just move it out to the actual snippet and template logic for now

gsantner: Is FormatRegistry.isFileSupported()  what you look for?

harshad1: I think that will do. Pushed the change. 

I also think we should work on unifying the various mime type / file type / isTextFile functions. There are way too many now.

app/src/main/res/layout/document__fragment__edit.xml
```diff
@@ -60,19 +60,6 @@
             android:orientation="horizontal" />
     </HorizontalScrollView>
 
-    <TextView
```
gsantner, line 63: this now is replaced by/resulting in a IO-Error instead now, instead of this message. Right?

app/src/main/res/layout/document__fragment__edit.xml
```diff
@@ -60,19 +60,6 @@
...
-        android:id="@+id/document__fragment__edit__content_editor__permission_warning"
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content"
-        android:layout_gravity="bottom"
-        android:background="#F8D507"
-        android:minHeight="@dimen/textactions_bar_height"
-        android:padding="12dp"
-        android:text="@string/warning_disallowed_sd_access"
```
gsantner, line 71: this is replaced by/resulting in a IO-Error now, instead of this error banner. Right?

gsantner:
is it ready for next check? / merge

harshad1:
Yes. It should be done.

harshad1:
@gsantner  I have fixed the way we set default preview state here.

Current behavior is:
- Quicknote and todo.txt open in edit mode in main activity or if launched through the dedicated shortcuts
- If line number is specified we open in edit mode
- If the intent specifies preview state we start in preview
- if the file starts with "index." we start in preview

Otherwise we use the last state

gsantner (approved):

PR #1849, opened 2022-09-18
Remove ambiguity and streamline checklist items in issue templates

opusforlife2:
That last checklist item is ambiguous. Should you check the box or not if the answer is yes or no? I've changed the wording to make it more straightforward, and also harmonised the two versions in each template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
@@ -18,7 +18,7 @@ body:
           required: true
         - label: Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub. 
           required: true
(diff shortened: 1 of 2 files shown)


opusforlife2:
Changes to files in the `.github` folder, and maybe others, could be excluded from the CI.

gsantner:
The reason it is currently written as it is, is because it's rather ""if you are able to do it, do it please"". Might be that not everybody is able to install some debug apk, or use ADB.

> Changes to files in the .github folder, and maybe others, could be excluded from the CI.

Most changes happen to code. Thats rather a optimization for the very fast minority. Then with some CI configuration format change it suddenly not works anymore and breaks something else. Don't see the need for this specifically.

opusforlife2:
> "if you are able to do it, do it please"

I would suggest stating that explicitly, then. Users sometimes come up with the wildest imaginable interpretations of instructions.

gsantner:
Yeah well unfortunatley reporters often just tick boxes without really doing anything of them.

Or ignoring templates and writing just something without much info.

Or leave out many details.

opusforlife2:
Over at Newpipe we use a `template-ignored` label and close such issues, if you're considering cracking down on such behaviour.

opusforlife2:
Too true.

PR #1869, opened 2022-10-02
Post action to fragment, fixes FAB crash, by arshad1

gsantner:
Pre-merge the fragment posting change from #1839 as that fixes the biggest issue I can currently spot in Google Play dashboard.

image


gsantner:
@harshad1  
Pre-merging that fragment portion as found that makes up for the majority of issues. As not sure how much you still want to do at #1839 and when you think it's ready

harshad1:
I keep thinking it's ready and then find issues.

I'll try to get it done today. Getting annoyed with how long it's taking :)

H

Oct. 2, 2022 05:39:22 Gregor Santner ***@***.***>:

gsantner:
No hurry

PR #1883, opened 2022-10-06
handle file intent by mimetype rather than file extension

### LINKED ISSUE #1881: Open *.xml files

- [X] The bug is not present in the latest code version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Mordor**. Don't worry; Markor and Mordor appear as completely separate applications. You can install both side-by-side, and Markor settings are not touched. In case the issue is resolved there, you don't need to create a bug report. The change will be part of the next Markor update.)

### Description

Opening *.xml files doesn't give markor as option. I would expect markor to be an option.

### Steps to reproduce

Open any *.xml file with e.g. file browser.

### Information

Android version: 12
App Version: 2.10.5

### Source

F-Droid

### Format / File type

Plaintext

### Additional info / Log

_No response_

fabian-thomas:
Fixes #1881.

I've taken a look how [Simple Editor](https://github.com/billthefarmer/editor) handles mime types and file extension associations. I think their approach is superior. Rather than having an app hardcoded list of supported file extensions, they filter only by mimetype `text/*`, thereby giving the control of deciding which type the file has to the OS. I further think that it is a good idea that any non-binary file can be opened by my text editor.

Please review where I have added `text/*`. I don't really understand what the first occurence (line 201) does. Also note, that the SEND intent still uses `text/plain`.


gsantner:
If you have a file for example `something.ldg` there is a good chance it's not recognized as text mime-type. While it might work good for many things, for many other it's not. So please, don't remove whats there just because you think it's unnecassary. Mime types and their recognized very very often don't work well, and so what other applications tell about a file when they make a intent.

If you want to add something, fine, but don't drop the existing explicit mappings.

gsantner (changes requested):
If you have a file for example `something.ldg` there is a good chance it's not recognized as text mime-type. While it might work good for many things, for many other it's not. So please, don't remove whats there just because you think it's unnecassary. Mime types and their recognized very very often don't work well, and so what other applications tell about a file when they make a intent.

If you want to add something, fine, but don't drop the existing explicit mappings.

fabian-thomas:
@gsantner Ok, I guess we can leave the file extension checks then. But we should replace all mime type matches of text with one `text/*`.

I'm gonna add new extension checks for xml and some other extensions tomorrow. I'm currently thinking of: .xml, .tsv, .atom, .rss

gsantner:
Added the wildcard to all intent-filters, and well Mordor/Markor appears at last from the AOSP filemanager when clicking on a xml file.

screenshot-2022-10-07__00-17-32

So guess it doesn't need other removals/additions, unless you notice an issue with that.

PR #1884, opened 2022-10-06
fix confusing typo in bug template

fabian-thomas:


gsantner:
It is not a typo. By checking you should confirm that the bug IS NOT 
present on the git version.

fabian-thomas:
@gsantner Ok, please explain that further. I had to check that box to create my issue. What is the point of that box?

gsantner:
That you download, install and try the latest development of Markor, built from git master.

fabian-thomas:
But that's exactly what was wrong.

The template says:
I confirm that the bug is *not* on master.

And now it says:
I confirm that the bug is *still* on master.

gsantner:
I've cleaned it up in both templates, albeit differently than that one word change

PR #1898, opened 2022-10-13
correction of the translation into brazilian portuguese

ricardo-ramos-moura:
Correction of the translation into brazilian portuguese. Tested on Samsung Galaxy physical device and Pixel emulator.


gsantner:
SimpleDateFormat for example must not be translated.

https://developer.android.com/reference/java/text/SimpleDateFormat

Users need to find SimpleDateFormat, not whatever "simple + date + format" means in some other language.

gsantner:
also generally thanks for translating, but translations are done on the translation system, not here. What happens here will be overwritten.

-->> <https://crowdin.com/project/markor/invite>

ricardo-ramos-moura:
> also generally thanks for translating, but translations are done at the translation system, not here. What happens here will be overwritten.
> 
> -->> https://crowdin.com/project/markor/invite

ok man. If I can help with any activity, I'm available.

PR #1901, opened 2022-10-17
Todo.txt advanced search system

harshad1:
In this PR I have created a fully functional advanced search / filter system for todo.txt

For example `@home & (pri:A | pri:B)` will return tasks with priority A or B and which have the context @home. The existing filter system has been updated to work on top of this query language (i.e. by generating the appropriate query).

This is now fully functional and works correctly.

Remaining tasks:
- [x] Document the query language
- [ ] Add more unit tests
- [x] More testing

Other fixes in this PR include:
1. Done file is now remembered per file (each todo.txt file can have it's own done file)
2. If the keyboard was open when a dialog was opened, we will open the keyboard again when the dialog is closed (this is only enabled in some dialogs where it seemed to be useful / easy to implement)

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtFilter.java
+++ b/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtFilter.java
@@ -228,7 +228,8 @@ public class TodoTxtFilter {
             final CharSequence expression = parseQuery(task, processed);
             final boolean accept = evaluateExpression(expression);
             return accept;
-        } catch (EmptyStackException e) {
+        } catch (EmptyStackException | IllegalArgumentException e) {
+            // TODO - display a useful message somehow
             return false;
         }
     }
@@ -354,7 +355,7 @@ public class TodoTxtFilter {
             } else if (op == '!') {
                 stack.push(toChar(rhs == 'F'));
             } else {
-                return;
+                throw new IllegalArgumentException("Unexpected character");
             }
         }
     }
@@ -366,7 +367,7 @@ public class TodoTxtFilter {
             if (c == ')') {
                 final char value = stack.pop();
                 if (stack.pop() != '(') {
-                    return false; // Bad expression
+                    throw new IllegalArgumentException("Mismatched parenthesis");
                 }
                 stack.push(value);
             } else {
@@ -374,6 +375,9 @@ public class TodoTxtFilter {
             }
             evaluateOperations(stack);
         }
-        return stack.size() == 1 && stack.pop() == 'T';
+        if (stack.size() == 1 && isValue(stack)) {
+            return stack.pop() == 'T';
+        }
+        throw new IllegalArgumentException("Malformed expression");
     }
 }
--- a/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtFilter.java
+++ b/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtFilter.java
@@ -24,7 +24,6 @@ import java.util.List;
 import java.util.Set;
 import java.util.Stack;
 import java.util.TreeSet;
-import java.util.regex.Matcher;
 
 public class TodoTxtFilter {
 
--- a/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtFilter.java
+++ b/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtFilter.java
@@ -24,18 +24,19 @@ import java.util.List;
 import java.util.Set;
 import java.util.Stack;
 import java.util.TreeSet;
+import java.util.regex.Matcher;
 
 public class TodoTxtFilter {
 
     // Special query keywords
     // ----------------------------------------------------------------------------------------
 
-    public final static String QUERY_PRIORITY_NONE = Character.toString(TodoTxtTask.PRIORITY_NONE);
+    public final static String QUERY_PRIORITY_ANY = "pri";
     public final static String QUERY_DUE_TODAY = "due=";
     public final static String QUERY_DUE_OVERDUE = "due<";
     public final static String QUERY_DUE_FUTURE = "due>";
     public final static String QUERY_DUE_ANY = "due";
-    public final static String QUERY_DONE = "x";
+    public final static String QUERY_DONE = "done";
 
     // ----------------------------------------------------------------------------------------
 
@@ -122,8 +123,8 @@ public class TodoTxtFilter {
             nullKey = "!+";
             prefix = "+";
         } else if (type == TYPE.PRIORITY) {
-            nullKey = QUERY_PRIORITY_NONE;
-            prefix = "";
+            nullKey = "!" + QUERY_PRIORITY_ANY;
+            prefix = "pri:";
         } else { // type due
             nullKey = "!" + QUERY_DUE_ANY;
             prefix = "";
@@ -258,14 +259,18 @@ public class TodoTxtFilter {
     }
 
     // Evaluate a word (element) for truthyness or falsiness
+    // Step through all the possible conditions
     private static char evalElement(final TodoTxtTask task, final String element) {
 
-        // Prioritiy
         final boolean result;
-        if ((element.length() == 1) && element.matches("[A-Z]")) {
-            result = Character.toLowerCase(task.getPriority()) == Character.toLowerCase(element.charAt(0));
-        } else if (QUERY_PRIORITY_NONE.equals(element)) {
-            result = task.getPriority() == TodoTxtTask.PRIORITY_NONE;
+        if (element.startsWith(QUERY_PRIORITY_ANY)) {
+            if (QUERY_PRIORITY_ANY.equals(element)) {
+                result = task.getPriority() != TodoTxtTask.PRIORITY_NONE;
+            } else if (element.length() == 5 && element.charAt(3) == ':') {
+                result = task.getPriority() == element.charAt(4);
+            } else {
+                result = false;
+            }
         } else if (QUERY_DUE_TODAY.equals(element)) {
             result = task.getDueStatus() == TodoDueState.TODAY;
         } else if (QUERY_DUE_OVERDUE.equals(element)) {
--- a/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtFilter.java
+++ b/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtFilter.java
@@ -2,9 +2,13 @@ package net.gsantner.markor.format.todotxt;
 
 import android.content.Context;
 import android.content.SharedPreferences;
+import android.util.Pair;
 import android.widget.Toast;
 
+import androidx.annotation.VisibleForTesting;
+
 import net.gsantner.markor.R;
+import net.gsantner.markor.format.todotxt.TodoTxtTask.TodoDueState;
 import net.gsantner.opoc.model.GsSharedPreferencesPropertyBackend;
 import net.gsantner.opoc.wrapper.GsCallback;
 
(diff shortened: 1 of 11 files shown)


app/src/main/java/net/gsantner/markor/frontend/MarkorDialogFactory.java
```diff
@@ -47,6 +50,7 @@
 import net.gsantner.markor.frontend.textview.TextViewUtils;
 import net.gsantner.markor.model.AppSettings;
 import net.gsantner.opoc.frontend.GsSearchOrCustomTextDialog;
+import net.gsantner.opoc.frontend.GsSearchOrCustomTextDialog.DialogOptions;
```
harshad1, line 53: As we use this class in dozens of places in this file, I imported it directly in order to make the code look cleaner

app/src/main/java/net/gsantner/markor/frontend/MarkorDialogFactory.java
```diff
@@ -233,74 +234,66 @@ public static void showSttSortDialogue(Activity activity, final GsCallback.a2<St
         GsSearchOrCustomTextDialog.showMultiChoiceDialogWithSearchFilterUI(activity, dopt);
     }
 
-    public static void showSttContextDialog(Activity activity, List<String> availableData, GsCallback.a1<String> callback) {
```
harshad1, line 236: Unified the context and project dialogs. They shared 99% of the same code

app/src/main/java/net/gsantner/opoc/frontend/GsSearchOrCustomTextDialog.java
```diff
@@ -86,12 +87,12 @@ public static class DialogOptions {
         public int dialogHeightDp = WindowManager.LayoutParams.WRAP_CONTENT;
         public int gravity = Gravity.NO_GRAVITY;
         public int searchInputType = InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS;
-        public boolean searchIsRegex = false;
         public GsCallback.a1<Spannable> highlighter = null;
         public String extraFilter = null;
         public List<Integer> preSelected = null;
-
         public GsCallback.a1<AlertDialog> neutralButtonCallback = null;
+        public GsCallback.b2<CharSequence, CharSequence> searchFunction = GsSearchOrCustomTextDialog::standardSearch;
```
harshad1, line 94: Made the search / filter function configurable

app/src/main/java/net/gsantner/opoc/frontend/GsSearchOrCustomTextDialog.java
```diff
@@ -389,19 +395,20 @@ private static View makeTitleView(final Context context, final DialogOptions dop
         titleLayout.setOrientation(LinearLayout.VERTICAL);
         titleLayout.setLayoutParams(new LinearLayout.LayoutParams(LinearLayout.LayoutParams.MATCH_PARENT, LinearLayout.LayoutParams.WRAP_CONTENT));
         titleLayout.setGravity(Gravity.CENTER_VERTICAL | Gravity.START);
-        titleLayout.setPadding(paddingSide, paddingSide, paddingSide, paddingBetween);
+        titleLayout.setPadding(paddingSide, 2 * paddingBetween, paddingSide, paddingBetween);
```
harshad1, line 398: Added a few tweaks to the title display layout

harshad1:
# Description of the advanced filter syntax. 

I will add this to the source and a few other places soon.

The todo.txt advanced filter syntax allows one to filter todo items using a simple query language.

For example `(A or B) and +home` will match  tasks with priority `A` or priority `B` and which have the project `+home`

## Filter types

- **Priority**: Use a single capital letter (A through Z). The keyword 'nopriority' or 'nopri' will match tasks with no priority
- **Context**: Words starting with '@' will match contexts. The keyword 'nocontext' or 'nocontexts' will match tasks with no context
- **Project**: Words starting with '+' wll match projects. The keyword 'noproject' or 'noprojects' will match tasks with no project
- **Due date**:
    - 'nodue' will match tasks with no due date
    - 'today' or 'due' will match tasks with due date today
    - 'overdue' or 'past' will match tasks with due date which has already passed
    - 'future' will match tasks which are due in the future
- **Done**: the word `done` will match done tasks 

## Operators
- **Not**: ` not `, ` NOT ` or `!` can be used as a 'not' to negate the following condition / expression. Not that `!` requires no spaces on either side, spelling out the word requires spaces to the right and either a space or the start of an expression to the left
    - `(!A and due)` is valid
    - `(not A and due)` is valid
    - `(notA and due`) is not valid (note lack of space)
- **And**: ` and `, ` AND ` , `&&` or `&` can be used to 'and' two conditions
- **Or**: ` or `, ` OR `, `||` or `|` can be used to 'or' two conditions
- **Parenthesis**: `(` and `)` can be used to group expressions

## Evaluation
The filter expression is evaluated left to right (i.e. 'and' and 'or' have the same precedence). 'Not' is applied to the result of the expression to it's immediate right.

Done tasks are not excluded by default. They must be excluded explicitly.

### Examples:

- `nocontext and noproject and nopriority and not done` - should be self-evident :)
- `nocontext or noproject and not done` - will match tasks which have either no projects or no contexts and which are not done
    - i.e. this is equivalent to `(nocontext or noproject) and not done`

## A note on implementation:

- We internally use `&`, `|`, `!` to represent 'and', 'or' and 'not' all the other equivalents are translated into these in a preprocessing step
- Spaces are then removed. `(C|D|nopriority)&!done` is perfectly valid. Spaces are only needed to handle the long form variants of the operators (` and `, ` not ` etc)
- Each word of the expression is evaluated to construct a boolean expression. For example the expression `(C | D | nopriority) & !done` applied to the todo `D 2022-01-01 foo` will evaluate to the string `"( F | T | F ) & !F"` (spaces added here for clarity)
- The boolean expression is then evaluated to 'T' or 'F' using a simple stack based algorithm
- And errors (syntax errors, mismatched parenthesis etc) will cause a task to evaluate to 'F' (i.e. false)

harshad1:
This is basically ready to go. TBD:

- [ ] Clean up the description above and put it somewhere accessible.
- [ ] Link to the description from the advanced search menu
- [ ] Make saved filters copyable
- [ ] Add a few example filters

gsantner:
My suggestion: Reduce the amount of possible operators to minimum. This makes it later more easy to migrate to something else if required, and to support by other means (uh oh, suddenly somebody comes up with the idea to use this operator stuff somewhere else?!).

What happens when "or" "and" "AND" "OR" are reserved keyword is also that they are excluded from being using (you can't search for todos containing "and"). With it also comes, how to deal with translations, would a german user type "blau oder gruen und gelb"?

So I suggest to keep it simple: Boolean operators, no **logical** keywords generally. For todo.txt specific case with "today", "nopriority" - this is fine, and we also can't magically invent a operator for them :smiley:.

gsantner (commented):
My suggestion: Reduce the amount of possible operators to minimum. This makes it later more easy to migrate to something else if required, and to support by other means (uh oh, suddenly somebody comes up with the idea to use this operator stuff somewhere else?!).

What happens when "or" "and" "AND" "OR" are reserved keyword is also that they are excluded from being using (you can't search for todos containing "and"). With it also comes, how to deal with translations, would a german user type "blau oder gruen und gelb"?

So I suggest to keep it simple: Boolean operators, no **logical** keywords generally. For todo.txt specific case with "today", "nopriority" - this is fine, and we also can't magically invent a operator for them :smiley:.

harshad1:
Yeah. I see your point. I'll make the change when I am able - probably not this week :(

I thought nopriority etc were acceptable as we already use 'due' etc as keywords

gsantner:
No stress, and make it like you want it. It's my suggestion, not the must.

For due etc I think thats fine

harshad1:
I've updated this to reduce the reliance on special words. Still not happy with `~` for no priority and `x` for done. Will need to think more about this.

harshad1:
I know. I just want to get it to a stage where I am satisfied (after all, I am building this for me :p)

harshad1:
Updated. I think I am happy with this form.

Of interest: I found this a couple days ago: https://github.com/ransome1/sleek/wiki/Filter-Expressions-for-Advanced-Search

The syntax is not exactly the same, but it is close enough for comparison / inspiration

app/src/main/java/net/gsantner/markor/model/AppSettings.java
```diff
@@ -355,6 +351,22 @@ public void toggleFavouriteFile(File file) {
     private static final String PREF_PREFIX_AUTO_FORMAT = "PREF_PREFIX_AUTO_FORMAT";
     private static final String PREF_PREFIX_VIEW_SCROLL_X = "PREF_PREFIX_VIEW_SCROLL_X";
     private static final String PREF_PREFIX_VIEW_SCROLL_Y = "PREF_PREFIX_VIEW_SCROLL_Y";
+    private static final String PREF_PREFIX_TODO_DONE_NAME = "PREF_PREFIX_TODO_DONE_NAME";
+
+    public void setLastTodoDoneName(final String path, final String name) {
```
gsantner, line 356: setPerFileTodoTxtDoneFilepath ?

app/src/main/java/net/gsantner/markor/frontend/textview/TextViewUtils.java
```diff
@@ -637,4 +640,12 @@ public static boolean checkRange(final CharSequence seq, final int... indices) {
     public static boolean checkRange(final int length, final int... indices) {
         return indices != null && indices.length >= 2 && inRange(0, length, indices) && indices[1] > indices[0];
     }
+
+    // Check if keyboard open. Only available after android 11 :(
```
gsantner, line 644: I have had something around for older android versions, but it very often is/was unreliable - virtual keyboard detection is horrible, better not trust it generally

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtFilter.java
```diff
@@ -15,114 +19,142 @@
...
-import java.util.HashSet;
+import java.util.EmptyStackException;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
+import java.util.Stack;
+import java.util.TreeSet;
+import java.util.regex.Matcher;
 
 public class TodoTxtFilter {
```
gsantner, line 29: just to be clear - basically the new filtering complete replaces the previous todo filter stuff including filter saving by you right? So need to make sure we use new preference keys for the stuff and get rid of the existing ones

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtFilter.java
```diff
@@ -197,4 +216,143 @@ public static List<Group> loadSavedFilters(final Context context) {
         }
         return Collections.emptyList();
     }
+
+    // Query matching
+    // -------------------------------------------------------------------------------------------
+
+    public static boolean isMatchQuery(final TodoTxtTask task, final CharSequence query) {
```
gsantner, line 223: may I suggest to move the query matching stuff together to an own **inner** class  there?

harshad1: Hmmm, what would be benefit of this? IMO they are all nicely contained in this TodoTxtFilter class (the entire file is query matching, saving, and related things).

app/src/main/java/net/gsantner/markor/frontend/textview/TextViewUtils.java
harshad1: Yeah, that's why I have it set to use a Boolean - true, false and null for indeterminate

app/src/main/java/net/gsantner/markor/model/AppSettings.java
harshad1: We force the done file to be in the same folder as the todo.txt file, so it is a name and not a path. And the other per file functions are all setLast*. We can change this, but I think we should change them all if we change this one.

gsantner: Oh I thought you did some change to select the file from anywhere, didn't check that

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtFilter.java
gsantner: my feeling just tells me that the matching stuff might might be a whole own bigger thing at some point

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtFilter.java
harshad1: The `loadSavedViews` function will delete all the saved filters / views if it encounters a format it can't parse. This means that

1. When this pr is deployed, all the existing filters will get deleted on first load of the filter dialog
2. If there is malformed json (not sure how this could happen), the filters would be reset to a clean (empty state) rather than the dialog closing / not working.

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtFilter.java
harshad1: :). For what it's worth, this matching query class is pretty small. 300 lines for the whole parser / matcher etc etc.

How would you separate the inner class? Right now, TodoTxtFilter contains:

1. Routines for making a query from the various filter dialogs
2. Routines for saving and loading queries
3. Routines for parsing a query

How would you suggest these be split?

IMO there is no clean split and and they are best left as is - a static class with related static functions.

harshad1:
@gsantner Not sure if you are waiting for me to address any issues here. Please let me know if you are.

Thanks
H

gsantner (approved):

PR #1902, opened 2022-10-17
Adjust the value of web. to net.

ricardo-ramos-moura:
Adjust in file AndroidManifest.xml to "net.gsantner.markor.activity.MainActivity"

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -179,7 +179,8 @@
             tools:targetApi="lollipop">
             <meta-data
                 android:name="android.support.PARENT_ACTIVITY"
-                android:value="web.gsantner.markor.activity.MainActivity" />
+                android:value="net.gsantner.markor.activity.MainActivity" />
+
 
             <intent-filter>
                 <action android:name="android.intent.action.SEND" />


harshad1:
What would the effect of this be? (My android dev experience outside markor is pretty minimal)

gsantner:
The top-left activity back button uses that packagename to detect "where to go up to". But we don't have the arrow enabled anywhere so don't know any practical difference either.

(setUpEnabled/setUpAsHomeEnabled/android.R.id.home)

gsantner:
??

ricardo-ramos-moura:
@gsantner I opened this PR because I was debugging on Android version 7.1 (Bug in this android version)
and I noticed it was pointing to a non-existent package.

In the project itself it doesn't affect anything in the execution, I just pointed to the right package for a
question of organization and leave the imports as they should be.

Sorry if I caused any inconvenience, I was just trying to help.

gsantner:
No inconvenience, its helpful, thanks!
 But when you open a merge request without giving a description what it is about, and whats the problem, probably nobody else knows either.

ricardo-ramos-moura:
Yes you are right. My mistake.

PR #1914, opened 2022-11-11
Open file location from search dialog

harshad1:
Long press filename to open location


app/src/main/java/net/gsantner/opoc/frontend/filebrowser/GsFileBrowserListAdapter.java
```diff
@@ -468,8 +486,6 @@ public void loadFolder(final File folder) {
             @Override
             public void run() {
                 synchronized (LOAD_FOLDER_SYNC_OBJECT) {
-                    ArrayList<File> oldAdapterData = new ArrayList<>();
-                    Collections.copy(_adapterData, oldAdapterData);
```
harshad1, line 472: These lines did nothing?

gsantner: They did and it was keeping for data backup. But I generified the logic in GsFileUtils so this is leftover ^^

gsantner:
Really still WIP ?

harshad1:
Yeah. Doesn't work often due to timing issues. (Folder not loaded yet when you try to find item etc)

I have some ideas which I haven't had time to implement.

Nov. 19, 2022 15:49:17 Gregor Santner ***@***.***>:

gsantner:
OK no, normal question, no hurry!

C0D3D3V:
Just an random idea, that is maybe nonsense, because I know nothing about the code base: Maybe you could use a callback on `setCurrentFolder` and execute the scroll code inside the callback instead of inside the `synchronized` block. For me it now jumps in the correct subdirectory but does not scroll to the file position, even if it is already in the correct directory.

harshad1:
Yeah, you're on the right track. I am working on using android's callbacks to do this correctly. I expect to be done with my latest version by this weekend. Would be good to test then.

Nov 24, 2022 06:53:24 Daniel Vogt ***@***.***>:

app/thirdparty/java/other/writeily/model/WrMarkorSingleton.java
```diff
@@ -49,17 +50,13 @@ public void setNotesLastDirectory(File notesLastDirectory) {
         WrMarkorSingleton.notesLastDirectory = notesLastDirectory;
     }
 
-    // Returns true if b is not a child of a. A file is not a child of itself
-    private boolean notChild(final File a, final File b) {
-        try {
-            return b.equals(a) || !b.getParentFile().getCanonicalPath().startsWith(a.getCanonicalPath());
-        } catch (IOException e) {
```
harshad1, line 56: bug - `foo/abc` would be detected as a parent of `foo/abcde`

harshad1:
This now works, but as it relies on several postDelayed timings it should be tested on other devices. My current device is pretty fast (got myself a new Pixel 7 pro :p) so results may vary on other devices. Some cleanup likely still needed.

C0D3D3V:
For me it works on my OnePlus 3T. I tested it for the last 3 days.

harshad1:
@gsantner It looks like this works. Its ready for review and merge. We can improve it if people report that it doesn't work on some devices or something.

harshad1:
Hmmm. None of that was touched in the changes I made. Looks like a side effect :(.

I will have a little time this weekend. Will try to investigate.

Dec 8, 2022 02:11:18 Gregor Santner ***@***.***>:

gsantner (approved):

gsantner:
thanks, merged

PR #1930, opened 2022-12-04
Zim: Fix NPE for non-resolvable relative links with unreachable notebook dir (closes #1882) + Add compatibility support for anchor links

### LINKED ISSUE #1882: zim: links within a page to a heading/object result in NullPointer exception

- [X] The bug is not present in the latest code version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Mordor**. Don't worry; Markor and Mordor appear as completely separate applications. You can install both side-by-side, and Markor settings are not touched. In case the issue is resolved there, you don't need to create a bug report. The change will be part of the next Markor update.)

### Description

A zim document containing a page internal link (starting with '#', resolves to a header or link) creates an exception in view mode:
`java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.io.File.equals(java.lang.Object)' on a null object reference`

Minimal example: [Markor_Zim_Pagelinks.txt](https://github.com/gsantner/markor/files/9727237/Markor_Zim_Pagelinks.txt)

The `[[#42]]` creates a link to the header `== 42 ==`

### Steps to reproduce

Open a zim page with a page-internal link. A minimal example is attached in the description.

### Information

Android version: 8.1
Device: Moto G5 Plus
App Version: v2.10.5 (141)

### Source

Google Play

### Format / File type

Wikitext, Zim

### Additional info / Log

```shell
-
```

fredericjacob:
This PR fixes the issue described in #1882 by introducing an additional null-check. In case that a relative page link cannot be resolved AND the notebook directory is set incorrectly/not reachable from the current page, Markor stops the traversal now at the root of the filesystem and doesn't throw an exception any more (which could be observed e.g. when entering view mode for pages with such links).

Additionally, compatibility support for links with inner page references is added. Zim 0.74.0 introduced "link anchors", which can be inserted somewhere in a page, and referenced and jumped to directly by writing something like `[[#my-anchor]]` or `[[+MyPage:OtherPage#some-heading-in-the-page]]`. Pages to these links could not be resolved in Markor up to now. Still, there's no support for jumping to these anchors - the anchor parts are in fact just ignored for now - but it is now possible again to open the referenced page of the link.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/wikitext/WikitextLinkResolver.java
+++ b/app/src/main/java/net/gsantner/markor/format/wikitext/WikitextLinkResolver.java
@@ -1,5 +1,6 @@
 package net.gsantner.markor.format.wikitext;
 
+import net.gsantner.opoc.format.GsTextUtils;
 import net.gsantner.opoc.util.GsFileUtils;
 
 import org.apache.commons.io.FilenameUtils;
@@ -74,7 +75,7 @@ public class WikitextLinkResolver {
 
         // inner page references are not yet supported - for compatibility just get the page
         wikitextPath = stripInnerPageReference(wikitextPath);
-        if ("".equals(wikitextPath)) {
+        if (GsTextUtils.isNullOrEmpty(wikitextPath)) {
             return null;
         }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/net/gsantner/markor/format/wikitext/WikitextLinkResolver.java
```diff
@@ -71,6 +72,12 @@ private String resolveWikitextPath(String wikitextPath) {
             _isWebLink = false;
         }
 
+        // inner page references are not yet supported - for compatibility just get the page
+        wikitextPath = stripInnerPageReference(wikitextPath);
+        if ("".equals(wikitextPath)) {
```
gsantner, line 77: GsTextUtils.isNullOrEmpty() ?

gsantner (approved):
thanks!

PR #1942, opened 2022-12-28
Add new Zim checkbox state "left arrow"

fredericjacob:
With Zim 0.74.0 (see [Zim changelog](https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/develop/CHANGELOG.md)), a new checkbox state was introduced: a left arrow ("back arrow") with the syntax `[<] some item`.
This PR adds support for the new state by:
* considering the new state in all relevant regexes
* highlighting it correctly
* include the state when toggling the checkbox state (which works like in the Zim desktop application, in the order "unchecked", "checked", "crossed", "right arrow", "left arrow", "unchecked")


gsantner:
Thanks for improving compatibility also for latest features 😄

fredericjacob:
Thanks for merging! 🙂

PR #1949, opened 2023-01-07
Show which mode an action applies to

harshad1:
In the action order activity, show which mode (view vs edit) an action is used for


gsantner:
does dropping ANY and having duplicate(ishL actions/items for view and edit mode aomething better?

harshad1:
That _may_ be a better approach. But I am not sure. What we have now seems to work fine. This PR just makes it somewhat clearer when rearranging icons.

image

gsantner:
What do you think of this generally, that the actions buttons being available in view mode? 

Currently the amount isn't that big, but would generally want to add more such functions, in terms of buttons, to view mode.

Min the Table of contents or jump top/bottom was a very often asked thing, and now its not anymore hidden behind some special button like top bar

harshad1:
IMO making things explicit / not hidden is a good thing. I have used markor every day for many years now and even I can't remember which thing to long press for some useful thing.

That said, I don't use view mode very much.

gsantner (approved):

PR #1952, opened 2023-01-09
Remove REQUEST_INSTALL_PACKAGES for Google Play builds, disallowing APK installation from filemanager

gsantner:
Remove REQUEST_INSTALL_PACKAGES from Android permissions for Google Play flavor

Even though it's a valid reason to install APKs from a filemanager, Google denies Markor this permission.

This uses dominance to restrict apps from making installation request for Android apps without phoning home to Google Play.
Thus only allowing to install apps and updates from Google Play.

On all non-Google Play builds of Markor, the default & test flavor, thus everywhere including F-Droid and GitHub apks,
Markor will still contain functionality to request apk installation by clicking a apk file at builtin filemanager.

Because of this, currently Google blocks releasing any new version of Markor. 

-----

Message from Google Play

Screenshot_20230110-000054


app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
```diff
@@ -1368,10 +1368,23 @@ public boolean requestApkInstallation(final Context context, final File file) {
...
 
         if (fileUri != null) {
-            final Intent intent = new Intent(Build.VERSION.SDK_INT >= Build.VERSION_CODES.N ? Intent.ACTION_INSTALL_PACKAGE : Intent.ACTION_VIEW)
-                    .setFlags(Build.VERSION.SDK_INT >= Build.VERSION_CODES.N ? Intent.FLAG_GRANT_READ_URI_PERMISSION : Intent.FLAG_ACTIVITY_NEW_TASK)
-                    .setDataAndType(fileUri, "application/vnd.android.package-archive");
-            startActivity(context, intent);
+            final String MIME_TYPE_APK = "application/vnd.android.package-archive";
+
+            boolean hasRequestInstallPackagesPermission = true;
+            try {
```
gsantner, line 1374: Note: I tried with PackageManager to check permission granted, but it would not work for this case. When the permission is defined or not in the Manifest the check would always result same.

However, canRequestPackageInstalls() always throws a exception when permission not present in manifest. And not throws one if it's present. Thus this used for detecting permission granted.

PR #1956, opened 2023-01-17
Copy to clipboard from filebrowser

DumnychDU:
Copy to clipboard from filebrowser


gsantner:
I don't see any app code changes in the merge request,

Issue #1886, opened 2022-10-08
Error: Could not open file / File permission granting issue


opusforlife2:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**. Not a [question or feature request](https://github.com/gsantner/markor/issues/new/choose).
- [X] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [X] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [X] The bug is still present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Mordor**. Don't worry; Markor and Mordor appear as completely separate applications. You can install both side-by-side, and Markor settings are not touched. In case the issue is resolved there, you don't need to create a bug report. The change will be part of the next Markor update.)

### Description

If you manually remove the files permission for Markor and then try to open it, the system shows a toast saying "Error: Could not open file."

Then, if you grant the permission again, Markor works normally.

This doesn't occur when the app is opened after the first install. It correctly asks for permission then. Which means revocation of the permission is a scenario that needs to be handled separately.

### Steps to reproduce

1. Manually revoke files permission for Markor.
2. Try opening Markor.
3. System toast appears saying can't open.

### Information

Android version: 12 (LOS 19.1)
App Version: 2.10.5

### Source

F-Droid

### Format / File type

Not specific

gsantner:
> 2. Don't use Markor for a few days (no idea how many).

Thats rather hard to reproduce^^.


But what sure would be helpful is if you can post a log related to the crash when you run into this again.

opusforlife2:
Okay, I don't know why I didn't think of this before. I just tried manually revoking the permission and got the same results. It's fully reproducible.

And I paid closer attention this time. I can't say for certain that there is a crash. I don't get that standard dialogue saying the app has crashed.

Instead, I get a toast with the system app icon (that white android on a green background) saying "Error: Could not open."

Restoring the permission lets Markor open again, like normal. No side effects of this behaviour are apparent.

gsantner:
@opusforlife2   
How do you **remove** the permission grant **manually**? I don't see any storage related permission or anyting at all. But app has asked before and can access just fine.

[...]

PR #1957, opened 2023-01-21
Do not merge: Permission crash fixes

harshad1:
This attempts to address #1886

I made several changes and it appears to work on my devices. Needs testing by affected users and general testing to make sure I didn't break something else.


app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
```diff
@@ -173,30 +173,41 @@ protected void onNewIntent(Intent intent) {
     private void handleLaunchingIntent(final Intent intent) {
```
harshad1, line 173: Reorganized a little. Improved error message when file in intent does not exist

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
```diff
@@ -103,41 +105,49 @@ public void onPageSelected(int position) {
...
         // Save references to fragments
-        final FragmentManager manager = getSupportFragmentManager();
-        manager.putFragment(outState, Integer.toString(R.id.nav_notebook), _notebook);
-        manager.putFragment(outState, Integer.toString(R.id.nav_quicknote), _quicknote);
-        manager.putFragment(outState, Integer.toString(R.id.nav_todo), _todo);
-        manager.putFragment(outState, Integer.toString(R.id.nav_more), _more);
+        try {
+            final FragmentManager manager = getSupportFragmentManager();
+            // Put and get notebook first. Most important for correct operation.
+            manager.putFragment(outState, Integer.toString(R.id.nav_notebook), _notebook);
```
harshad1, line 128: Catching case when these are null. Happens when we request permissions on resume

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
```diff
@@ -182,10 +192,8 @@ private void optShowRate() {
 
     public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
         super.onRequestPermissionsResult(requestCode, permissions, grantResults);
-        MarkorPermissionChecker permc = new MarkorPermissionChecker(this);
-        permc.checkPermissionResult(requestCode, permissions, grantResults);
-
-        if (_cu.checkExternalStoragePermission(this, false)) {
+        _permc.checkPermissionResult(requestCode, permissions, grantResults);
```
harshad1, line 195: Made the Checker a member

app/src/main/java/net/gsantner/opoc/frontend/base/GsActivityBase.java
```diff
@@ -83,6 +87,21 @@ protected void onResume() {
...
+        // Only do this when not being restored
+        if (_activityFirstTimeVisible.getAndSet(false) && _savedInstanceState == null) {
+            new Handler(getMainLooper()).post(this::onActivityFirstTimeVisible);
+        }
+    }
+
+    /**
+     * This will be called when this activity gets the first time visible
+     */
+    public void onActivityFirstTimeVisible() {
```
harshad1, line 102: Added to activities too.

app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
```diff
@@ -2347,44 +2347,49 @@ public String formatDateTime(@Nullable final Context context, @NonNull final Str
         return formatDateTime(locale, format, datetime, def);
     }
 
-    @SuppressWarnings("ConstantConditions")
-    public boolean checkExternalStoragePermission(final Activity context, final boolean doRequest, String... optionalDescription) {
+    public void requestExternalStoragePermission(final Activity activity, final String description) {
```
harshad1, line 2350: reorganized this code without some of the options

app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
```diff
@@ -2347,44 +2347,49 @@ public String formatDateTime(@Nullable final Context context, @NonNull final Str
...
-                        .setMessage(optionalDescription[0])
-                        .setCancelable(false)
-                        .setPositiveButton(android.R.string.yes, (dialog, which) -> permissionRequest.get().callback())
-                        .setNegativeButton(android.R.string.no, null)
-                        .show();
-                d.setCanceledOnTouchOutside(false);
-            }
-        }
+    @SuppressWarnings("ConstantConditions")
+    public boolean checkExternalStoragePermission(final Activity context) {
```
harshad1, line 2391: The check function is now much simpler and cleaner

gsantner:
So have you started Mordor not Markor and from this build?

app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
```diff
@@ -1564,41 +1569,29 @@ public void draftEmail(final Context context, final String subject, final String
...
-            }
-        };
+        File result = null;
 
         if ((Intent.ACTION_VIEW.equals(action) || Intent.ACTION_EDIT.equals(action)) || Intent.ACTION_SEND.equals(action)) {
+
             // Màrkor, SimpleMobileTools FileManager
             if (receivingIntent.hasExtra((tmps = EXTRA_FILEPATH))) {
-                probeFiles.add(receivingIntent.getStringExtra(tmps));
+                result = checkPath(receivingIntent.getStringExtra(tmps));
```
harshad1, line 1583: Instead of adding all the test paths to a list and then filtering it, we just test each time. The subsequent tests don't happen if a previous test passes. This prevents additional work and cleans up a couple of cases.

app/src/main/res/values/strings.xml
```diff
@@ -206,6 +206,7 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="unordered_list_char_asterisk">Asterisk (*)</string>
     <string name="unordered_list_char_plus">Plus (+)</string>
     <string name="filemanager_doesnot_supply_required_data__appspecific">Markor can only edit local, offline available documents and does not support file managers that do not provide the full file path.</string>
+    <string name="file_does_not_exist">File or directory does not exist</string>
```
harshad1, line 209: Warn that intent files don't exist (most common case is shortcut which has been deleted)

app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
```diff
@@ -173,30 +173,46 @@ protected void onNewIntent(Intent intent) {
...
         }
 
-        if (file != null && (file.isDirectory() || !FormatRegistry.isFileSupported(file))) {
+        // Decide what to do with the file
+        // -----------------------------------------------------------------------
+
+        if (file == null) {
+            final String msg = getString(R.string.filemanager_doesnot_supply_required_data__appspecific) + "\n\n" + getString(R.string.sync_to_local_folder_notice);
+            showErrorMessage(Html.fromHtml(msg.replace("\n", "<br/>")));
+        } else if (file.canRead() || GsFileUtils.canCreate(file)) {
```
harshad1, line 209: We warn on files which can't be read or created. If we ask to open a file which does not exist but can be created we will allow it's creation

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
gsantner: >         _permc = new MarkorPermissionChecker(this);

Problem: This keeps a reference to the activity as a member.   
Suggestion:The better option here is probably similar to ContextUtils, to pass the activity reference on call as argument. Whats not stored somewhere hardly can make a memory leak 😄

app/src/main/java/net/gsantner/opoc/frontend/base/GsActivityBase.java
gsantner: Makes sense, unfortunately hard to make something that works in the same way at fragment and activity, and to share code between. Unfortunately similar but not same 😄 .

app/src/main/java/net/gsantner/opoc/frontend/filebrowser/GsFileBrowserListAdapter.java
```diff
@@ -71,7 +73,7 @@ public class GsFileBrowserListAdapter extends RecyclerView.Adapter<GsFileBrowser
     private final List<File> _adapterDataFiltered; // Filtered list of current folder
     private final Set<File> _currentSelection;
     private File _currentFolder;
-    private final Context _context;
+    private final Activity _activity;
```
gsantner, line 76: While yes it's a Activity refererece, it's OK to store it as `Activity _context` variable.

app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
```diff
@@ -1556,6 +1558,11 @@ public void draftEmail(final Context context, final String subject, final String
         showChooser(context, intent, null);
     }
 
+    private static File checkPath(final String path) {
+        final File f;
+        return (!TextUtils.isEmpty(path) && (f = new File(path)).canRead()) ? f : null;
+    }
```
gsantner, line 1564: If it's required still and it's those ~4 lines, I think it would fit to FileUtils class

harshad1:
@gsantner Please not that I am in the middle of some large changes.

Your feedback is always appreciated :). But it may be a better use of your time to wait a bit.

app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
gsantner: Need to try this with lots of cases and apps. Hope it not breaks

app/src/main/java/net/gsantner/markor/activity/StoragePermissionActivity.java
```diff
@@ -0,0 +1,69 @@
+package net.gsantner.markor.activity;
```
gsantner, line 1: Is that activity used somewhere? I checked the PR and didn't see any startActivity or alike for this one. It's from a previous fix experiment right

gsantner:
No worries, just looked over the merge request as I had time

harshad1:
This PR now includes a somewhat complicated way of dynamically asking for file permissions when required. i.e. if you navigate to a folder where `file.canWrite()` is false, markor will ask for file permissions. This is however brittle and not super maintainable.

I will be simplifying to the following behaviors:

1. Prevent navigation in file browser with a warning
2. Kill document activity with warning if requested file does not have permissions
3. On MainActivity start, ask for permission if notebook, quicknote or todo require it. If permission not given, revert to default locations with warning.

harshad1:
After even more experimentation and testing, I think that the more complex / correct way of doing this may better.

gsantner:
Hello @harshad1 ,  
Should I take a look, or not ready yet? (not sure from last comment)

harshad1:
Pretty close to ready. I think you can go ahead and review.

Feb 11, 2023 04:26:56 Gregor Santner ***@***.***>:

harshad1:
I am much happier with the other PR #1981

Issue #1982, opened 2023-02-18
First Start after Install: missing file write permission: App does not start.


k3b:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**. Not a [question or feature request](https://github.com/gsantner/markor/issues/new/choose).
- [X] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [X] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [X] The bug is still present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor settings are not touched. In case the issue is resolved there, you don't need to create a bug report. The change will be part of the next Markor update.)

### Description

The App does not start. Immediate crash. Same issue as #1976 #1964 .

### Steps to reproduce

1. Using debug version Mordor instead of Markor 
2. uninstall Mordor
3. In Filemanager make shure that folder /storage/emulated/0/Documents/mordor/ does not exist (delete or rename)
4. In Androidstudio put a breakpoint at DocumentEditAndViewFragment.isStateBad()
5. install and debug Mordor
6. => isStateBad fails with  testCreateParent /storage/emulated/0/Documents/mordor/todo.txt
7. => isStateBad fails with  testCreateParent /storage/emulated/0/Documents/mordor/QuickNote.md
8. => App crashes with "android.view.WindowLeaked: Activity net.gsantner.markor.activity.MainActivity has leaked window ..."
------

Here is the complete Logcat of the crash

<pre>
2023-02-18 12:55:55.523 22978-22978 net.gsantn...ewFragment net.gsantner.markor_test             I  newInstance(document=/storage/emulated/0/Documents/mordor/todo.txt, lineNumber=-919385553, preview=false
2023-02-18 12:55:55.647 22978-22978 net.gsantn...ewFragment net.gsantner.markor_test             I  isStateBad testCreateParent /storage/emulated/0/Documents/mordor/todo.txt
[...]

Issue #1976, opened 2023-02-10
Cannot create new note


km53:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**. Not a [question or feature request](https://github.com/gsantner/markor/issues/new/choose).
- [X] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [X] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [X] The bug is still present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Mordor**. Don't worry; Markor and Mordor appear as completely separate applications. You can install both side-by-side, and Markor settings are not touched. In case the issue is resolved there, you don't need to create a bug report. The change will be part of the next Markor update.)

### Description

I cannot create a new note.  I need to kill markor to do this.   This is new behaviour, perhaps after one of the new updates.   

### Steps to reproduce

1. Open Markor
2. Click icon to create new note

### Information

Android version: 13
Device: Samsung Note 20 Ultra
App Version: 2.10.7

### Source

F-Droid

### Format / File type

Not specific

### Additional info / Log

```shell
None
```

gsantner:
Thanks for the bug report, but it is not really helpful. It not happens here / not a general issue. But check a newer version of Markor when it's out.

Issue #1964, opened 2023-01-29
I built the app in android studio and it always crash on start.  


ELY3M:
I think I am missing something?       

* App version: Latest commit from this repo - I did git clone from this repo and built it in Android Studio
* Android version/rom: Android 13 
* Device: Samsung S21+  

## Explain the Problem
the app always crash on start.           

## Steps to Reproduce
I built the app in Android Studio.    
Tap on app and it will just crash.  

I may be missing something??????       

## Log
2023-01-29 11:08:11.024 AndroidRuntime          pid-4266                             E  FATAL EXCEPTION: main
                                                                                        Process: net.gsantner.markor, PID: 4266
                                                                                        java.lang.NullPointerException: Attempt to read from field 'androidx.fragment.app.FragmentManager androidx.fragment.app.Fragment.mFragmentManager' on a null object reference in method 'void androidx.fragment.app.FragmentManager.putFragment(android.os.Bundle, java.lang.String, androidx.fragment.app.Fragment)'
                                                                                        	at androidx.fragment.app.FragmentManager.putFragment(FragmentManager.java:936)
                                                                                        	at net.gsantner.markor.activity.MainActivity.onSaveInstanceState(MainActivity.java:121)
                                                                                        	at android.app.Activity.performSaveInstanceState(Activity.java:2265)
                                                                                        	at android.app.Instrumentation.callActivityOnSaveInstanceState(Instrumentation.java:1563)
                                                                                        	at android.app.ActivityThread.callActivityOnSaveInstanceState(ActivityThread.java:6487)
                                                                                        	at android.app.ActivityThread.callActivityOnStop(ActivityThread.java:5862)
                                                                                        	at android.app.ActivityThread.performStopActivityInner(ActivityThread.java:5796)
                                                                                        	at android.app.ActivityThread.handleStopActivity(ActivityThread.java:5893)
                                                                                        	at android.app.servertransaction.StopActivityItem.execute(StopActivityItem.java:43)
                                                                                        	at android.app.servertransaction.ActivityTransactionItem.execute(ActivityTransactionItem.java:45)
                                                                                        	at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:176)
                                                                                        	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:97)
                                                                                        	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2571)
[...]

PR #1983, opened 2023-02-18
#1982: feat: Added additional logging to analyse app crash #1976 #1964

k3b:
The bug is not fixed but there is additional logcat output to find out what is happending.


k3b:
not needed any more because new permission implementation is on it-s way in https://github.com/harshad1/markor/tree/simpler_permissions

Issue #1980, opened 2023-02-17
Add fileformat CSV to view/export as a html table


k3b:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **enhancement/feature request**. Not a [bug or question](https://github.com/gsantner/markor/issues/new/choose).
- [X] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [X] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [X] The wanted feature/enhancement is not present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor's settings are not touched. If your desired feature is present, you don't need to open this issue. The change will be part of the next Markor update.)

### Description

[CSV Files](https://en.wikipedia.org/wiki/Comma-separated_values) are plain-text files that are used to import and export tabular data from other apps. Many spreadsheet applications like MS-Excel or Libre-Office-Calc can read and write csv.

It would be great if Markor can preview/export csv files as html-tables.

---

Implementation details:

on android csv files can use these mimeTypes

* `text/csv` see https://www.rfc-editor.org/rfc/rfc7111
* `text/comma-separated-values` non standard but used by android filemanager

----

There is a java lib that can read and write csv at https://opencsv.sourceforge.net/ (License Apache 2)
Gradle dependency `implementation 'com.opencsv:opencsv:5.7.1'`

I have successfully used the lib in https://github.com/SecUSo/privacy-friendly-finance-manager/pull/52

### Information

Android version: 10
Device: lg
App Version: (F-Droid) 2.10.7

### Source

F-Droid

### Format / File type

Something else (-> Additional info)

### Additional info / Log

```shell
[CSV File](https://en.wikipedia.org/wiki/Comma-separated_values) 

mimeTypes

* `text/csv` see https://www.rfc-editor.org/rfc/rfc7111
* `text/comma-separated-values` non standard but used by android filemanager
```

k3b:
Related : https://github.com/gsantner/markor/discussions/1667

gsantner:
You can open csv with Markor generally as it is currently.

[...]

PR #1986, opened 2023-02-19
Implementation of Csv is formatted/exported as html-Table (#1980)

k3b:


k3b:
closed because these changes are now part of mergerequest #1988

Issue #1987, opened 2023-02-21
Implement Csv SyntaxHighlighter


k3b:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **enhancement/feature request**. Not a [bug or question](https://github.com/gsantner/markor/issues/new/choose).
- [X] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [X] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [X] The wanted feature/enhancement is not present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor's settings are not touched. If your desired feature is present, you don't need to open this issue. The change will be part of the next Markor update.)

### Description

This is a follow up to #1980 

_As a (role)_ user that wants to edit csv on android
_I want (feature)_ each column should be shown in a different unique color (SyntaxHighlighter)
_so that (benefit)_ it is easy to find out which csv-data belongs to which colum/header

![grafik](https://user-images.githubusercontent.com/1374583/220289635-66dc3ffb-9894-4bf1-84e1-4dad9c666936.png)
(Example taken from the Androidstudio Plugin https://github.com/SeeSharpSoft/intellij-csv-validator )

### Information

Android version: 10
Device: LG
App Version:  2.10.7 (143)

### Source

F-Droid

### Format / File type

Something else (-> Additional info)

### Additional info / Log

Markor uses regular expressions to implement syntax highlighting.

The CSV implementation uses a java matchers that also infers currentls used `csv-field-delimiter-char` and `quoting-char`

k3b:
Current status Markor support for csv-files (as of 2023-03-15):

* Working prototype in [k3b-s branch '1987-Implement_Csv_SyntaxHighlighter'](https://github.com/k3b/markor/tree/1987-Implement_Csv_SyntaxHighlighter) that also contains implementation for issue #1980 
* There is a pending mergerequest #1988 for adding the changes to the next Markor release

Issue #1978, opened 2023-02-11
Support the task check of the number list


ina6ra:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **enhancement/feature request**. Not a [bug or question](https://github.com/gsantner/markor/issues/new/choose).
- [X] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [X] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [X] The wanted feature/enhancement is not present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor's settings are not touched. If your desired feature is present, you don't need to open this issue. The change will be part of the next Markor update.)

### Description

I want you to support the task check of the number list.

```
(Now)

1. [ ] one           1. [ ] one
2. [ ] two      ->   -[ ] [ ] two
3. [ ] three         -[ ] [ ] three
```

```
(Support)

1. [ ] one           1. [ ] one
2. [ ] two      ->   2. [x] two
3. [ ] three         3. [x] three
```

### Information

Android version: 10
Device: Rakuten Hand (P710)
App Version: v2.10.7(144)

### Source

Google Play

### Format / File type

Markdown

### Additional info / Log

```shell
-
```

gsantner:
Is this syntax generally support at Markdown, and does this syntax work at other applications? (see <https://spec.commonmark.org/0.30/>)

ina6ra:
Supported by Obsidian.
Markor and Github will display unnumbered lists.

In the previous version, it was as follows, so I thought it would be supported.

```
1. [ ] one           1. [ ] one
2. [ ] two      ->   2. x[ ] two
3. [ ] three         3. x[ ] three
```

gsantner:
>  Markor and Github will display unnumbered lists.

OK yes, it's not something part of CommonMark or GFM (which Markor as well GitHub supports). I've checked the Markdown parser Markor uses ([flexmark](https://github.com/vsch/flexmark-java/issues)) and it's not something I can find.

[...]

Issue #1329, opened 2021-05-03
Android 4.4 crash when moving files


whunger:
#### General information

* **App version: 2.6.0** 
* **System: Android 4.4.4** 

#### Description

When moving a file from one folder to another, the file is copied, then Markor crashes with the following exception. The file has been copied successfully to the target location, but it has not been removed from the source folder.

#### Log

```
05-03 11:45:44.972 E/dalvikvm(12677): Could not find class 'java.util.function.Consumer', referenced from method org.apache.commons.io.IOUtils.closeQuietly
05-03 11:45:44.972 W/dalvikvm(12677): VFY: unable to resolve check-cast 4643 (Ljava/util/function/Consumer;) in Lorg/apache/commons/io/IOUtils;
05-03 11:45:44.972 D/dalvikvm(12677): VFY: replacing opcode 0x1f at 0x0001
05-03 11:45:44.972 W/dalvikvm(12677): VFY: unable to find class referenced in signature (Ljava/util/function/Consumer;)
05-03 11:45:44.972 I/dalvikvm(12677): Could not find method java.util.function.Consumer.accept, referenced from method org.apache.commons.io.IOUtils.closeQuietly
05-03 11:45:44.982 W/dalvikvm(12677): VFY: unable to resolve interface method 35121: Ljava/util/function/Consumer;.accept (Ljava/lang/Object;)V
05-03 11:45:44.982 D/dalvikvm(12677): VFY: replacing opcode 0x72 at 0x0009
05-03 11:45:45.012 D/AndroidRuntime(12677): Shutting down VM
05-03 11:45:45.012 W/dalvikvm(12677): threadid=1: thread exiting with uncaught exception (group=0x41d63ba8)
05-03 11:45:45.052 E/AndroidRuntime(12677): FATAL EXCEPTION: main
05-03 11:45:45.052 E/AndroidRuntime(12677): Process: net.gsantner.markor, PID: 12677
05-03 11:45:45.052 E/AndroidRuntime(12677): java.lang.NoClassDefFoundError: java.util.function.Consumer
05-03 11:45:45.052 E/AndroidRuntime(12677): 	at org.apache.commons.io.IOUtils.closeQuietly(IOUtils.java:345)
05-03 11:45:45.052 E/AndroidRuntime(12677): 	at org.apache.commons.io.IOUtils.closeQuietly(IOUtils.java:498)
[...]

PR #1988, opened 2023-02-22
Add new format: CSV -  Preview text converter and Editor syntax highlighting #1987

k3b:
See also: #1978 #1980 #1667


app/src/main/java/net/gsantner/markor/util/StringUtils.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.apache.commons.lang3.ArrayUtils;
+
+public class StringUtils extends org.apache.commons.lang3.StringUtils {
+    // modified version of org.apache.commons.lang3.StringUtils#indexOfAny
+    /**
+     * Same as {@link org.apache.commons.lang3.StringUtils#indexOfAny(CharSequence, char...)
+     * where you can specify the search intervall}
+     */
+    public static int indexOfAny(final CharSequence cs, int csFirst, int csLen, final char... searchChars) {
```
harshad1, line 11: We have [GsTextUtils](https://github.com/gsantner/markor/blob/master/app/src/main/java/net/gsantner/opoc/format/GsTextUtils.java) and [TextViewUtils](https://github.com/gsantner/markor/blob/master/app/src/main/java/net/gsantner/markor/frontend/textview/TextViewUtils.java)

These should go in one of them.

app/src/main/java/net/gsantner/markor/util/StringUtils.java
```diff
@@ -0,0 +1,67 @@
...
+                }
+            }
+        }
+        return INDEX_NOT_FOUND;
+    }
+
+    /**
+     * returns search for begin of line starting for csFirst down to 0
+     */
+    public static int beginOfLine(final CharSequence cs, int csFirst) {
```
harshad1, line 39: https://github.com/gsantner/markor/blob/master/app/src/main/java/net/gsantner/markor/frontend/textview/TextViewUtils.java#L67

We already have this

app/src/main/java/net/gsantner/markor/format/FormatRegistry.java
```diff
@@ -101,6 +106,16 @@ public static FormatRegistry getFormat(int formatId, @NonNull final Context cont
         final AppSettings appSettings = ApplicationObject.settings();
 
         switch (formatId) {
+            case FORMAT_CSV: {
```
harshad1, line 109: General comment: We really need to spend some time making each format cleaner and more independent. One has to touch like 10 files to add a format and this can lead to bugs + reduced maintainability. Ideally we should just have to implement the classes and 'register' them in one place

k3b:
@harshad1 thanks for taking the time to review my code. I will add comments to any issue you found

app/src/main/java/net/gsantner/markor/util/StringUtils.java
k3b: I was not aware that these exist. Moved indexOfAny from [net.gsantner.markor.util.StringUtils] to [net.gsantner.opoc.format.GsTextUtils]

app/src/main/java/net/gsantner/markor/util/StringUtils.java
k3b: Moved from [net.gsantner.markor.util.StringUtils] to [net.gsantner.opoc.format.GsTextUtils]. Replaced implementation with getNeighbourLineEndings() . Created unittests for getNeighbourLineEndings() to find out what it exactly does and updated docs

app/src/main/java/net/gsantner/markor/format/FormatRegistry.java
k3b: I agree. it was painfull for me to find all places i have to fix to make csv work. Probably i missed some.

this should be done once your "make permissions simpler" is finished.

app/src/main/java/net/gsantner/markor/format/csv/Csv2MdTable.java
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+
+    private static void addColumnsLine(StringBuilder mdMarkup, String[] colums, int headerLength) {
+        for (int i = 0; i < max(headerLength, colums.length); i++) {
+            addColumnContainingNL(mdMarkup.append(MD_COL_DELIMITER), getCol(colums, i));
+        }
+        mdMarkup
+                .append(MD_COL_DELIMITER)
+                .append(MD_LINE_DELIMITER);
```
gsantner, line 109:
```suggestion
        mdMarkup.append(MD_COL_DELIMITER).append(MD_LINE_DELIMITER);
```

app/src/main/java/net/gsantner/markor/format/csv/Csv2MdTable.java
```diff
@@ -0,0 +1,160 @@
...
+import java.io.Reader;
+import java.io.StringReader;
+
+/**
+ * Simple csv to md-table converter using OpenCsv implementation
+ * https://opencsv.sourceforge.net/ Licensed under Appache2
+ * <p>
+ * This file should not have dependencies to Android or Markor-Architecture.
+ */
+public class Csv2MdTable implements Closeable {
```
gsantner, line 30: Suggestion: Combine Csv2Mdable, CsvConfig, CsvMatcher to a single class optionally with inner classes. Don't see the need to scatter it

app/src/main/java/net/gsantner/markor/format/csv/CsvConfig.java
```diff
@@ -0,0 +1,50 @@
...
+public class CsvConfig {
+    public static final CsvConfig DEFAULT = new CsvConfig(';', '"');
+    public static final char[] CSV_DELIMITER_CANDIDATES = {DEFAULT.getFieldDelimiterChar(), ',', '\t', ':', '|'};
+    public static final char[] CSV_QUOTE_CANDIDATES = {DEFAULT.getQuoteChar(), '\''};
+    private final char fieldDelimiterChar;
+    private final char quoteChar;
+
+    public CsvConfig(char fieldDelimiterChar, char quoteChar) {
+        this.fieldDelimiterChar = fieldDelimiterChar;
+        this.quoteChar = quoteChar;
```
gsantner, line 27: Use `m_` or `_` instead of `this.`. (Leaving this comment only once here, but applies to all new files)

app/src/main/java/net/gsantner/markor/format/csv/CsvSyntaxHighlighter.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+
+    protected void createSpanForColumn(int from, int to, int color, int colNumner) {
+        if (DEBUG_COLORING) {
+            Log.d(TAG, "#" + colNumner +
+                    "(" + from +
+                    "," + to +
+                    "," + color +
+                    ") = " + _spannable.subSequence(from, to));
```
gsantner, line 66: String.format("#%...."

app/src/main/java/net/gsantner/markor/format/csv/README.md
```diff
@@ -0,0 +1,32 @@
+.../app/src/main/java/net/gsantner/markor/format/csv/README.md
```
gsantner, line 1: Please don't put non-code files into java code directory. If you want to make code comments/code documentatoin, leave it in code (javadoc), for markdown files, theres the `<gitrepo>/doc/` folder.

app/src/main/res/values/strings.xml
```diff
@@ -464,4 +464,7 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="document_error_clip">Error encountered: Text copied to clipboard.</string>
     <string name="wont_save_min_length">Files with fewer than %d characters not saved automatically in order to prevent data loss.</string>
     <string name="advanced_filtering_help"><![CDATA[<small><a href="https://www.github.com/gsantner/markor/discussions/1940">Advanced filtering syntax</a></small>]]></string>
+
+    <string name="csvtext">csv</string>
```
gsantner, line 468: can be in not_translatable

app/build.gradle
```diff
@@ -124,6 +124,10 @@ dependencies {
     implementation "com.vladsch.flexmark:flexmark-ext-typographic:${version_library_flexmark}"
     implementation "com.vladsch.flexmark:flexmark-ext-admonition:${version_library_flexmark}"
 
+    // csv support
+    // https://opencsv.sourceforge.net/licenses.html License: Appache2
```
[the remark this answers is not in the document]
k3b: fixed in build.gradle and Csv2MdTable.java

app/src/main/java/net/gsantner/markor/format/csv/Csv2MdTable.java
k3b: done

app/src/main/java/net/gsantner/markor/format/csv/CsvConfig.java
k3b: done

app/src/main/java/net/gsantner/markor/format/csv/README.md
k3b: moved to .../doc/CSV_README.md

app/src/main/res/values/strings.xml
k3b: done

app/src/main/java/net/gsantner/markor/format/csv/Csv2MdTable.java
k3b: I would prefer to not merge Csv2MdTable with CsvMatcher because of separation of concerns. (The same way as XxxSyntaxHighlighter and XxxTextConverter are seperated classes)

* CsvMatcher  is an implementation detail for CsvSyntaxHighlighter
* Csv2MdTable is an implementation detail for CsvTextConverter

app/src/main/java/net/gsantner/markor/format/csv/CsvSyntaxHighlighter.java
k3b: done

app/src/main/java/net/gsantner/markor/format/csv/Csv2MdTable.java
gsantner: Then how about CsvMatcher --> to CsvSyntaxHighlighter and  
Csv2MdTable --> to CsvTextConverter  

?

gsantner:
@k3b 

Comparing the dependency tree shows that there is more added by this PR implicitly, not just opencsv. So these projects should be mentioned as well regarding license.

grafik


I also checked the OpenCSV library regarding Android support - as Markor support quite old versions of Android too. It seems like this library has minimum of Java 8 and there is something broken regarding beans. Do you know something about this by chance? <https://sourceforge.net/p/opencsv/wiki/FAQ/>

Otherwise, it seems to me like there is no deeper connection to OpenCSV in the code of this merge request. Only at the textconverter for html conversion.  
Does a other CSV converter come to your mind that may have less dependencies and/or is maybe better suited to legacy devices?

gsantner:
Markor has the minimum device support with API16 = Android 4.1 Jelly Bean.  
You can find out the corresponding Java version that is available on a device correspdonding to OS version here <https://source.android.com/docs/setup/build/older-versions?hl=de#jdk>.

So this means at the end of the day, Markor will be running on Java 6 at runtime on oldest devices.

However there is quite some magic behind the scenes going on at build time so the Java 7/8 code is compiled into a fashion it still works on older Java 6 devices. Unfortunatley it of course doesn't work for some things that is just not available on Java 6, and can't be compiled into a fashion it works.   
You can find one example of how it appears to happen at this Markor bug #1329 . So as you can see verything links and starts fine, but at some point of runtime the app crashes on older device then because it doesn't have some builtin in java classes/methods for example.  
So thats why I asked to take this into consideration.

As of using proguard to remove some dependency - it doesn't much, then it crashes somewhere else because can't find the  classes of that dependency :-D. You don't need to look into it.


If you search the web a bit of available java CSV libraries I would not be surprised that you get a few results. And it might happens that there is one that works fine on older Java versions and maybe even has less dependencies.  
When there is nothing other that would be an option, we can try continue with OpenCSV.  
At the end it's adding the library, maybe a few lines of parser configuration, and the conversion - guess whatever library it is, it won't be much of a differnce at amount of changes.

k3b:
I have replaced opencsv 5.7.1 (current version) with opencsv 3.10 (last java-6 version) so we should be on the safe side for android-4.1. 

```
+--- com.vladsch.flexmark:flexmark-ext-admonition:0.42.14
|    +--- com.vladsch.flexmark:flexmark-util:0.42.14
|    +--- com.vladsch.flexmark:flexmark:0.42.14 (*)
|    \--- com.vladsch.flexmark:flexmark-formatter:0.42.14 (*)
+--- com.opencsv:opencsv:3.10
|    +--- org.apache.commons:commons-lang3:3.5
|    \--- commons-beanutils:commons-beanutils:1.9.3
|         +--- commons-logging:commons-logging:1.2
|         \--- commons-collections:commons-collections:3.2.2
+--- com.pixplicity.generate:library:1.1.8

```

k3b:
i just removed the beans code and the csv code still works as expected

```
+--- com.opencsv:opencsv:3.10
|    \--- org.apache.commons:commons-lang3:3.5
+--- com.pixplicity.generate:library:1.1.8

```

app/src/main/java/net/gsantner/markor/format/csv/CsvTextConverter.java
```diff
@@ -0,0 +1,186 @@
...
+
+        private static void addColumnsLine(StringBuilder mdMarkup, String[] columns, int headerLength) {
+            for (int i = 0; i < max(headerLength, columns.length); i++) {
+                addColumnContainingNL(mdMarkup.append(MD_COL_DELIMITER), getCol(columns, i));
+            }
+            mdMarkup.append(MD_COL_DELIMITER).append(MD_LINE_DELIMITER);
+        }
+
+        private static String getCol(String[] columns, int i) {
+            if (i >= 0 && i < columns.length) return columns[i];
```
gsantner, line 135: reutrn condition ? columnsi : "";

gsantner, line 135: Please use i.e. `return condition ? a : b`. Inline if with a action afterwards is prone to writing bad code

app/src/main/java/net/gsantner/markor/format/csv/CsvTextConverter.java
```diff
@@ -0,0 +1,186 @@
...
+            do {
+                m_lineNumber++;
+                columns = m_csvReader.readNext();
+            } while (columns != null && isComment(columns));
+            return columns;
+        }
+
+        private boolean isComment(String[] columns) {
+            // empty line without content
+            if (columns.length == 1 && columns[0].trim().length() == 0) return true;
```
gsantner, line 174: Please use

```
if (condition) {
  code
}
```

(everywhere where this can be found in the merge request)

app/src/main/java/net/gsantner/markor/format/csv/CsvTextConverter.java
```diff
@@ -0,0 +1,186 @@
...
+            StringBuilder mdMarkup = new StringBuilder();
+            try (BufferedReader bufferedReader = new BufferedReader(csvMarkup, BUFFER_SIZE)) {
+                CsvConfig csvConfig = inferCsvConfiguration(bufferedReader);
+                try (Csv2MdTable toMdTable = new Csv2MdTable(csvConfig, bufferedReader)) {
+                    String[] headers = toMdTable.readNextCsvColumnLine();
+
+                    if (headers != null && headers.length > 0) {
+                        addColumnsLine(mdMarkup, headers, headers.length);
+
+                        for (int i = 0; i < headers.length; i++) {
```
gsantner, line 89: foreach can be used here

for (final String h : headers) {

app/src/main/java/net/gsantner/markor/format/csv/CsvTextConverter.java
```diff
@@ -0,0 +1,186 @@
...
+            } while (columns != null && isComment(columns));
+            return columns;
+        }
+
+        private boolean isComment(String[] columns) {
+            // empty line without content
+            if (columns.length == 1 && columns[0].trim().length() == 0) return true;
+
+            // comments start with "#" char
+            return columns[0].startsWith("#");
```
gsantner, line 177: When columns is a empty array this crashes.

Is it possible to get a empty array here?

app/src/main/java/net/gsantner/markor/format/csv/CsvSyntaxHighlighter.java
```diff
@@ -0,0 +1,186 @@
...
+            }
+            from = to + 1;
+        }
+    }
+
+    private void createSpanForColumn(int from, int to, int color, int colNumner) {
+        if (DEBUG_COLORING) {
+            Log.d(TAG, String.format("#%d(%d,%d,%d) = %s", colNumner, from, to, color, _spannable.subSequence(from, to)));
+        }
+        if (colNumner >= 0 && from > 0 && Math.abs(to - from) >= 0) {
```
gsantner, line 76: typo Numner

app/src/main/java/net/gsantner/markor/format/csv/CsvTextConverter.java
k3b: done

app/src/main/java/net/gsantner/markor/format/csv/CsvTextConverter.java
k3b: done in CsvConfig, CsvSyntaxHighlighter, CsvTextConverter

app/src/main/java/net/gsantner/markor/format/csv/CsvTextConverter.java
k3b: done

app/src/main/java/net/gsantner/markor/format/csv/CsvTextConverter.java
k3b: i dont know if this is possibile. for safety reasons i added  

           if (columns.length == 0) {
                return true;
            }

done

app/src/main/java/net/gsantner/markor/format/csv/CsvSyntaxHighlighter.java
k3b: fixed

gsantner (approved):
Thank you for your contribution! I tried it on the device and it worked quite nice. I pushed some minor improvements too.

PR #1995, opened 2023-03-12
Search in both filename and contents

harshad1:
In this PR I have added functionality for searching both contents _and_ title if so desired.

Also included are a few tweaks which should improve search performance.


app/src/main/java/net/gsantner/markor/frontend/filesearch/FileSearchEngine.java
```diff
@@ -129,13 +136,13 @@ public QueueSearchFilesTask(final SearchOptions config, final GsCallback.a1<List
                     }
                 }
             }
-            _regex = pattern;
+            _matcher = pattern != null ? pattern.matcher("") : null;
```
harshad1, line 139: resetting a matcher is slightly faster than creating one

app/src/main/java/net/gsantner/markor/frontend/filesearch/FileSearchEngine.java
```diff
@@ -198,49 +205,37 @@ protected List<FitFile> doInBackground(Void... voidp) {
...
+            final Queue<File> subQueue = new LinkedList<>();
 
             try {
                 if (!currentDir.canRead() || currentDir.isFile()) {
                     return subQueue;
                 }
 
-                File[] subDirsOrFiles = currentDir.listFiles();
+                final File[] subDirsOrFiles = currentDir.listFiles();
+                final int trimSize = _config.rootSearchDir.getCanonicalPath().length() + 1;
```
harshad1, line 216: Just trim the first n characters rather than find replace

harshad1:
/home/harshad/Downloads/Screenshot_20230311-222704.png

Search options

harshad1:
Results
image

harshad1:
@adelobosko Any feedback you have here would be nice

adelobosko:
> @adelobosko Any feedback you have here would be nice

Seems all ok. I like this PR.  It is much better.

gsantner:
To be honest I would suggest to make it rather simple: always scan the filename at search too

adelobosko:
Only i think need to rename the variable "count" in the commit [2cd2279](https://github.com/gsantner/markor/pull/1995/commits/2cd2279cb194e3b75c7b3f2d6f27e68848666e0a)

harshad1:
@gsantner Making this so we don't need an option and always search filename too.

gsantner:
Of course if you want to, just a suggestion. If it's preffered to be a choice you can keep it^^

harshad1:
I actually can't think of any scenario where I want to search content but _not_ filename. I would rather make a simpler PR and add functionality later if I feel the need.

app/src/main/java/net/gsantner/markor/frontend/filesearch/FileSearchEngine.java
```diff
@@ -97,7 +97,9 @@ public static FileSearchEngine.QueueSearchFilesTask queueFileSearch(Activity act
     public static class QueueSearchFilesTask extends AsyncTask<Void, Integer, List<FitFile>> {
         private final SearchOptions _config;
         private final GsCallback.a1<List<FitFile>> _callback;
-        private final Pattern _regex;
+
+        // _matcher.reset() is _not_ thread safe. Will need alternate approach when we make search parallel
+        private final Matcher _matcher;
```
gsantner, line 102: Does adding a synchronized(object) around the two reset calls solve the problem?

gsantner (dismissed):
Looks OK, do you still work on it or is it ready from your side?

About the matcher synchronized, it's just a comment and feel free to check it out, but it doesn't hinder merging.

app/src/main/java/net/gsantner/markor/frontend/filesearch/FileSearchEngine.java
harshad1: There is no issue right now as this code is not parallel. In the future I think searching needs to be made faster and I just left this as a note for myself or another dev in the future.

harshad1:
I pushed up a change to also search folder names. Clicking on these opens the folder in the main activity.

@adelobosko I also pushed up some cleanups. Please take a look at the changes again. Any comments will be valuable.

app/src/main/java/net/gsantner/markor/frontend/filesearch/FileSearchResultSelectorDialog.java
```diff
@@ -287,21 +295,4 @@ public void registerDataSetObserver(DataSetObserver dataSetObserver) {
         public void unregisterDataSetObserver(DataSetObserver dataSetObserver) {
         }
     }
-
-    public static class GroupItemsInfo {
```
harshad1, line 291: This class is nearly identical to FitFile. Just using FitFile in both places

app/src/main/java/net/gsantner/opoc/util/GsFileUtils.java
```diff
@@ -61,8 +66,8 @@ public class GsFileUtils {
     public final static SimpleDateFormat DATEFORMAT_IMG = new SimpleDateFormat("yyyyMMdd-HHmmss", INITIAL_LOCALE); //20190511-230845
 
     // Used on methods like copyFile(src, dst)
-    private static final int BUFFER_SIZE = 4096;
-    private static final GsHashMap<String, String> MIME_TYPE_CACHE = new GsHashMap<>();
```
harshad1, line 65: This is just a path -> mime cache. IMO many thousands of entries will not affect performance. I think we can just use a ConcurrentHashMap and we will gain some speed and safety

app/src/main/java/net/gsantner/opoc/util/GsFileUtils.java
```diff
@@ -420,7 +425,7 @@ public static String relativePath(File src, File dest) {
     }
 
     @SuppressWarnings("StatementWithEmptyBody")
-    private final static GsCallback.s1<File> gatherMimeType = file -> {
```
harshad1, line 423: Made this a regular function

app/src/main/java/net/gsantner/markor/frontend/filesearch/FileSearchResultSelectorDialog.java
```diff
@@ -217,11 +226,15 @@ public View getGroupView(final int groupPosition, final boolean isExpanded, fina
...
-            final int iconResId = groupInfo.isDirectory || groupInfo.children.isEmpty() ? 0 : isExpanded
-                    ? R.drawable.ic_baseline_keyboard_arrow_up_24
-                    : R.drawable.ic_baseline_keyboard_arrow_down_24;
+            final int iconResId;
+            if (groupInfo.isDirectory || groupInfo.children.isEmpty()) {
+                iconResId = 0;
+            } else if (isExpanded) {
+                iconResId = R.drawable.ic_baseline_keyboard_arrow_up_24;
+            } else {
+                iconResId = R.drawable.ic_baseline_keyboard_arrow_down_24;
```
harshad1, line 235: Right now we don't specify an icon for file name results. Would it be better to use a folder or file icon instead?

I am adding a trailing slash to folders in order to indicate that they are folders

gsantner: Usability improvements are sure not wrong. There is also the bug with the bottom snackbar not dismissing, or continue searching wrongly.

gsantner (approved):
Thanks, works good

👍🏻

PR #2022, opened 2023-04-15
Support AsciiDoc View-Mode by TextConverter with asciidoctor.js

TimReset:
Hi.

I added supporting asciidoctor.js to view adoc file.

I added 2 files:
asciidoctor.css
asciidoctor.min.js


gsantner:
@TimReset  
Can you please add a license notice for the new added files? There is a licenses_3rd_party file which contains such licenses and is also possible to lookup while Markor is running.

TimReset:
@gsantner 
done

gsantner:
Note: Exporting more or less is broken with asciidoc, due how it's currently implemented

gsantner (approved):
Notes:

* Exporting more or less is broken with asciidoc, due how it's currently implemented
* The asciidoc sample document is mostly broken / doesn't contain much asciidoc syntax elements. Guess the images needs to be fixed so it works properly. But for missing "good" asciidoc sample, it's not in scope of the PR

PR #2025, opened 2023-04-21
Keep keyboard state onResume

harshad1:
Currently, the soft keyboard will close when one navigates away from and back to a markor document.

After this PR, the keyboard state will be remembered (by the android system).

The way keyboard state is remembered after a dialog closes is changed to handle the new behavior

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -65,7 +65,7 @@
             android:label="@string/app_name"
             android:launchMode="singleTop"
             android:taskAffinity=".activity.MainActivity"
-            android:windowSoftInputMode="stateAlwaysHidden|adjustResize">
+            android:windowSoftInputMode="stateUnchanged|adjustResize">
             <intent-filter>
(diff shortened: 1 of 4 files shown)


gsantner:
Do you notice some dialog which leave the keyboard open/does not close it when leaving the activity/page/dialog? Like from MainActivity to DocumentActivity

harshad1:
It used to be that if the keyboard was opened in a dialog it would stay open when the dialog closed (for example the search dialog). I fixed this some time ago so the keyboard would close properly (in most cases - each dialog needs a close callback to restore the state correctly).

A side effect of my fix was that the keyboard would close when one navigated away from and back to a document. So if you were editing a file and switched to another app, the keyboard would be gone when you switched back. This is very annoying when editing a file and referring to a webpage or something.

In this PR I have tweaked the behavior such that the keyboard is closed when dialogs close (and remains open if it was open before the dialog was opened) but the keyboard state is also preserved when switching navigating away from / to an open document in Android.

It's a small quality of life improvement.

gsantner:
Thanks looks good 😄

Issue #2026, opened 2023-04-22
Sorting: folders and files are sorted separately / "folder first"-setting has no effect


min7-i:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**. Not a [question or feature request](https://github.com/gsantner/markor/issues/new/choose).
- [X] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [X] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [X] The bug is still present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor settings are not touched. In case the issue is resolved there, you don't need to create a bug report. The change will be part of the next Markor update.)

### Description

Probably due to the changed sorting logic in the last update I noticed that folders and files are now sorted separately. As a result, the "folder first"-setting has no effect anymore.

Personally, I don't mind that folders and files are sorted separately. However, I'd still like to have the option to show folders at the top while sorting the files by date (latest modification date at the top below the folders). In the latest version this is not possible anymore.

Here is a comparison of the old and new sorting logic:

| old | new |
| --- | --- |
| ![Screenshot_20230421-214059 (Benutzerdefiniert)](https://user-images.githubusercontent.com/97251923/233778460-1fdef07a-20e2-4790-85c9-76e79f4ccdf5.png) | ![Screenshot_20230421-210055](https://user-images.githubusercontent.com/97251923/233778475-78b676d7-a0ac-4a59-a516-1f6d527699df.png) |
| ![Screenshot_20230421-214105 (Benutzerdefiniert)](https://user-images.githubusercontent.com/97251923/233778490-26a4eb88-bd1c-42d1-ac5e-4e35bcd5c51f.png) | ![Screenshot_20230421-210102](https://user-images.githubusercontent.com/97251923/233778500-b31091ab-e3cf-41be-94b1-6780d6f49b35.png) |
[...]

PR #2029, opened 2023-04-25
Folders first even when reversed

harshad1:
Another fix for sorting. This ought to be tested more before merge :) #2026


harshad1:
@gsantner I think this is ready to be merged now

gsantner:
Thanks looks good

PR #2030, opened 2023-04-25
ShareInto: Always show button to add to ToDo

harshad1:
In this PR I have removed the limitations for share into todo (< 300 chars, single line).

Instead of restricting newlines, we simply replace all newlines with spaces if the user selects to share text into a file with todo format.


gsantner:
Looked OK, it still applies the separator to markdown file, while I get the date prefix at todo.txt.

gsantner (approved):
Looked OK, it still applies the separator to markdown file, while I get the date prefix at todo.txt.

PR #2043, opened 2023-05-18
Fix NPE on devices that don't support widgets

IlyaBizyaev:
Turns out, `AppWidgetManager.getInstance(context)` can return null:

* https://github.com/chromium/chromium/blob/08033c4b62f48e46e66f97be3a39c99ab9163402/chrome/android/java/src/org/chromium/chrome/browser/quickactionsearchwidget/QuickActionSearchWidgetProvider.java#L289
* https://github.com/tasks/tasks/blob/d25f859385e8301ae274835297f151125f7cefb7/app/src/main/java/org/tasks/widget/AppWidgetManager.kt#L18

I've tested that this change fixes Markor crash on the device.

### DIFF
--- a/app/thirdparty/java/other/writeily/widget/WrMarkorWidgetProvider.java
+++ b/app/thirdparty/java/other/writeily/widget/WrMarkorWidgetProvider.java
@@ -108,6 +108,10 @@ public class WrMarkorWidgetProvider extends AppWidgetProvider {
     public static void updateLauncherWidgets() {
         final Context context = ApplicationObject.get().getApplicationContext();
         final AppWidgetManager appWidgetManager = AppWidgetManager.getInstance(context);
+        if (appWidgetManager == null) {
+            // The device does not support widgets.
+            return;
+        }
         final ComponentName comp = new ComponentName(context, WrMarkorWidgetProvider.class);
         final int[] appWidgetIds = appWidgetManager.getAppWidgetIds(comp);
 


app/thirdparty/java/other/writeily/widget/WrMarkorWidgetProvider.java
```diff
@@ -108,6 +108,10 @@ public void onUpdate(Context context, AppWidgetManager appWidgetManager, int[] a
     public static void updateLauncherWidgets() {
         final Context context = ApplicationObject.get().getApplicationContext();
         final AppWidgetManager appWidgetManager = AppWidgetManager.getInstance(context);
+        if (appWidgetManager == null) {
+            // The device does not support widgets.
```
gsantner, line 112: Is it possible to show some Toast message or alike in case this happens?

gsantner:
Thanks.

Is it possible to show some Toast message or alike in case this happens? 

And how even can users pick widgets if it's not supported? Isn't it a bug of the launcher in this case? Allowing to pick but not supporting any sounds weird

gsantner:
Thanks. So do I understand right: This is about crashing the app, without even having placed any widgets?

IlyaBizyaev:
In my case, it's this device: https://supernote.com/products/supernote

The crash was happening during application startup, I never added any widgets for it.

I guess it's an Android platform feature, depending on build-time configuration of system components, so there's not even a launcher as such to pick widgets for.

gsantner:
OK I see..some E-Ink reading device.

Thanks for the fix and great that Markor works then for you.

Generally said, how good does it work there, is everything visible and clickable that should be? I never got hands on any such device. Would you mind to make a short video or using the app on it?Just intersting how it looks like there :D.

IlyaBizyaev:
Yes, that's the one! It has different menu items depending on the context.

gsantner:
OK, if you want to make further improvements you are welcome. I want Markor to be accessible to everybody for whats there.

PR #2048, opened 2023-05-21
Edit mode TOC Heading level filtering

harshad1:
In this PR I implement heading filtering for markdown and Wikitext. 

An option in the headings dialog lets us filter which heading levels we want shown.

I have implemented and tested this with Markdown and Implemented this with Wikitext.

I need help testing wikitext and implementing + testing asciidoc.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/wikitext/WikitextActionButtons.java
+++ b/app/src/main/java/net/gsantner/markor/format/wikitext/WikitextActionButtons.java
@@ -289,10 +289,10 @@ public class WikitextActionButtons extends ActionButtonBase {
 
     @Override
     public boolean runTitleClick() {
-        final Matcher m = WikitextSyntaxHighlighter.HEADING.matcher(_hlEditor.getText());
-        MarkorDialogFactory.showHeadlineDialog(getActivity(),  _hlEditor, _disabledHeadings, (start, end) -> {
-            if (m.region(start, end).find(start)) {
-                return m.end(2) - m.start(2);
+        final Matcher m = WikitextSyntaxHighlighter.HEADING.matcher("");
+        MarkorDialogFactory.showHeadlineDialog(getActivity(),  _hlEditor, _disabledHeadings, (text, start, end) -> {
+            if (m.reset(text.subSequence(start, end)).find()) {
+                return 7 - (m.end(2) - m.start(2));
             }
             return -1;
         });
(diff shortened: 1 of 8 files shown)


harshad1:
image

app/src/main/java/net/gsantner/opoc/frontend/GsSearchOrCustomTextDialog.java
```diff
@@ -116,35 +118,48 @@ public static class DialogOptions {
         public int clearInputIcon = android.R.drawable.ic_input_delete;
     }
 
-    private static class Adapter extends ArrayAdapter<Integer> {
+    public static class Adapter extends BaseAdapter {
```
harshad1, line 121: Cleaned up / simplified the adapter system significantly.

app/src/main/java/net/gsantner/markor/format/wikitext/WikitextActionButtons.java
```diff
@@ -285,7 +289,13 @@ public static String createWikitextHeaderAndTitleContents(String fileNameWithout
 
     @Override
     public boolean runTitleClick() {
-        MarkorDialogFactory.showHeadlineDialog(WikitextSyntaxHighlighter.HEADING.toString(), getActivity(), _hlEditor);
+        final Matcher m = WikitextSyntaxHighlighter.HEADING.matcher(_hlEditor.getText());
```
harshad1, line 292: Needs to be tested - I don't use wikitext

harshad1:
@aisbergde I think it is up to you whether you want this feature for asciidoc too

gsantner:
@harshad1   
Thanks for the PR. Just, are you also about to use it/did you miss this or anything specific how you came up with the idea?  
Humble opinion it might rather feel like bloat for people using the TOC/heading search.

aisbergde:
To answer the question: AsciiDoc Syntax has built in toc support, and also optional settings for the depth:
https://docs.asciidoctor.org/asciidoc/latest/toc/levels/

Btw, this build in toc was the reason, why I had a look on AsciiDoc: There was something missing in GitHub GUI, related to toc, I don't remember what, but they did not implement. And there was a comment / solution about AsciiDoc native toc support. And based on this comment I started to investigate AsciiDoc.

harshad1:
@aisbergde This is for edit mode. i.e. to jump to a heading when you tap on the title.

@gsantner I tend to maintain very long log files (one per month) and jumping to a specific section to look something up can a be a pain as I have to scroll through all the lower level headings too.

This should not affect users who don't need it. It isn't a 2 stage process. The filter button is just there in the heading dialog if wanted.

harshad1:
headings

aisbergde:
Now I see it! This could be very useful when editing longer files. Most time I prefer to edit longer files on a desktop PC, using VSC or idea, where I can have edit and preview windows side by side and the preview can be used for navigation. This is not possible in Markor, but this feature could solve the issue. I like the idea!

gsantner:
Feel free to also create a merge request to implement this for adoc

PR #2052, opened 2023-05-26
Chunked undo redo

harshad1:
In this PR, the undo - redo system has been updated to chunk inputs.

As the user types, we try to undo one word at a time rather than 1 character. This is much more useful and is how most editors behave.


harshad1:
Now handles trailing spaces. I.e. they are grouped with the previous word

Before:
```
foo bar|
```

After:
```
foo bar baz |
```

Undo:
```
foo bar |
```

Cursor seems to be positioned badly after some changes (renumbering). Need to see if this can be easily fixed.

harshad1:
All issues resolved. This is ready for review / merging

gsantner:
Thanks, merged!

PR #2062, opened 2023-06-05
Add line numbers support

guanglinn:
1. Add line numbers support for Edit mode and code blocks in View mode;
2. Update Prism from v1.17.1 to v1.29.0.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
+++ b/app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
@@ -70,8 +70,7 @@ public class HighlightingEditor extends AppCompatEditText {
     private int _defaultPaddingLeft;
     private static final int LINE_NUMBERS_PADDING_LEFT = 14;
     private static final int LINE_NUMBERS_PADDING_RIGHT = 10;
-    private final int[] _firstVisibleLine = {-1, -1}; // {line index, actual line number}
-    private boolean _firstVisibleLineUpdated;
+    private final int[] _firstVisibleLine = {-1, 0}; // {line index, actual line number}
 
 
     public HighlightingEditor(Context context, AttributeSet attrs) {
@@ -150,8 +149,8 @@ public class HighlightingEditor extends AppCompatEditText {
         super.onAttachedToWindow();
         _scrollView = getParent() instanceof ScrollView ? (ScrollView) getParent() : (ScrollView) getParent().getParent();
         _scrollView.setOnScrollChangeListener((v, scrollX, scrollY, oldScrollX, oldScrollY) -> {
-            if (_firstVisibleLineUpdated) {
-                _firstVisibleLineUpdated = false;
+            if (_firstVisibleLine[0] > -1) {
+                _firstVisibleLine[0] = -1;
             }
         });
     }
@@ -188,45 +187,44 @@ public class HighlightingEditor extends AppCompatEditText {
             setPadding(_x + LINE_NUMBERS_PADDING_RIGHT + 10, getPaddingTop(), getPaddingRight(), getPaddingBottom());
         }
 
-        // Draw the right border
+        // Draw the vertical line
         _paint.setColor(Color.LTGRAY);
         canvas.drawLine(_x + LINE_NUMBERS_PADDING_RIGHT, top, _x + LINE_NUMBERS_PADDING_RIGHT, bottom, _paint);
 
-        // Draw first line number
+        // Draw line numbers
         _paint.setColor(Color.GRAY);
-        canvas.drawText(String.valueOf(1), _x, layout.getLineBounds(0, null) + offsetY, _paint);
+        canvas.drawText("1", _x, layout.getLineBounds(0, null) + offsetY, _paint);
 
-        // Draw other line numbers
-        if (text != null) {
-            final int count = getLineCount();
-            int i = 1, number = 1;
+        if (text == null || text.length() == 0) {
+            return;
+        }
 
-            if (_firstVisibleLineUpdated) {
-                // Set and then iterate from the first visible line
-                i = _firstVisibleLine[0];
-                number = _firstVisibleLine[1];
-            } else {
-                // Set the first visible line invalid, it needs to be updated
-                _firstVisibleLine[0] = -1;
-                _firstVisibleLine[1] = -1;
-            }
+        final int count = getLineCount();
+        int i = 1, number = 1;
 
-            for (int y; i < count; i++) {
-                if (text.charAt(layout.getLineStart(i) - 1) == '\n') {
-                    number++;
-                    y = layout.getLineBounds(i, null);
-                    if (y > bottom) {
-                        break;
+        if (_firstVisibleLine[0] > -1) {
+            // Set and then iterate from the first visible line
+            i = _firstVisibleLine[0];
+            number = _firstVisibleLine[1];
+        } else {
+            // Set the first visible line invalid, it needs to be updated
+            _firstVisibleLine[0] = -1;
+        }
+
+        for (int y; i < count; i++) {
+            if (text.charAt(layout.getLineStart(i) - 1) == '\n') {
+                number++;
+                y = layout.getLineBounds(i, null);
+                if (y > bottom) {
+                    break;
+                }
+                if (y > top) {
+                    if (_firstVisibleLine[0] < 0) {
+                        // Update the first visible line
+                        _firstVisibleLine[0] = i;
+                        _firstVisibleLine[1] = number - 1;
                     }
-                    if (y > top) {
-                        if (_firstVisibleLine[0] < 0) {
-                            // Update the first visible line
-                            _firstVisibleLine[0] = i;
-                            _firstVisibleLine[1] = number - 1;
-                            _firstVisibleLineUpdated = true;
-                        }
-                        canvas.drawText(String.valueOf(number), _x, y + offsetY, _paint);
-                    }
+                    canvas.drawText(String.valueOf(number), _x, y + offsetY, _paint);
                 }
             }
         }


app/src/main/java/net/gsantner/markor/model/AppSettings.java
```diff
@@ -490,6 +498,11 @@ public boolean getDocumentHighlightState(final String path, final CharSequence c
         return getBool(PREF_PREFIX_HIGHLIGHT_STATE + path, lengthOk && isHighlightingEnabled());
     }
 
+    public boolean getEditorLineNumbersState(final CharSequence chars) {
+        final boolean lengthOk = chars != null && chars.length() < (_isDeviceGoodHardware ? 100000 : 35000);
```
gsantner, line 502: I see you used some performance related values here... How well is performance overall, does it greatly impact i.e. scrolling or loading speed? Last time I visited this, it made everything horribly slow

app/src/main/res/xml/preferences_master.xml
```diff
@@ -350,6 +350,12 @@
                 android:key="@string/pref_key__editor_enable_line_breaking"
                 android:summary="@string/soft_break_if_not_fit_to_width_else_enable_horizontal_scrolling"
                 android:title="@string/wrap_words" />
+            <CheckBoxPreference
+                android:defaultValue="false"
+                android:icon="@drawable/ic_line_numbers_24dp"
+                android:key="@string/pref_key__editor_line_numbers"
+                android:summary="@string/show_line_numbers_on_the_left_side_of_the_editor"
+                android:title="@string/show_line_numbers" />
```
gsantner, line 358: I suggest this time this approach on settings and on this topic:

* Should be enough to have one settings key like "show_line_numbers", which enables both line numbers at code blocks (view mode) as well at editor.
* No new option at Settings screen (=remove the 2 added blockes at prefereences_master.xml)
* Just have one isShowLineNumbers() method for checking if it is enabled. And a toggleShowLineNumbers() which toggles it globally. So one global (non-per-file-specific) setting.

Would that be OK to you? It also means the logic gets a bit easier and less 😄 .

app/thirdparty/assets/prism/Prism v1.29.0
gsantner, line 1: this seems to be a empty file? Would delete it and add uppercase `Prism*` to .gitignore

app/thirdparty/assets/prism/index.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
```
gsantner, line 1: looks like some prism example, which we don't need. Would delete it and add it to .gitignore

app/thirdparty/java/other/flexmark/ext/codeblocks/LineNumbersAttributeProvider.java
```diff
@@ -0,0 +1,33 @@
...
+import com.vladsch.flexmark.html.IndependentAttributeProviderFactory;
+import com.vladsch.flexmark.html.renderer.AttributablePart;
+import com.vladsch.flexmark.html.renderer.LinkResolverContext;
+import com.vladsch.flexmark.util.ast.Node;
+import com.vladsch.flexmark.util.html.Attributes;
+
+/**
+ * AttributeProvider of flexmark-java extension for prism syntax highlighter.
+ */
+public class LineNumbersAttributeProvider implements AttributeProvider {
```
gsantner, line 15: Suggestion: Move LineNumbersAttributeProvider and LineNumbersExtension to one class (with 2 inner classes) together.

Did you write this on your own or is it a copy from some suggestion at some flexmark discussion? If so, attribution is required

gsantner (commented):
Thank you very much for your help! I looked through and added my thoughts

app/thirdparty/assets/prism/Prism v1.29.0
guang-lin: It's just used to indicate which version of Prism was used, so it can be deleted, or add it to .gitignore.

app/thirdparty/assets/prism/index.html
guang-lin: Yes, but it is convenient to test line numbers display appearance(e.g. line numbers fence padding) in browser instead of rebuilding, by custom the file /prism/plugins/line-numbers/prism-line-numbers.css.

app/src/main/res/xml/preferences_master.xml
guang-lin: OK, maybe one is enough. I just think that someone may always hope to enable line numbers, so they can open it at Settings screen. another option for others just occasionally enable line numbers, they don't to need to Settings screen to switch. You can make some adjustments with your ideas 😄.

app/src/main/java/net/gsantner/markor/model/AppSettings.java
guang-lin: Sorry, I didn't do enough testing. It is important but I'm not good at testing, could you have a test for it.

app/thirdparty/java/other/flexmark/ext/codeblocks/LineNumbersAttributeProvider.java
guang-lin: I referenced codes at https://github.com/riversun/flexmark-ext-prism-linenumber.git. Sorry, I didn't add attribution.

k3b (commented):
From my point of view the code looks good

app/thirdparty/java/other/flexmark/ext/codeblocks/LineNumbersAttributeProvider.java
k3b: @guang-lin  can you also update app/src/main/res/raw/licenses_3rd_party.md ?

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -46,6 +49,7 @@ public class HighlightingEditor extends AppCompatEditText {
     private boolean _isDynamicHighlightingEnabled = true;
     private Runnable _hlDebounced;        // Debounced runnable which recomputes highlighting
     private boolean _hlEnabled;           // Whether highlighting is enabled
+    private boolean _nuEnabled;           // Whether show line numbers is enabled
```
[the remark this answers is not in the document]
guang-lin: The name style of "_nuEnabled" just referenced to "_hlEnabled", but I also think that _showLineNumbersEnabled or _LineNumbersEnabled maybe more suitable.

app/src/main/res/values-zh-rCN/strings.xml
```diff
@@ -308,6 +308,9 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="list_in_recent_files">最近文件列表</string>
     <string name="reload">刷新</string>
     <string name="wrap_words">折行</string>
+    <string name="show_line_numbers">显示行号</string>
```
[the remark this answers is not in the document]
guang-lin: I see, what should I do, do I put them to string-not_translatable.xml?

app/thirdparty/java/other/flexmark/ext/codeblocks/LineNumbersAttributeProvider.java
guang-lin: OK, I will.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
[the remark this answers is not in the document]
gsantner: suggestion: showLineNumbers

app/src/main/res/values-zh-rCN/strings.xml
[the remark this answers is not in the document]
gsantner: generally dont add translations (as in i.e. chinese) - only english. And here, if it is not translatable (like Markor or Markdown is not) put it in the not_translatable file. And otherwise in the strings.xml.

Translations will come then later in after merge.

k3b:
Note: Today i go on holyday for 3 weeks where i have no computer so i cannot do any code reviews in this time

guang-lin:
> Note: Today i go on holyday for 3 weeks where i have no computer so i cannot do any code reviews in this time

Fine, have a good holiday!

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -54,6 +58,13 @@ public class HighlightingEditor extends AppCompatEditText {
     private boolean _autoFormatEnabled;
     private boolean _saveInstanceState = true;
 
+    // For drawing line numbers fence
+    private Paint paint = new Paint();
+    private int lineNumbersFencePaddingLeft = 16;
+    private int lineNumbersFencePaddingRight = 10;
+    private int defaultPaddingLeft;
```
harshad1, line 65: Should these be `private static final CAPITAL_CASE`?

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -95,6 +107,52 @@ public void afterTextChanged(final Editable s) {
...
+        }
+    }
+
+    private void drawLineNumbers(Canvas canvas) {
+        // Draw line numbers
+        paint.setColor(Color.GRAY);
+        int firstBaselineToTopHeight = getPaddingTop() - getPaint().getFontMetricsInt().top;
+        int x = getLeft() + lineNumbersFencePaddingLeft;
+        int y = getTop() + firstBaselineToTopHeight;
+        int number = 1;
```
harshad1, line 135: IMO marking things final if they can be final is good for readability

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -95,6 +107,52 @@ public void afterTextChanged(final Editable s) {
...
     }
 
+    @Override
+    protected void onFinishInflate() {
+        super.onFinishInflate();
+        defaultPaddingLeft = getPaddingLeft();
+    }
+
+    @Override
+    protected void onDraw(Canvas canvas) {
```
harshad1, line 117: Good thing to test: see how line numbers behave if you
1. disable / enable Line wrapping
2. disable / enable highlighting (especially with bigger headings enabled)
3. Change text size

when line numbers are enabled. These should all trigger a redraw, but it would be good to check

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -95,6 +107,52 @@ public void afterTextChanged(final Editable s) {
...
+        int firstBaselineToTopHeight = getPaddingTop() - getPaint().getFontMetricsInt().top;
+        int x = getLeft() + lineNumbersFencePaddingLeft;
+        int y = getTop() + firstBaselineToTopHeight;
+        int number = 1;
+        // Draw first line number
+        canvas.drawText(String.valueOf(number), x, y, paint);
+        // Draw others line number
+        for (int i = 1; i < getLineCount(); i++) {
+            y = y + getLineHeight();
+            if (getText().charAt(getLayout().getLineStart(i) - 1) == '\n') {
```
harshad1, line 141: Get references to the layout and editable outside the loop.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -95,6 +107,52 @@ public void afterTextChanged(final Editable s) {
...
+            drawLineNumbers(canvas);
+        } else if (getPaddingLeft() != defaultPaddingLeft) {
+            // Reset padding without line numbers fence
+            setPadding(defaultPaddingLeft, getPaddingTop(), getPaddingRight(), getPaddingBottom());
+        }
+    }
+
+    private void drawLineNumbers(Canvas canvas) {
+        // Draw line numbers
+        paint.setColor(Color.GRAY);
```
harshad1, line 131: Does this color work well in all cases? (Light vs dark etc).

iirc we use dark grey in some places for better readability,

guang-lin: It can work well in dark mode, like this

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
guang-lin: Now, it all work well with these options.

app/src/main/java/net/gsantner/markor/model/AppSettings.java
gsantner: Have a moderate to big textfile and scroll a bit. You will immediately notice when it's slow.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
gsantner: Looks nice! 😄

app/src/main/java/net/gsantner/markor/model/AppSettings.java
harshad1: We should only draw line numbers for the visible lines +- 2 pages.

The highlighter class does this. We can look at it / reuse some code imo.

guang-lin: First, the algorithm or method for displaying line numbers is still worth optimizing. Second, the editor's loading strategy for large files may also need to be optimized, such as just loading nearby 1000 lines at a time, which can gracefully load large files. I could try to optimize it if I'm free.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -178,6 +237,19 @@ public boolean setHighlightingEnabled(final boolean enable) {
         return prev;
     }
 
+    public boolean getLineNumbersEnabled() {
+        return _nuEnabled;
+    }
+
+    public boolean setLineNumbersEnabled(final boolean enable) {
+        final boolean prev = _hlEnabled;
```
k3b, line 245: why does setLineNumbersEnabled return the old value of `_hlEnabled` instead of `_nuEnabled` ?

Since the returnvalue of `setLineNumbersEnabled(...)` is never used you can also make the method return void.

k3b (changes requested):

k3b:
From my point of view every thing else looks good

app/src/main/java/net/gsantner/markor/model/AppSettings.java
harshad1: @guang-lin Note that Incrementally loading files with Android can be a pain as the underlying file system is unreliable. Our current system (trying to save multiple times etc) is hacky but prevents data loss 99.99% of the time.

harshad1:
What is the status here? Any way I can help move this along?

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -95,6 +107,66 @@ public void afterTextChanged(final Editable s) {
...
+            bottom = scrollView.getScrollY() + scrollView.getHeight();
+        }
+
+        final int count = getLineCount();
+        final Editable text = getText();
+        final Layout layout = getLayout();
+        for (int i = 1; i < count; i++) {
+            if (text.charAt(layout.getLineStart(i) - 1) == '\n') {
+                number++;
+                y = layout.getLineTop(i);
```
harshad1, line 154: How does this look in todo.txt where we add extra padding to the top of the line?

guang-lin: Here is the screenshot, not bad.

guang-lin:
> What is the status here? Any way I can help move this along?

I think the status is stable. It can be better improved, if there are better ideas in the future.

gsantner:
Hello, I try to find some time on this (hot) weekend to review again and hopefully merge this week.

Regarding todo.txt, would it be possible to add a offset at the number so it better fits to the line? That offset can be conditionally activated by checking the format

harshad1:
> Hello, I try to find some time on this (hot) weekend

You're in Austria Right? I was in Germany a few weeks ago and it was quite warm :)

> Regarding todo.txt, would it be possible to add a offset

I think we can look at Font Metrics and align the bottom of the number to the line's baseline. This will be uniform in all cases and will not require special a case for Todo.

gsantner:
Yes!

Hm, I think at the end it would be the best solution if the text is vertical-centered to the line. Espcially also when it's multiline (regarding line wrapping)

harshad1:
> Hm, I think at the end it would be the best solution if the text is vertical-centered to the (full) line

I have to disagree. Vim and VsCode both align the bottom of the0 number to bottom of the first letter in the line. I think this is now it should go.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -95,6 +107,66 @@ public void afterTextChanged(final Editable s) {
...
+    private void drawLineNumbers(Canvas canvas) {
+        final int firstBaselineToTopHeight = getPaddingTop() - getPaint().getFontMetricsInt().top;
+        int number = 1;
+
+        _paint.setColor(Color.GRAY);
+        // Draw first line number
+        canvas.drawText(String.valueOf(number), LINE_NUMBERS_PADDING_LEFT, firstBaselineToTopHeight, _paint);
+        // Draw others line number
+        float y;
+        final float offsetY = firstBaselineToTopHeight - getTextSize() * 0.14f;
```
harshad1, line 138: Why 0.14?

gsantner:
Fine too I guess. But at last how it is currently it is really bad for todo.txt. The number at the side is rather a rough hint of amount of lines, rather then getting to know what line exactly it is in.

It would be also OK to improve that later and force disable it for todo.txt for now, but it's not good for this case.

Example from another editor, gedit. What I noted besides is that it also right-aligns text instead of left

a

gsantner:
pushed some improvements:

* Simplified line number settings - only one button now that works as a global setting for editor+view line numbers
* Reformat code
* Minor improvements
* Added license notice to flexmark line number files

gsantner:
with the todo.txt improvement I pushed 
screenshot-2023-07-09__20-29-36

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
gsantner: with the todo.txt improvement I pushed 
screenshot-2023-07-09__20-29-36

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
gsantner: line spacing I guess

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
guang-lin: It looks better 😊

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
guang-lin: It is a vertical downward offset with using getLineTop(i), 0.14 was got by trial.

guang-lin:
I have aligned line numbers text to the right.

gsantner:
@guang-lin 
Thanks for additional improvements. Do you think it's ready to merge now?

guang-lin:
> @guang-lin Thanks for additional improvements. Do you think it's ready to merge now?

I think yes. It's my pleasure to contribute to this project. 😊

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -95,6 +136,72 @@ public void afterTextChanged(final Editable s) {
...
+        _paint.setTextAlign(Paint.Align.RIGHT);
+        _scrollView = getParent() instanceof ScrollView ? (ScrollView) getParent() : (ScrollView) getParent().getParent();
+        return super.onPreDraw();
+    }
+
+    @Override
+    protected void onDraw(Canvas canvas) {
+        super.onDraw(canvas);
+        // If line numbers can be drawn
+        if (_nuEnabled && _maxLineNumber < (AppSettings._isDeviceGoodHardware ? 5000 : 3000)) {
```
harshad1, line 157: As line numbers have to be enabled manually, perhaps we should just issue a toast warning instead of hard blocking on `isDeviceGoodHardware`

harshad1: We should also just try to make line numbers dynamic (draw them only for visible lines). That will avoid this completely. Lets remove this test for the line count at first and then I can help with the dynamic line numbers...

guang-lin:
Now supports iterating from the first visible line rather than the second line if the scroll view is in a stopped state, even if at around 30,000 lines.

https://github.com/gsantner/markor/assets/60415467/f66af769-a5d1-4b30-bc1c-b21e2a62dc8b

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -95,6 +138,100 @@ public void afterTextChanged(final Editable s) {
...
+        // Draw first line number
+        _paint.setColor(Color.GRAY);
+        canvas.drawText(String.valueOf(1), _x, layout.getLineBounds(0, null) + offsetY, _paint);
+
+        // Draw other line numbers
+        if (text != null) {
+            final int count = getLineCount();
+            int i = 1, number = 1;
+
+            if (_firstVisibleLineUpdated) {
```
harshad1, line 204: Do we need the variable `_firstVisibleLineUpdated`? I think we can check `_firstVisibleLine[0] >= 0`? I could be wrong here.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -95,6 +138,100 @@ public void afterTextChanged(final Editable s) {
...
+                // Set and then iterate from the first visible line
+                i = _firstVisibleLine[0];
+                number = _firstVisibleLine[1];
+            } else {
+                // Set the first visible line invalid, it needs to be updated
+                _firstVisibleLine[0] = -1;
+                _firstVisibleLine[1] = -1;
+            }
+
+            for (int y; i < count; i++) {
```
harshad1, line 214: since y is not the loop variable, I think it should not be declared here.

Just do 

`for (; i < count; i++)`

`final int y = layout.getLineBounds(i, null);`

harshad1: I am attempting to make these changes in a local branch. If they work I will push it so you can pull them in if you think they work

gsantner:
Changed the branch to linenumbers branch. Merging this PR and improvements can be continued via additional PRs to linenumbers branch. Then we can merge it to master when you think it's fine.

Thanks for the contribution!

PR #2075, opened 2023-07-13
Remove Trailing Space in Link Formating

ruedigerblock:
Hello !

If I share a URL from my Browser App into Markor - Quicknote, the result contains a space before the trailing ")" which results in vim not opening the URL and throwing an error instead.
IMHO the trailing Space is not needed and caused by a typo.


gsantner (commented):
Thank you.

This space is there intentionally. When you re-share text containing a link to other apps - like WhatsApp, Telegram, .... they usually have a link preview for the first link that is discovered and automatically make the links clickable.
  
What happens when you re-share a text that is `https://something)`, every other application with such a feature takes as it is with a `)` at the end to display preview and open the link. Leading to that getting a page opened with wrong content/404 as they are wrong - `https://something/a.html` vs `https://something/a.html)`.

While I can't ask those applications to adjust their stuff to better work with some Markdown text...vim on on the other hand is very adaptive. I'm pretty sure you will find a way to configure it. Also here in my terminal application links are underlined pretty fine and are openable by hover-rightclick.

screenrecord-2023-07-13_20 06 09

So with Markor I can at last make something that works together fine with the rest and for vim I'm pretty sure you will find a solution.

PR #2081, opened 2023-08-04
Snippet token for inserting selection (closes #2059)

harshad1:
This addresses https://github.com/gsantner/markor/discussions/2059 

The token `%%INSERT_SELECTION_HERE%%` will be replaced with any selected text. This token can be used multiple times.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
+++ b/app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
@@ -367,6 +367,7 @@ public class HighlightingEditor extends AppCompatEditText {
         final Editable edit = getText();
         if (edit != null && newText != null) {
 
+            // TODO - should consider moving any snippet specific logic out of here
             // Fill in any instances of selection
             final int[] sel = TextViewUtils.getSelection(this);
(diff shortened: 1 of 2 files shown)


harshad1:
@gsantner

I am not a fan of the %%FOO_BAR%% token format. Do you have any alternative ideas. Shorter tokens would be nicer too

gsantner:
@harshad1 
At the end it has to be something that doesn't conflict with text users might have as real text in it. And what's quite common for these kind of things are `{{token}}` / `%token%` / `%%token%%`.  Though as you shouldn't be creating templates again and again I think it shouldn't matter that much, you place it like once and that#s it.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -190,6 +191,11 @@ private int[] hlRegion(final Rect rect) {
         }
     }
 
+    @Override
+    public boolean bringPointIntoView(int i) {
+        return super.bringPointIntoView(i);
+    }
```
gsantner, line 197: Is there a reason for this? As far I see it's a override that just calls the original implementation

harshad1:
NOTE: this code got included in the line numbers branch.

harshad1:
Closed as this is included in #2090

PR #2088, opened 2023-08-12
Reworking line numbers + other fixes

harshad1:
@gsantner @guang-lin

I reworked some of the line number stuff in this PR:

- Simplified how the line number calculations and drawing are done.
- Make the line number setting per-file

Please take a look and feel free to take anything you feel is useful.

PR #2090, opened 2023-08-15
Line numbers improvements

guang-lin:
1. Performance improvements
(1) avoid unnecessary loops;
(2) avoid unnecessary parameter updates.
2. Move the codes to a separate class.

based on #2088, #2062

### DIFF
--- a/app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
+++ b/app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
@@ -60,15 +60,8 @@ public class HighlightingEditor extends AppCompatEditText {
     private TextWatcher _autoFormatModifier;
     private boolean _autoFormatEnabled;
     private boolean _saveInstanceState = true;
+    private final LineNumbersDrawer _lineNumbersDrawer = new LineNumbersDrawer(this);
 
-    // For drawing line numbers
-    private final Paint _paint = new Paint();
-    private int _defaultPaddingLeft;
-    private static final int LINE_NUMBERS_PADDING_LEFT = 14;
-    private static final int LINE_NUMBERS_PADDING_RIGHT = 10;
-    private final Rect _numRect;
-
-    private final List<Integer> _numYPositions; // Y positions of numbers to draw
 
     public HighlightingEditor(Context context, AttributeSet attrs) {
         super(context, attrs);
@@ -86,11 +79,8 @@ public class HighlightingEditor extends AppCompatEditText {
         _numEnabled = false;
         _oldHlRect = new Rect();
         _hlRect = new Rect();
-        _numRect = new Rect();
-        _numYPositions = new ArrayList<>();
 
         addTextChangedListener(new GsTextWatcherAdapter() {
-
             @Override
             public void onTextChanged(CharSequence s, int start, int before, int count) {
                 if (_hlEnabled && _hl != null) {
@@ -111,20 +101,19 @@ public class HighlightingEditor extends AppCompatEditText {
         observer.addOnScrollChangedListener(() -> updateHighlighting(false));
         observer.addOnGlobalLayoutListener(() -> updateHighlighting(false));
 
-        // Fix for android 12 perf issues - https://github.com/gsantner/markor/discussions/1794
+        // Fix for Android 12 perf issues - https://github.com/gsantner/markor/discussions/1794
         setEmojiCompatEnabled(false);
     }
 
     @Override
     protected void onFinishInflate() {
         super.onFinishInflate();
-        _defaultPaddingLeft = getPaddingLeft();
-        _paint.setTextAlign(Paint.Align.RIGHT);
+        _lineNumbersDrawer.setDefaultPaddingLeft(getPaddingLeft());
     }
 
     @Override
     public boolean onPreDraw() {
-        _paint.setTextSize(getTextSize());
+        _lineNumbersDrawer.setTextSize(getTextSize());
         return super.onPreDraw();
     }
 
@@ -132,68 +121,10 @@ public class HighlightingEditor extends AppCompatEditText {
     protected void onDraw(Canvas canvas) {
         super.onDraw(canvas);
 
-        // If line numbers can be drawn
         if (_numEnabled) {
-            drawLineNumbers(canvas);
-        } else if (getPaddingLeft() != _defaultPaddingLeft) {
-            // Reset padding without line numbers fence
-            setPadding(_defaultPaddingLeft, getPaddingTop(), getPaddingRight(), getPaddingBottom());
-        }
-    }
-
-    private void drawLineNumbers(final Canvas canvas) {
-        if (!getLocalVisibleRect(_numRect)) {
-            return;
-        }
-
-        final CharSequence text = getText();
-        final Layout layout = getLayout();
-        if (text == null || layout == null) {
-            return;
-        }
-
-        // Current visible area
-        final int height = _hlRect.height();
-        final int top = _hlRect.top - height;
-        final int bottom = _hlRect.bottom + height;
-
-        final int offsetY = getPaddingTop();
-        final int maxLayoutLine = layout.getLineCount();
-
-        // We make a single pass through the text to determine
-        // 1. y positions and line numbers we want to draw
-        // 2. max line number (to gauge gutter width)
-        _numYPositions.clear();
-        int startNumber = 1, maxNumber = 1;
-        for (int i = 0; i < maxLayoutLine; i++) {
-            final int start = layout.getLineStart(i);
-            if (start == 0 || text.charAt(start - 1) == '\n') {
-                final int numY = layout.getLineBaseline(i);
-                if (numY < top) {
-                    startNumber++;
-                } else if (numY < bottom) {
-                    _numYPositions.add(numY);
-                }
-                maxNumber++;
-            }
-        }
-
-        // Draw the gutter
-        final int width = Math.round(_paint.measureText(String.valueOf(maxNumber)));
-        final int numX = LINE_NUMBERS_PADDING_LEFT + width;
-        final int lineX = numX + LINE_NUMBERS_PADDING_RIGHT;
-        final int padding = lineX + 10;
-        if (padding != getPaddingLeft()) {
-            setPadding(padding, getPaddingTop(), getPaddingRight(), getPaddingBottom());
-        }
-        _paint.setColor(Color.LTGRAY);
-        canvas.drawLine(lineX, top, lineX, bottom, _paint);
-
-        _paint.setColor(Color.GRAY);
-        for (int i = 0; i < _numYPositions.size(); i++) {
-            final String number = String.valueOf(startNumber + i);
-            final float numY = _numYPositions.get(i) + offsetY;
-            canvas.drawText(number, numX, numY, _paint);
+            _lineNumbersDrawer.draw(canvas);
+        } else {
+            _lineNumbersDrawer.suspend();
         }
     }
 
@@ -212,7 +143,6 @@ public class HighlightingEditor extends AppCompatEditText {
 
             // Don't highlight unless shifted sufficiently or a recompute is required
             if (recompute || (visible && _hl.hasSpans() && isScrollSignificant())) {
-
                 _oldHlRect.set(_hlRect);
 
                 final int[] newHlRegion = hlRegion(_hlRect); // Compute this _before_ clear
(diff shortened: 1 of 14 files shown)


app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -111,89 +101,30 @@ public void afterTextChanged(final Editable s) {
...
-        canvas.drawLine(lineX, top, lineX, bottom, _paint);
-
-        _paint.setColor(Color.GRAY);
-        for (int i = 0; i < _numYPositions.size(); i++) {
-            final String number = String.valueOf(startNumber + i);
-            final float numY = _numYPositions.get(i) + offsetY;
-            canvas.drawText(number, numX, numY, _paint);
+            _lineNumbersDrawer.draw(canvas);
+        } else {
+            _lineNumbersDrawer.suspend();
```
harshad1, line 124: Instead of calling this here I think calling this in setLineNumbersEnabled(false) is sufficient.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -535,4 +465,156 @@ public int setSelectionExpandWholeLines() {
...
+                _maxNumberDigits = 3;
+            } else if (number < 10000) {
+                _maxNumberDigits = 4;
+            } else {
+                _maxNumberDigits = 5;
+            }
+            return _maxNumberDigits;
+        }
+
+        public boolean isVisibleAreaChanged() {
```
harshad1: See `isScrollSignificant()` https://github.com/gsantner/markor/blob/master/app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java#L101-L104

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -535,4 +465,156 @@ public int setSelectionExpandWholeLines() {
...
+                            _numberYPositions.add(y);
+                        }
+                        _maxNumber++;
+                    }
+                }
+            }
+
+            // Only if the text size or the max line number of digits changed
+            // Update the gutter parameters and set padding left
+            if (_paint.getTextSize() != _lastTextSize || getNumberDigits(_maxNumber) != _lastMaxNumberDigits) {
```
harshad1, line 613: So in the version I wrote I just compare the newly computed padding to the old padding.

I don't see how checking text size or max digits is more comprehensive. If either of these changes then the computed padding will change and that is sufficient to trigger an update.

IMO we are doing extra work here which is not necessary. (But please correct me if I am wrong)

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -535,4 +465,156 @@ public int setSelectionExpandWholeLines() {
...
+            for (int i = 0; i < _numberYPositions.size(); i++) {
+                final String number = String.valueOf(_startNumber + i);
+                canvas.drawText(number, _numberX, _numberYPositions.get(i) + offsetY, _paint);
+            }
+        }
+
+        /**
+         * Suspend drawing the line numbers.
+         */
+        public void suspend() {
```
harshad1, line 635: IMO this method should be called `clear` and should be called only once when line numbers are disabled, not every time in onDraw.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -546,4 +462,182 @@ public int setSelectionExpandWholeLines() {
...
+            }
+
+            final int offsetY = _editor.getPaddingTop();
+            final int layoutLineCount = layout.getLineCount();
+
+            // Only if the visible area is out of current line numbers area or the layout line count changed
+            // We make a single pass through the text to determine
+            // 1. y positions and line numbers we want to draw
+            // 2. max line number (to gauge gutter width)
+            if (isOutOfLineNumbersArea() || layoutLineCount != _lastLayoutLineCount) {
```
harshad1, line 590: I don't think we should test this. I think we should redraw line numbers for _every_ `onDraw`.

This will automatically account for font size changes, highlighting state changes etc etc.

The performance penalty seems small.

harshad1: Also, not doing this test will eliminate the need for many state tracking variables, which is always good.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -546,4 +462,182 @@ public int setSelectionExpandWholeLines() {
...
+                private final Pattern pattern = Pattern.compile("\n");
+                private Matcher matcher;
+
+                @Override
+                public void beforeTextChanged(CharSequence s, int start, int count, int after) {
+                    if (after == 0 && count > 0) {
+                        CharSequence deleted = s.subSequence(start, start + count);
+                        matcher = pattern.matcher(deleted);
+                        while (matcher.find()) {
+                            _maxNumber--;
```
harshad1, line 513: Is tracking maxNumber really necessary? I don't in my version (I compute this on the fly in every onDraw) and there is not visible performance issue even on a very large file (10000 lines+)

harshad1:
I made a bunch of comments here. Please ask for clarification if needed. 

IMO we are still over complicating several things here with minimal benefit

gsantner:
there is a class around called NanoProfiler. Maybe it's helpful to get an idea how it overall performs (onDraw) or various sections of it.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
guang-lin: The padding wont automatically resize when the text size or the max line numbers of digits changed, so if just compare the newly computed padding with the old padding, this needs to update several parameters about the padding (especially the calculation of text width) every time, I think this is very inefficient, we know onDraw() be called very frequently.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
guang-lin: Fine, I will improve it.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
guang-lin: I used isOutOfLineNumbersArea() instead of isVisibleAreaChanged() later. About detection of line numbers area, drawing line numbers have some different requirements than highlight. Besides, using `LineNumbersDrawer` class own detection method is also for the sake of improving the cohesion and reduceing coupling.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
guang-lin: Note this will still redraw line numbers for every onDraw via using value of `List` assigned last time, and only recalculate it in the loop when the status changed, instead of executing the loop every onDraw.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
guang-lin: I think it is necessary. For example, a very large file (10000 lines+), when we view it at around 100th line, your method will still iterate from 1 to 10000+ (when executing the loop every time, and do this very frequently when scrolling), but my method will break the loop at around 100+, because it doesn't need to calculate the max line number.

guang-lin:
We can test a time consumption to the method in practice, and then decide which strategies can be used.

guang-lin:
> As well, I think we should move on at some point :D. After rewriting it like 10 times it probably isn't fun anymore ^^.

Indeed

harshad1:
Agreed too.

My only concern is that we are doing a lot that we don't need to be doing - making it harder to debug etc etc. I'm pushing for simplifying.

I am willing to take on a good amount of the work + maintenance in the service of this.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
harshad1: True., but I don't see a performance hit from this (I am running my branch on my phone for daily use). I can flick or drag scroll the entire text of Homer without any issues.

Remember that the textview has to layout the entire text anyway. Getting a line count is relatively trivial. 

I think optimizing for the performance at the end of the file (where one is more likely to view / edit) is what matters and both methods have the same performance here.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
harshad1: As mentioned with other comments, I don't see a visible performance penalty from just running this loop every time. IMO this is kind of a premature optimization.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
harshad1: Same observation about not seeing a performance impact in real life.

harshad1:
IMO lets just merge this soon. 

We can do performance experiments and make simplifying changes in follow up PRs if necessary. The split into a contained class makes this very clean and easy to do (thanks @guang-lin 👍 )

guang-lin:
Your opinions are also reasonable. If there is anything inappropriate, you can adjust it. Additionally, it was your suggestions that led to significant improvements in these codes, with many ideas that I had not anticipated. I have gained a lot, thank you @harshad1 .

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -546,4 +463,181 @@ public int setSelectionExpandWholeLines() {
...
```
harshad1, line 513: If we are going to keep this I suggest

`_maxLineNumber += TextViewUtils.countChar(s, start, start + count, '\n')`
`_maxLineNumber -= TextViewUtils.countChar(s, start, start + count, '\n')`

Simpler, faster, avoids need for creating a matcher and subsequence etc etc.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -546,4 +463,181 @@ public int setSelectionExpandWholeLines() {
...
+        }
+
+        public void setTextSize(float textSize) {
+            if (_lastTextSize != _paint.getTextSize()) {
+                _lastTextSize = _paint.getTextSize();
+            }
+            _paint.setTextSize(textSize);
+        }
+
+        public int getNumberDigits(int number) {
```
harshad1, line 538: `Min(Integer.toString(number), 5)` ?

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -546,4 +463,181 @@ public int setSelectionExpandWholeLines() {
...
+                        while (matcher.find()) {
+                            _maxNumber++;
+                        }
+                    }
+                }
+            });
+        }
+
+        public void setTextSize(float textSize) {
+            if (_lastTextSize != _paint.getTextSize()) {
```
harshad1, line 532: Checking inequality has no value here.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -546,4 +463,181 @@ public int setSelectionExpandWholeLines() {
...
+            }
+
+            // Draw the gutter line
+            _paint.setColor(Color.LTGRAY);
+            canvas.drawLine(_gutterLineX, _top, _gutterLineX, _bottom, _paint);
+
+            // Draw the line numbers
+            _paint.setColor(Color.GRAY);
+            final int size = _lineIndex.size();
+            for (int i = 0; i < size; i++) {
```
harshad1, line 627: The reason I had 2 loops with an ArrayList is because the first loop computes both the maxLineNumber and the number y positions.

If we are going to compute max line number in a text watcher, then this loop can be combined with the first one and we can eliminate the need for the ArrayList and _startNumber etc etc

guang-lin:
I agree with your opinions.

harshad1:
Can I do anything to help get this merged

guang-lin:
I made some adjustments based on your suggestions, and I will submit later.

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -546,4 +461,167 @@ public int setSelectionExpandWholeLines() {
...
+            }
+
+            // Draw the right border of the gutter
+            _paint.setColor(Color.LTGRAY);
+            canvas.drawLine(_gutterX, _lineNumbersArea.top, _gutterX, _lineNumbersArea.bottom, _paint);
+
+            int i = _startLine[0], number = _startLine[1];
+            // Only if the visible area is out of current line numbers area,
+            // update the parameters of the line numbers which will be drawn
+            if (isOutOfLineNumbersArea()) {
```
harshad1, line 588: This will be triggered only if there is a shift right? Not if highlighting state changes or we change text size etc etc (all of which can affect Y Positions).

Do line numbers draw correctly when you toggle highlighting in a todo.txt file with just 2-3 lines

guang-lin: Yes, it behaved well in these situations. It does not save fixed y positions, so it can access and use new y positions of lines if the they adjusted.

harshad1:
I am trying to do some performance profiling and here is what I see:

On a smaller file the total time to onDraw (including super) takes ~3000us of which the line numbers part takes ~500us

On a large file (the entire text of the odyssey) if I enable line numbers the app stalls and doesn't work. (I had to disable the line count check).  I am still working on understanding why, but clearly something is breaking the performance here.

harshad1:
I tried this on my rewrite too. It works with the large file and here are the numbers:

image

`Drawtime` is the total time taken by onDraw
`Line time` is the total time taken by the drawLineNumbers call
`iter time` is the time taken to iterate over every line of time file to determine y positions and max line number

all numbers are in microseconds and are for a file with 10440 lines and 601017 characters

harshad1:
After more testing, and looking at the current code. I think this is ready to merge. IMO we should still investigate ways of making line numbers much faster (there is still some lag with very large files). But it is good enough to merge now

guang-lin:
I agree, at least it is fine now, and there may not be significant improvements in the short term.

harshad1:
Putting this down here for future reference.

One idea is to maintain the newline indices. Possibly by subclassing the DynamicLayout class

This way we would not have to do change tracking in the editor itself and would not have to iterate from the top to find current line numbers

harshad1:
One other thought (for this pr, not the future). Currently we redraw if the visible rect goes out of the last drawn area (3x current height). We should instead redraw if we go 50% (say) from last drawn area. The current system will not redraw until we are outside the drawn area. We should instead redraw _slightly before_ we are outside the drawn area. Easiest way to do this is make the last drawn rect 25% smaller than the actually drawn area.

gsantner:
Ia there a viable open todo here which would be a reason to not merge it?

I would like to merge everything to master and make sure that other PRs/improvements not get into huge rework needs due the amount of changes :D.

Regarding performance, sure good to improve but at the end the feature is disabled by default anyway - so shouldn't hurt people unless they active enable this :-).

guang-lin:
The other PRs/improvements should all be included, these changes are based on the last commits of other PRs/improvements. 
Performance loss may be minor, I have been using it for a while, and I didn't feel some unusual experience, even more than 10,000 lines.

gsantner:
Thanks everybody, I merged everything to master now

harshad1:
@gsantner @guanglinn This PR appears to have broken image rendering in view mode. Still investigating

guanglinn:
I tried just now, it's true, but it works well in my develop branch. I will also help investigate it.

https://github.com/gsantner/markor/assets/60415467/8d32315a-0ccf-4dda-98e2-b6221ab279a2

harshad1:
I believe I have solved the issue. Look at latest commit in 

https://github.com/gsantner/markor/pull/2106

PR #2118, opened 2023-10-03
Editor/viewer: Side margin improvements, closes #2111

guanglinn:
Adjust horizontal margin and padding for Edit mode and View mode.

### DIFF
--- a/app/src/main/res/values/dimens.xml
+++ b/app/src/main/res/values/dimens.xml
@@ -1,7 +1,7 @@
 <resources>
     <!-- Default screen margins, per the Android Design guidelines. -->
-    <dimen name="activity_horizontal_margin">16dp</dimen>
-    <dimen name="activity_vertical_margin">16dp</dimen>
+    <dimen name="activity_horizontal_margin">12dp</dimen>
+    <dimen name="activity_vertical_margin">12dp</dimen>
     <dimen name="editor_bottom_margin">32dp</dimen>
     <dimen name="appbar_padding_top">8dp</dimen>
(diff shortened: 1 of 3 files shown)


guanglinn:
The display area is expanded now.

# View mode

- Before


- After



# Edit mode

- Before


- After

harshad1:
Does this interfere with grabbing the scrollbar for drag scroll on the right side?

Otherwise I like this change.

guanglinn:
The scrollbar is not interfered.

app/src/main/res/values/dimens.xml
```diff
@@ -1,7 +1,7 @@
 <resources>
     <!-- Default screen margins, per the Android Design guidelines. -->
-    <dimen name="activity_horizontal_margin">16dp</dimen>
-    <dimen name="activity_vertical_margin">16dp</dimen>
+    <dimen name="activity_horizontal_margin">12dp</dimen>
+    <dimen name="activity_vertical_margin">12dp</dimen>
```
gsantner, line 4: This affects everything. All activities. Not just editor/viewer.

But let's try. It can be reverted if it is bad.

guanglinn:
Oh, Yes. We could take a look at the experience feedback. @gsantner

gsantner:
Jup!

Thank you very much for your time, it is merged now! 🎉

app/src/main/res/values/dimens.xml
guanglinn: I may have set the margin too small. 😬

PR #2119, opened 2023-10-09
Side margin improvements

guanglinn:
Adjust the side margins of editor/viewer.


guanglinn:
Hi, @gsantner. I think the changes to margins that I made before are a bit radical 😟, so I made some revert.

guanglinn:
## View mode

### Before [v2.10.9 (146)]


### Now

gsantner:
Thanks, I merge it again. Note that it basically reverts the previous one (in one dimension) while keeping the change on the other.

PR #2120, opened 2023-10-09
New file in file browser dialog

harshad1:
When sharing something into markor, it makes sense for the new file dialog and existing file dialogs to be combined.

This way you can select an existing file if it is already present or create a new one if it isn't.

I have also reduced the icon size and made them transparent so that they don't block information.

TBD
- [x] Should we immediately append to new file once created or show the new file in the browser to be selected
- [x] Should we show the editor after appending
- [x] Adjust icon size, position and transparency for best result


harshad1:
Screenshot_20231009-120112

How it currently looks

harshad1:
Forgot to add this earlier. The dialog is now simplified - no more extra button 

image

gsantner (approved):
Thanks

PR #2128, opened 2023-10-13
Add Orgmode Preiview Support

bigger124:
Hello, I have added the orgmode preivew support, Please merge it.

ac159f207cbc43e31fc714a7689562d

4cd3cfcb60085851085e110d17ecb85


gsantner:
@bigger124 
 there is no example showcase document like for all other formats.


Would you mind to add the one you show in screenshot? 
This is especially there so users can see what the format supports, what they can do with it and how they do it. 

Left side, no orgmode example, right side is the markdown example I created

a-append

Note: I don't use orgmode, and don't know all it's features and how it's written.

bigger124:
> @bigger124 there is no example showcase document like for all other formats.
> 
> Would you mind to add the one you show in screenshot? This is especially there so users can see what the format supports, what they can do with it and how they do it.
> 
> Left side, no orgmode example, right side is the markdown example I created
> 
> a-append
> 
> Note: I don't use orgmode, and don't know all it's features and how it's written.

@gsantner  I have added the orgmode  showcase file.

gsantner:
@bigger124 
Thank you for adding the sample document, thats surely helpful! 😄 

Question where did you get the org-bundle.js and the css file from? I see that you linked the org-js repository. But I can't find any bundle+css build like is commited here. I just want to make sure in the future that it can be updated, and that the hash matches between the project release and what we have here. This one [org.js](https://github.com/mooz/org-js/blob/master/org.js) at least seems to be different then ours.

bigger124:
Because org.js is a npm module, It can not be used directly in the html file, So I use [browserify](https://github.com/browserify/browserify)  to export the org-bundle.js. 

For example, `browserify  org.js > org-bundle.js`.

I added this CSS file myself to address the issue of the serial number and title being too close together

gsantner:
@bigger124 Thank you for the hint.

But when I do that browserify-ication I get different results.  [browserify-bundle-org.js.txt](https://github.com/gsantner/markor/files/13351916/browserify-bundle-org.js.txt)

Where did you download the source js file from and how did you use browserify on it? At the end I or anybody else need to follow the same steps to make sure we can update this later on when it's updateable.

bigger124:
@gsantner Okay, I will provide a detailed description of the packaging process
1. Install browserify
`npm install -g browserify`
2.Create a folder called **org-compile**
3.Create new file called **test.js** in **org-compile**, and put following content in to the file
`let org = require('org');
window.org = org;`
4.Install org, run command in **org-compile**, note the org-js version in npm  is 0.2.0
`npm install org`
5.Build, run command in **org-compile**
`browserify test.js > org-bundle.js`

bigger124:
@gsantner  I've resolved the conflict。

gsantner (approved):

gsantner:
Thank you, merged

PR #2125, opened 2023-10-12
BUGFIX: Using the correct format for correct types

harshad1:
This is a bugfix for #2106. Image and audio links added via the dialog were not being formatted correctly.

PR #2106, opened 2023-09-10
Reworked attachments

harshad1:
In this PR I have reworked the attachments system to fix bugs and reduce clicks.

- The attachment actions have been broken out of a nested menu into the main actions bar
- Camera, gallery and recording attachments add immediately when clicking ok. They can be edited later if needed.
- Camera permissions fixed
- User no longer given choice for where to save attachment
    - Setting include option for default attachment folder name
    - Need to consider making this per-file

PR #2132, opened 2023-10-17
Reworked attachments, nr3 - #2125  #2106 #2132

harshad1:
This is a second bugfix for the attachment PR 

PRs: #2125 #2106 #2132

### DIFF
--- a/app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
+++ b/app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
@@ -143,8 +143,6 @@ import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
-import java.net.URI;
-import java.net.URISyntaxException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -162,8 +160,6 @@ public class GsContextUtils {
     //## Constructor
     //########################
     public static final GsContextUtils instance = new GsContextUtils();
-
-    private static String _authority = null;
 
     public GsContextUtils() {
     }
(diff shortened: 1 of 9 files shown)


harshad1:
@gsantner This is ready to merge IMO

app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
```diff
@@ -1932,18 +1906,18 @@ public void onReceive(Context context, Intent intent) {
...
      * @param file File that should be edited
      */
-    public void requestPictureEdit(final Context context, final File file) {
-        final Uri uri = getUriByFileProviderAuthority(context, file);
+    public void requestFileEdit(final Context context, final File file) {
+        final Uri uri = FileProvider.getUriForFile(context, getFileProvider(context), file);
         final Intent intent = new Intent(Intent.ACTION_EDIT);
-        intent.setDataAndType(uri, "image/*");
-        intent.addFlags(Intent.FLAG_GRANT_WRITE_URI_PERMISSION | Intent.FLAG_GRANT_READ_URI_PERMISSION);
+        intent.setData(uri);  // Mime type inferred automatically
```
gsantner, line 1916: Sure that this always works as expected? While the original intent here was to edit pictures and thus image/ was specified, I think it makes sense to keep it this way to explicitly give additional info.

My idea is to use the Fileutils mimetype method here and keep using setDataAndType

app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
```diff
@@ -1932,18 +1906,18 @@ public void onReceive(Context context, Intent intent) {
...
+        intent.addFlags(Intent.FLAG_GRANT_WRITE_URI_PERMISSION);
+        intent.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
         intent.putExtra(MediaStore.EXTRA_OUTPUT, uri);
-        intent.putExtra(EXTRA_FILEPATH, file.getAbsolutePath());
```
gsantner, line 1945: Should be fine to keep sending this bit of info? For some software (like Markor 😄 ) it makes it easier for them to work if they have the full real path.

gsantner (changes requested):

app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
harshad1: Added it back in

app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
harshad1: It seems to work both ways. Using your approach now.

gsantner (approved):

PR #2139, opened 2023-10-21
WIP / EXPERIMENTAL: Interpret javascript in snippets

harshad1:
In this PR I have added support for the evaluation of javascript when expanding snippets.

Additionally the following keywords are supporteed:

- {{selection}}: replaced with currently selected text
- {{date}}: Inserts today's date in yyyy-MM-dd format
- {{time}}: Inserts the current time in HH:mm format
- {{title}}: Inserts current / new document title
- {{cursor}}: Cursor placed here after snippet insertion

I have had to include an external library and bump up the min version to 18 in order to enable this. I am investigating if we can achieve JS in snippets in other ways. We should probably bump up min version to 20 anyway as it is now 10 years old.


gsantner:
to be honest, not really a fan of adding JS on every end. and neither too much about increasing minSdk.

It might make sense over time to increase to sensible number but trying to keep it as low possibke.

harshad1:
While this branch is working well (and now does not require external dependencies or an increase in API level) I have decided that adding JS interpolation to markor is not something we should try to do after all.

Closing this for now. I will pull the other changes into a new branch + PR

PR #2144, opened 2023-11-05
Fix exceptions when audio is recorded

mehw:
Hello,

This is to **prevent an exception when aborting the audio recorder**.  Also, it's possible that a `break` is missing before the `REQUEST_SAF` case, **the application may crash asking for read permission after accepting the audio recorded**, but `copyUriToFile` should suffice to get a copy of the audio file.

Let me know if this is the right solution,
Thanks.

# Aborting audio recorder exception bug

1. Open a markdown file for editing;
2. Touch the **Insert Audio** button;
3. Touch the **Record Audio** button;
4. Touch the **Cancel** button to abort the recording;
5. Markor may crash due to an exception.

audio_exception

# Accepting audio recorded exception bug

Like above, but the audio is accepted this time.

1. Open a markdown file for editing;
2. Touch the **Insert Audio** button;
3. Touch the **Record Audio** button;
4. Touch the **Accept** button to confirm the recording;
5. Markor may crash due to a permission exception.

### DIFF
--- a/app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
+++ b/app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
@@ -1832,13 +1832,18 @@ public class GsContextUtils {
                 break;
             }
             case REQUEST_RECORD_AUDIO: {
-                final Uri uri = intent.getData();
-                final String uriPath = uri.getPath();
-                final String ext = uriPath.substring(uriPath.lastIndexOf("."));
-                final String datestr = new SimpleDateFormat("yyyy-MM-dd'T'HH-mm-ss", Locale.ENGLISH).format(new Date());
-                final File temp = new File(context.getCacheDir(), datestr + ext);
-                GsFileUtils.copyUriToFile(context, uri, temp);
-                sendLocalBroadcastWithStringExtra(context, REQUEST_RECORD_AUDIO + "", EXTRA_FILEPATH, temp.getAbsolutePath());
+                String audioPath = null;
+                if (resultCode == Activity.RESULT_OK && intent != null) {
+                    final Uri uri = intent.getData();
+                    final String uriPath = uri.getPath();
+                    final String ext = uriPath.substring(uriPath.lastIndexOf("."));
+                    final String datestr = new SimpleDateFormat("yyyy-MM-dd'T'HH-mm-ss", Locale.ENGLISH).format(new Date());
+                    final File temp = new File(context.getCacheDir(), datestr + ext);
+                    GsFileUtils.copyUriToFile(context, uri, temp);
+                    audioPath = temp.getAbsolutePath();
+                }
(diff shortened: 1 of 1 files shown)


app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
```diff
@@ -1832,13 +1832,18 @@ public Object extractResultFromActivityResult(final Activity context, final int
...
+                if (resultCode == Activity.RESULT_OK && intent != null) {
+                    final Uri uri = intent.getData();
+                    final String uriPath = uri.getPath();
+                    final String ext = uriPath.substring(uriPath.lastIndexOf("."));
+                    final String datestr = new SimpleDateFormat("yyyy-MM-dd'T'HH-mm-ss", Locale.ENGLISH).format(new Date());
+                    final File temp = new File(context.getCacheDir(), datestr + ext);
+                    GsFileUtils.copyUriToFile(context, uri, temp);
+                    audioPath = temp.getAbsolutePath();
+                }
+                sendLocalBroadcastWithStringExtra(context, REQUEST_RECORD_AUDIO + "", EXTRA_FILEPATH, audioPath);
```
harshad1, line 1845: Should this be inside the `if` ?

harshad1:
No the break should be outside (thanks for catching this bug)

I meant the call to `sendLocalBroadcastWithStringExtra`. Should it be inside ?

Is there a point in sending it if we didn't get a valid intent back?

Nov 6, 2023 07:16:29 Matthew White ***@***.***>:

mehw:
> I meant the call to `sendLocalBroadcastWithStringExtra`. Should it be inside ? Is there a point in sending it if we didn't get a valid intent back?

https://github.com/gsantner/markor/blob/7142e24f31329b9f106235e4ae8b57fc543d8a18/app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java#L1845

The above `sendLocalBroadcastWithStringExtra` is meant to **unregister the broadcast receiver, either when the audio recording is aborted or accepted**:

https://github.com/gsantner/markor/blob/1bb3a565684739cb0df13f6d0727f19a982281e0/app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java#L1896

Please, see also https://github.com/gsantner/markor/pull/2143 for further insight.

harshad1:
Got it. Thanks.

I am not a fan of the way we register and unregister the receiver. Need to take another whack at improving the attachment system in the near future.

mehw:
> I am not a fan of the way we register and unregister the receiver. Need to take another whack at improving the attachment system in the near future.

I'm sure that you will conceive a tidier solution ;)

Right now, I tried to solve the current logic to match each _registration_ with an _unregistration_ to the broadcast receiver.

harshad1:
Anyway, this lgtm

Thanks for cleaning up after me :)

mehw:
Thanks, good to know ;)

gsantner:
Thank you!

mehw:
@gsantner you're welcome ;)

PR #2145, opened 2023-11-06
Improvements to templating and link handling

harshad1:
#2139 but without the Javascript


app/src/main/java/net/gsantner/markor/format/markdown/MarkdownActionButtons.java
```diff
@@ -168,6 +183,34 @@ public boolean onActionLongClick(final @StringRes int action) {
         }
     }
 
+    private boolean followLinkUnderCursor() {
```
harshad1, line 186: I have implemented this for markdown alone

PR #2148, opened 2023-11-11
Attachment improvements / bugfix v4

harshad1:
Another set of improvements to attachments:

1. Fixed how we set titles for attachments, preventing user-entered names from being overwritten
2. Replaced broadcast with callback - makes it much easier to eliminate double inserts etc.


app/src/main/java/net/gsantner/markor/frontend/AttachLinkOrFileDialog.java
```diff
@@ -259,46 +258,56 @@ private static void insertItem(
...
+        if (dialog != null) {
+            nameEdit = dialog.findViewById(R.id.ui__select_path_dialog__name);
+            pathEdit = dialog.findViewById(R.id.ui__select_path_dialog__url);
+        } else {
+            nameEdit = pathEdit = null;
+        }
+
+        // Defensive checks to make sure file has not changed
+        // Can happen if the callback is triggered after a long delay
+        final long hash = GsFileUtils.crc32(text);
```
harshad1, line 272: We use hashes to make sure we aren't editing a file which may have changed

PR #2173, opened 2023-11-29
Add setting to disable the highlighted text menu shortcut

ConcurrentCrab:
Hello! thank you for the great app!

This is a PR that adds an option to disable the "open with" option shown on highlighted text in many apps like chromium browsers, because I personally find it quite annoying and I never use it. The default behavior of it is set to enabled, and as such it doesn't affect any existing or new installations.

Things I'm not sure about:
- Should the alias component name be extracted into a string resource? Android has weird rules about it, even if the build is marder and the packagename is `net.gsantner.markor_test`, the component name for the alias is still `net.gsantner.markor.AliasDocumentProcessText`.
- It works... mostly. Mostly instantly. Sometimes it doesn't work for reasons unclear to me, without any exceptions thrown, and you need to go into settings and toggle it back twice for the setting to take effect. I don't really get the android intent model, but I think this has more to do with Android caching the intent filter than how we're setting it. We might need to find a way to reliably bust the cache on this setting flip.


FriendlyNeighborhoodShane:
@harshad1 I too would agree that it's important to find a reliable mechanism for implementation of this toggle. A non-functioning setting is no good to anybody. That is why I put up the PR, to see if anybody had a better approach to this (especially since it's hard enough to find support for android that isn't either extremely outdated or useless SEO spam).

That said, I wouldn't agree with that assessment of it being yet another option adding clutter to settings. Because the way I see it, the option Markor shows everytime I highlight text is, to me, unnecessary visual clutter as well. And I'd rather that clutter live in the settings than there.

Of course I have to also recognise the fact that this is just my own (and possibly some other people's) opinions on this, and others may not share it. Such differences of opinions are what configuration options are meant to solve right :)

ConcurrentCrab:
@harshad1 I too would agree that it's important to find a reliable mechanism for implementation of this toggle. A non-functioning setting is no good to anybody. That is why I put up the PR, to see if anybody had a better approach to this (especially since it's hard enough to find support for android that isn't either extremely outdated or useless SEO spam).

That said, I wouldn't agree with that assessment of it being yet another option adding clutter to settings. Because the way I see it, the option Markor shows everytime I highlight text is, to me, unnecessary visual clutter as well. And I'd rather that clutter live in the settings than there.

Of course I have to also recognise the fact that this is just my own (and possibly some other people's) opinions on this, and others may not share it. Such differences of opinions are what configuration options are meant to solve right :)

gsantner:
> I'll be honest, I really don't see what the deal is about there being another option in settings.

Everything that is added, means that it is there. Requires maintainence, results in bug reports if it suddenly does behave different. Requires translations. Is bloat to users who don't care about it/don't touch it. Steers further away from keeping-it-simple-and-lightweight.

ConcurrentCrab:
Amended the suggested changes.

gsantner (approved):

gsantner:
Thank you, merged

ConcurrentCrab:
@gsantner Thank you!

Issue #1989, opened 2023-02-22
Add monochrome icon


xmha97:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **enhancement/feature request**. Not a [bug or question](https://github.com/gsantner/markor/issues/new/choose).
- [X] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [X] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [X] The wanted feature/enhancement is not present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor's settings are not touched. If your desired feature is present, you don't need to open this issue. The change will be part of the next Markor update.)

### Description

Please add the themed icon.

![Screenshot_2023-02-23-03-07-43-114_de mm20 launcher2 release](https://user-images.githubusercontent.com/23485114/220788329-91136b9a-92a9-403f-9ac4-23058913b255.jpg)

### Information

Android version: 13
Device: T11 (Xiaomi)
App Version: 

### Source

GitHub Releases

### Format / File type

Something else (-> Additional info)

### Additional info / Log

```shell
-
```

gsantner:
Are you sure the icons are not coming from a icon theme you have installed?

xmha97:
No, it's the "Themed Icons".
It's available only on Android 13.

Read more:
https://www.androidpolice.com/android-13-themed-icons-how-to-activate/

gsantner:
Feel free to make icon suggestions 😄 . Just please don't cripple the icon in such a way it's unnoticeable that it's even Markor and not a generic text icon.

xmha97:
I am not a designer but if you need idea, see this issue:
https://github.com/gsantner/markor/issues/142

gsantner:
> I am not a designer

Well, me neither 😄 .

>  Also I think the current icon needs redesign because it's not minimalist.

I disagree on that, I don't make something complete new just because it's trendy. It's ok to be traditional.

xmha97:
> > I am not a designer
> 
> Well, me neither 😄 .
> 
[...]

PR #2181, opened 2023-12-19
Material You Themed Icons Support

mhwearn:
Small addition to allow Markor to display as a Themed Icon as requested in https://github.com/gsantner/markor/issues/1989.


gsantner (approved):
thank you

PR #2176, opened 2023-12-04
Small features and improvements

harshad1:
This PR includes a bunch of small features and improvements:

1. Surrounding text with delimiters (for example markdown bold) happens in 1 step. This makes undo easier
2. Removing surrounding delimiters is also made slightly better (select text and press corresponding action)
3. Long press bold, italic, strikethrough to make the whole selected line strikethrough etc
    - Only for markdown 
5. The up and down arrow buttons now repeat. Can hold them to move selected line(s) up and down easily
6. There is an option in the file menu to show the currently open document in the file explorer
7. Can add multiple files to favourites in multi select mode
8. Added option to share path of currently open file
9. The select snippet directory option will now show the selected path
10. Reorganized how actions are specified in derived action button classes. Common actions are moved to ActionButtonBase.

PR #2187, opened 2023-12-26
Add options to select/unselect all files

guanglinn:
Add options to select/unselect all files, now do some file operations more conveniently in Markor.


app/src/main/res/values/strings.xml
```diff
@@ -63,6 +63,7 @@ work. If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
     <string name="last_modified">Last modified</string>
     <string name="select_this_folder">Select this folder</string>
     <string name="select">Select</string>
+    <string name="select_all">Select all</string>
```
harshad1, line 66: IMO the button title should change to 'clear' if all items are selected.

harshad1:
I added several related improvements in #2176 (currently open PR). These changes are
1. Refreshing the files list by swiping no longer clears the selected items.
2. Made several improvements to how we handle the selection - cleaner logic and flow etc.
3. I added a clear button to clear the selection.

This PR is more complete. So I think it makes sense for me to remove the clear button I added and instead use this. I will keep my underlying changes.

@guanglinn @gsantner I think this should be merged in first and then I will make the requisite changes to my PR

app/src/main/res/values/strings.xml
harshad1: Better yet we can do the following:

1. If no items selected make the text `Select all`
2. If one or more items selected make the text `Clear (n)` where n is the number of items selected

Then we don't need to modify the title to add a clear button etc. Changes become much simpler

guanglinn:
> I think this should be merged in first and then I will make the requisite changes to my PR.

I agree with you, doing so can be better to improve this feature.

app/src/main/java/net/gsantner/opoc/frontend/filebrowser/GsFileBrowserFragment.java
```diff
@@ -208,6 +216,23 @@ public void onFsViewerDoUiUpdate(GsFileBrowserListAdapter adapter) {
...
+                _topNavSubTitle.setVisibility(View.VISIBLE);
+            }
+            if (_topNavCancel.getVisibility() != View.VISIBLE) {
+                _topNavCancel.setVisibility(View.VISIBLE);
+            }
+        } else {
+            _topNavSubTitle.setText("");
+            _topNavSubTitle.setVisibility(View.GONE);
+            _topNavCancel.setVisibility(View.GONE);
+        }
```
gsantner, line 234: const bool isFileSelected = ...size() > 0;

if (isFileSelected) {
  _topNavSubTitle.setText(getResources().getString(R.string.selected) + " " + count;);
}
 _topNavSubTitle.setVisibility(isFileSelected? View.GONE : View.VISIBLE);
 _topNavCancel.setVisibility(isFileSelected? View.GONE : View.VISIBLE);

gsantner:
Thank you, I'm merging this.

As noted, "clear all" is something we want too, but harshad has some things in the works anyway about the selection. So lets make this simple one as first 😄

harshad1:
I have reworked this in my PR #2200 

Here is what it looks like now:

image

- The count is shown in the subtitle (this is built into android title bars - no need to modify title layout)
- The select all button is shown if any items selected but not all items selected
- The clear button is shown if any items selected

guanglinn:
Nice, I should try it.

PR #2190, opened 2023-12-27
A better way to show TOC

guanglinn:
Add a better way to show table of contents - show table of contents in a dialog instead of jumping to the table of contents at the front of the document.


guanglinn:
https://github.com/gsantner/markor/assets/60415467/77e2373f-6f95-440b-a9c2-cfe7e1cb25a7

harshad1:
In edit mode we show the toc / outline by clicking on the title. I think we should do the same thing here.

guanglinn:
Here has been an option to jump/show the toc in the bottom of view mode, so I don't think it is more appropriate to show the toc by clicking the title in view mode, and using this option don't cause conflicts -- show the toc in the dialog if `toc` is disabled, jump to the toc at the beginning if `toc` is enabled. These two ways can complement each other.

gsantner:
My feedback:

* I'm feeling a bit dangered by adding potentially unneccassary webview-java bridging. 
* Also by potentially unnecessary javascript/css. I try generally to keep external stuff lowest, so everything works with export of one html file with minimal css/js included.
* We have many ways to generate/access/jump in TOC already. Do we _really_ need another way?

So to say, sorry ahead if it doesn't make it in /  in the way how it is as of now.

gsantner (commented):
My feedback:

* I'm feeling a bit dangered by adding potentially unneccassary webview-java bridging. Making java callable directly from javascript can lead to unwanted exploiting. -> h1/h2 scanning etc can all be done on java side too. also dynamically (speaking about orgmode alike where html is generated in js)
* Also do we need potentially unnecessary/unwated javascript/css? I try generally to keep external stuff lowest, so everything works with export of one html file with minimal css/js included.
* We have many ways to generate/access/jump in TOC already. Do we _really_ need another way?

So to say, sorry ahead if it doesn't make it in /  in the way how it is as of now.  
Sorry I currently can't say exactly how good it is and what to change, but that I'm not quite happy about it.

guanglinn:
I think the old way of jumping to the toc always break the consistency of reading, so I think this new way may be more convenient. If others feel that the new way is necessary, I can try to improve and make it not dependent on JavaScript and CSS.

What do you think about this? @harshad1

guanglinn:
https://github.com/gsantner/markor/assets/60415467/3ea95655-d11a-4659-95d7-d8086b792400

guanglinn:
Hi, @gsantner , after my careful consideration, especially due to concerns about the risks you mentioned, I decide to close this pull request.

Perhaps in the future, I will have better ideas for improving Markor's TOC.

PR #2203, opened 2024-01-11
New Color picker, minsdk 16-18, closes #1428

### LINKED ISSUE #1428: Hex color for basic theme instead of color wheel
## General information
* App version: 2.8.0
* Android version:  11
* Device: Samsung A70
* I understand that this issue will be closed without further notice in case I didn't describe the problem including general information, or didn't [search](https://github.com/gsantner/markor/issues?q=) for similar issues & [requests](https://github.com/gsantner/markor/discussions?discussions_q=).

### Description

In the following I would like to change **BG Light** to `#fdf6e3`, how can I type it instead of choosing it from the colour wheel. May be you could provide an option to type it under the colour wheel.

Screenshot_20210906-173427_Markor

halfdane:
1. Introduces [martin stone's colorpicker](https://github.com/martin-stone/hsv-alpha-color-picker-android) as a replacement for quadflask's.
2. Removes the obsolete 3rd-party quadflask code 
3. Removes an unused method with a dependency to the quadflask code
4. Bumps the minSDK from 16 to 18

Since I'm not intimately familiar with this code (or any android code), please make sure I didn't inadvertently breack stuff with the removal of (probably?) unused code or the version bump.


gsantner:
Hello thank you for your help!

As you asking about breaking, these things are especially important to work: Color picking at editor (insert color i.e.) and at the settings (picking background color in basic scheme for example).

halfdane:
I have checked that the picker works correctly in basic color scheme settings, but that seems to be the only place where a color picker is used: I don't know how to do "Color picking at editor (insert color i.e.)", so if you could point me to the right place in the UI, I'll gladly check.

app/src/main/res/xml/preferences_master.xml
```diff
@@ -420,11 +419,6 @@
         </PreferenceCategory>
 
         <PreferenceCategory android:title="@string/text">
-            <other.com.flask.colorpicker.ColorPickerPreference
-                android:defaultValue="@color/default_link_color"
-                android:icon="@drawable/ic_format_underlined_black_24dp"
-                android:key="@string/pref_key__view_mode_link_color"
-                android:title="@string/link_color" />
```
gsantner, line 427: This settings option is currently missing in the updated code/new colorpicker.

gsantner (changes requested):

gsantner:
> I have checked that the picker works correctly in basic color scheme settings, but that seems to be the only place where a color picker is used: I don't know how to do "Color picking at editor (insert color i.e.)", so if you could point me to the right place in the UI, I'll gladly check.

There was a "Attachment" button and dialog, which currently seems to be removed or at last doesn't work. There were quite a few changes on that part recently and it's being reworked.

The pooint is, the new library support that - just showing a color picker dialog and get the color string out, so it can be inserted in the text document?

gsantner:
> I have checked that the picker works correctly in basic color scheme settings, but that seems to be the only place where a color picker is used: I don't know how to do "Color picking at editor (insert color i.e.)", so if you could point me to the right place in the UI, I'll gladly check.

There was a "Attachment" button and dialog, which currently seems to be removed or at last doesn't work. There were quite a few changes on that part recently and it's being reworked.

The point is, does the new library support showing a color picker dialog and get the color string out, so it can be inserted in the text document?

gsantner:
Want to merge this after the upcoming update, so it still goes by the old minsdk version and the big mass of changes reaches also users with old devices.

app/build.gradle
```diff
@@ -13,7 +13,7 @@ if (enable_plugin_kotlin) {
     apply plugin: 'kotlin-kapt'
 }
 
-rootProject.ext.version_minSdk = 16
+rootProject.ext.version_minSdk = 18
```
harshad1, line 16: There are a bunch of

` if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) { `

in our codebase. May be good to remove them if we are bumping versions

app/src/main/res/xml/preferences_master.xml
gsantner: I restored it self, pushed to PR branch

gsantner (approved):
Thank you, it's time to drop the dated color picker. Merging

PR #2205, opened 2024-01-12
WIP: Fixes #2182 with a more compact display of file browser items:

### LINKED ISSUE #2182: File browser: Improve readability of file browser by displaying file description less prominent

### Description

I have a lot of folders and files in my Notebook, and I find navigation a bit hard because the file description is displayed so prominently. 99% of the time I'm only interested in the file name. Ideas to improve readability:

- Add an option to right align the file description
- Make the file description use less vertical space. Right now it looks like file name and file description share an equal amount of vertical space (including the padding), although the file name uses a slightly bigger font and is more bright and bold.
- Option to adjust the color of the file description (to be able to lower the contrast to background color).

**Semi-Workaround**:

Manual indentation using spaces:

markor

I have looked at the `SimpleDateFormat` docs but it looks like alignment cannot be set from within the date and time pattern strings (which is probably good).

### Information

Android version: 14 (GrapheneOS)
Device: Pixel 8
App Version: 2.11.1 (F-Droid says 2.11.1 but it looks like the latest release is only 2.11.0?)

### Source

F-Droid

### Format / File type

Not specific

### Additional info / Log

_No response_

halfdane:
- Introduces a preference to enable "compact view" of file browser items
- if preference is enabled, filename an description are layed out horizontally instead of vertically


harshad1:
Could you add a before and after screenshot here?

app/src/main/res/layout/opoc_filesystem_item.xml
```diff
@@ -46,6 +48,10 @@
             android:textSize="19sp"
             tools:text="how to write a readme.md" />
 
+        <Space
+            android:layout_width="8dp"
+            android:layout_height="wrap_content" />
+
```
gsantner, line 54: Makes default vertical layout 8dp bigger by adding this spacer. If it's needed for horiziontal, hide it in vertical please

app/src/main/res/layout/opoc_filesystem_item.xml
```diff
@@ -46,6 +48,10 @@
             android:textSize="19sp"
             tools:text="how to write a readme.md" />
 
+        <Space
+            android:layout_width="8dp"
+            android:layout_height="wrap_content" />
```
gsantner, line 53: Makes default vertical layout 8dp bigger by adding this spacer. If it's needed for horiziontal, hide it in vertical please

gsantner (changes requested):

app/src/main/res/layout/opoc_filesystem_item.xml
harshad1: To add to this. 

IMO if we can come up with a slightly more compact clean layout which doesn't compromise anything we should just switch to that as the new default.

gsantner:
Sorry to say, but I don't want to merge/implement this. I know some people would like every tiny bit finetuned to their personal taste, but I want to add to-my-feeling unnecessary things.

There is a certain size that works fine for fingers/thumbs (""minimum viable button touch width/height""), which shouldn't decrease too much then it's currently. If you find additional ways to improve it, feel free to. But it should be a general improvement that fits without a settings option (and not slams everything into one line). Like adjusting margins or so.

halfdane:
I'm fine keeping the layout like it is - I just saw this open bug and thought I'd give it a try 🙂

Issue #154, opened 2017-12-10
Keyboard shortcuts for editor (HW)


gsantner:
#### General information

* **App version:** 0.2.5
* **System:** 

#### Description
Add keybindings like Ctrl-Z, Ctrl-Y, Ctrl-A, Ctrl-S

https://stackoverflow.com/questions/12250563/custom-keyboard-shortcuts

gsantner:
The basics for this implemented, just calls to the methods need to happen ;)

conrad-heimbold:
CTRL+V (Paste), 
CTRL+C (Copy), 
CTRL+X (Cut), 
CTRL+Z (undo), 
CTRL+A (mark all) 
... all seem to work. 

Only CTRL+S (save) is missing.

gsantner:
CTRL+S can be mapped to the save method. Estimated this is like 10 lines. Contributions welcome!

gsantner:
Long pressing the 'Special key' button will jump to bottom/top of document in next release.

sojusnik:
Shortcuts for Markdown formatting would be very useful too:

H1, H2, H3: Ctrl + 1-3
Italic: Ctrl + i
Bold: Ctrl + b
Quote: Ctrl + q
List & Numbered List: Ctrl + l

If the text cursor is at the beginning, end or in the middle of a word, then pressing the above mentioned shortcut should format the word correspondingly.

mariellep:
Ability to set some of the existing shortcuts to a Bluetooth keyboard mapping would be useful. For example, indent/de-indent requires hitting the keyboard button in the editor and scrolling down as opposed to just hitting tab on a regular computer editor. Perhaps ctrl+tab for indent and ctrl+shift+tab for de-indent?

gsantner:
Yes thats among what I thought too. But should work on any keyboard, shouldnt matter that its BT

aisbergde:
I have no idea how to use some "Ctrl + ..." combination on android. I have no Ctrl on any of my keyboards.
What I mean is one more entry in this list, to be able to mark a line:

![Screenshot_2020-05-24-01-22-07-470_net gsantner markor](https://user-images.githubusercontent.com/9020187/82742357-531ab500-9d5d-11ea-86f7-20d22c29d19c.jpg)

gsantner:
Thanks now I know where you mean. Is quite easy to add, so when somebody reads this: for functionality it's two invocations, plus if+string.

gsantner:
[...]

PR #2207, opened 2024-01-24
Add several basic keyboard shortcuts

guanglinn:
Add several essential physical keyboard shortcuts, such as using `Ctrl+/` to toggle preview mode. These shortcuts are very functional, in particular Android tablets connected with physical keyboards.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
@@ -342,19 +342,11 @@ public class DocumentActivity extends MarkorBaseActivity {
 
     @Override
     public boolean onKeyDown(int keyCode, KeyEvent event) {
-        if (event.isCtrlPressed() && keyCode == KeyEvent.KEYCODE_S) {
-            GsFragmentBase fragment = getCurrentVisibleFragment();
-            if (fragment instanceof DocumentEditAndViewFragment && fragment != null) {
-                ((DocumentEditAndViewFragment) fragment).saveDocument(true);
-            }
-        }
-        return super.onKeyDown(keyCode, event);
+        return super.onReceiveKeyPress(getCurrentVisibleFragment(), keyCode, event) ? true : super.onKeyDown(keyCode, event);
     }
 
     public GsFragmentBase showFragment(GsFragmentBase fragment) {
-
         if (fragment != getCurrentVisibleFragment()) {
-
             _fragManager.beginTransaction()
                     .replace(R.id.document__placeholder_fragment, fragment, fragment.getFragmentTag())
                     .commit();
@@ -365,9 +357,7 @@ public class DocumentActivity extends MarkorBaseActivity {
     }
 
     public synchronized GsFragmentBase getExistingFragment(final String fragmentTag) {
-        FragmentManager fmgr = getSupportFragmentManager();
-        GsFragmentBase fragment = (GsFragmentBase) fmgr.findFragmentByTag(fragmentTag);
-        return fragment;
+        return (GsFragmentBase) getSupportFragmentManager().findFragmentByTag(fragmentTag);
     }
 
     private GsFragmentBase getCurrentVisibleFragment() {
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
@@ -16,6 +16,7 @@ import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
 import android.text.Html;
+import android.view.KeyEvent;
 import android.view.MotionEvent;
 import android.widget.TextView;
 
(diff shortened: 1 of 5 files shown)


harshad1:
I propose a more general approach:

1. Create an empty onReceiveKeyPress in MarkorBaseFragment
2. Receive the onReceiveKeyPress in MarkorBaseActivity and pass it along to every active child fragment
3. Override and act on Ctrl-C in DocumentEditFragment

- No redundant code
- Adds nice functionality

guanglinn:
I improved the code and added several functional shortcuts, such as undo, redo, bold, italic and toggle preview mode. They are basic keyboard shortcut functions.

app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
```diff
@@ -339,10 +340,13 @@ public void onBackPressed() {
         }
     }
 
-    public GsFragmentBase showFragment(GsFragmentBase fragment) {
+    @Override
+    public boolean onKeyDown(int keyCode, KeyEvent event) {
```
harshad1, line 344: Just move this into base class IMO.

onKeyDown -> pass to currentVisibleFragment() if it is not null

i.e instead of DocumentActivity -> BaseActivitiy -> DocumentEditFragment
just
BaseActivity (receive and forward) -> BaseFragment function -> Override in DocumentEditFragment

app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
```diff
@@ -353,9 +357,7 @@ public GsFragmentBase showFragment(GsFragmentBase fragment) {
     }
 
     public synchronized GsFragmentBase getExistingFragment(final String fragmentTag) {
-        FragmentManager fmgr = getSupportFragmentManager();
-        GsFragmentBase fragment = (GsFragmentBase) fmgr.findFragmentByTag(fragmentTag);
-        return fragment;
+        return (GsFragmentBase) getSupportFragmentManager().findFragmentByTag(fragmentTag);
```
harshad1, line 360: Thank you for also doing these cleanups as you make your PRs

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
```diff
@@ -357,6 +359,39 @@ public boolean onQueryTextChange(String text) {
...
+    public boolean onReceiveKeyPress(int keyCode, KeyEvent event) {
+        if (event.isCtrlPressed()) {
+            if (event.isShiftPressed() && keyCode == KeyEvent.KEYCODE_Z) {
+                if (_editTextUndoRedoHelper != null && _editTextUndoRedoHelper.getCanRedo()) {
+                    _hlEditor.withAutoFormatDisabled(_editTextUndoRedoHelper::redo);
+                    updateUndoRedoIconStates();
+                }
+                return true;
+            } else if (keyCode == KeyEvent.KEYCODE_B) {
+                _format.getActions().onActionClick(R.string.abid_markdown_bold);
```
harshad1, line 372: This will not work with other formats. IMO we should do one of the following:

Either
Option 1: Rename all format specific but actually common action strings to common values
i.e. `abid_markdown_bold`, `abid_wikitext_bold` etc all to `abid_bold`

Or

Option 2: Create actionBold(), actionItalics() etc in ActionButtonBase and override them in each type specific <Format>ActionButtons

guanglinn:
I will try to make some improvements, but it looks like there's a lot of work to be done.

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
guanglinn: I think the option 2 may be better.

guanglinn:
I have removed those shortcuts for specific formats. Perhaps, we can merge current code first, and add shortcuts for specific formats later.

app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
guanglinn: But in `BaseActivity`, there is no method `getCurrentVisibleFragment()` or `getPosFragment(getCurrentPos())`, we cannot get the fragment we need.

gsantner (approved):
Thanks! Better hardware keyboard and shortcut support is for sure something important not touched too much so far. Of course we improve it further.

harshad1:
@guanglinn I think I have an approach which can help us cleanly add format-specific keyboard shortcuts:

1. Handle Copy, Paste etc in DocumentEditAndViewFragment as done now
2. If the keypress is not handled, pass it into ActionButtonBase 
3. Let each format override the function in ActionButtonBase as desired
4. Let people interested in maintaining various formats add functionality for various formats

What do you think?

@gsantner Maybe we should create a discussion for contributors so we can share things like this.

gsantner:
There is issue #154 which I assume has a few subscribers already as it's one of the longest open topics :D

guanglinn:
@harshad1 I think we can add a method in `ActionButtonBase`, that method receives `keyCode` and `keyEvent` as parameters, and some default and common keyboard shortcuts should be implemented here.
Then we call back that method in `onReceiveKeyPress(keyCode, keyEvent)` of `DocumentEditAndViewFragment` or other class that required to be listened.

If we implement the above code. In the future, we just need to override that base class method in the format-specific subclass, for example, `MarkdownActionButton` to custom keyboard shortcuts for the specific format.

FEI352:
期待添加这个PR Markor，有了Ctrl+S 保存，可以代替安卓端的Obsidian了。

PR #2211, opened 2024-02-05
Add support for restoring scroll position of file lists

guanglinn:
This PR mainly implemented the following features for file browser:

1. Highlight recently opened files or folders;
2. Can restore the scroll position for the upper level list.

The screen recording as follows:

https://github.com/gsantner/markor/assets/60415467/ca3664f5-146a-48b9-ace8-ba15b527b331/


guanglinn:
Highlight item color in dark theme:

harshad1:
We already have logic to flash a view and this is used in other places.

https://github.com/gsantner/markor/blob/60d1f434a7fe94ba93fa357754636a8016880bb1/app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java#L2853

See how it is used here:

https://github.com/gsantner/markor/blob/master/app/src/main/java/net/gsantner/opoc/frontend/filebrowser/GsFileBrowserListAdapter.java#L550

We already have a function to show and blink a file in the folder view. Just need to call it at the appropriate time.

I have added an option to show the currently open file in the file explorer in this PR:

https://github.com/gsantner/markor/pull/2176/files#diff-a5f18ed8d9a7afd47e78969d79f34ea509f949c4f383113680d828316380943dR607-R610

An option has been added to the file menu to show the file in the file explorer. This works even if the file was launched from search or a shortcut.

This PR may not be necessary

guanglinn:
@harshad1 , Does your PR have the feature to restore scroll position of file list? If not, I can only keep this feature.

guanglinn:
In addition, I hope the feature in your PR can highlight recently opened folders, rather than only documents.

app/src/main/java/net/gsantner/opoc/frontend/filebrowser/GsFileBrowserListAdapter.java
```diff
@@ -716,6 +662,147 @@ public static boolean isVirtualStorage(File file) {
                 VIRTUAL_STORAGE_RECENTS.equals(file);
     }
 
+    private final static Object LOAD_FOLDER_SYNC_OBJECT = new Object();
```
harshad1, line 665: Note that I have also simplified / cleaned up this function in my PR.

app/src/main/java/net/gsantner/opoc/frontend/filebrowser/GsFileBrowserListAdapter.java
```diff
@@ -566,116 +617,11 @@ public int getFilePosition(final File file) {
...
-                            notifyDataSetChanged();
-                            if (_dopt.listener != null) {
-                                _dopt.listener.onFsViewerDoUiUpdate(GsFileBrowserListAdapter.this);
-                            }
-                        });
-                    }
-                }
-            }
-        }.start();
+        executorService.execute(new FolderLoader(folder));
```
harshad1, line 624: Interesting. What are the advantages of this vs just new Thread()?

harshad1:
No. All I have is the ability to jump from an open file to it's location in the folder.

I also like how the highlighting to show the file / folder looks. There are some good changes in this PR. I will think about this more.

guanglinn:
I'm still not sure which code should be kept in my PR. Perhaps, I think I can decide it after your PR #2176 is merged.

app/src/main/java/net/gsantner/opoc/frontend/filebrowser/GsFileBrowserListAdapter.java
guanglinn: When a user open folders (or other cases of loading folders) several times within a short period of time (such as 60s), this will frequently create many new threads by `new Thread()`, which may not be as efficient as using a thread pool, a thread pool can reuse existing idle threads. When a user no longer loads folders for a period of time (60s), these threads will automatically shutdown. Anyway, I think this can improve performance of the file browser.

> Thread lifecycle overhead. Thread creation and teardown are not free. The actual overhead varies across platforms, but thread creation takes time, introducing latency into request processing, and requires some processing activity by the JVM and OS. If requests are frequent and lightweight, as in most server applications, creating a new thread for each request can consume significant computing resources.

References:

[1] Brian Göetz et al. [Java Concurrency In Practice](https://library-it.com/wp-content/uploads/2020/06/java_concurrency_in_practice.pdf). Disadvantages of Unbounded Thread Creation. 2006;6.1.3:89-90.

[2] [Java - Difference between Thread and Threadpool - Stack Overflow](https://stackoverflow.com/questions/6592976/difference-between-thread-and-threadpool)

gsantner:
PR has conflicts. might be easier to take version from msater and just make the few line changes again

guanglinn:
@gsantner . Conflicts has been resolved.

app/src/main/AndroidManifest.xml
```diff
@@ -63,7 +63,7 @@
             android:name=".activity.MainActivity"
             android:exported="true"
             android:label="@string/app_name"
-            android:launchMode="singleTop"
+            android:launchMode="singleInstance"
```
harshad1, line 66: Why is this change being made? We do not want multiple instances of MainActivity floating around.

app/src/main/java/net/gsantner/opoc/frontend/filebrowser/GsFileBrowserListAdapter.java
```diff
@@ -771,7 +769,8 @@ public void run() {
                             final View view = _recyclerView.getChildAt(i);
                             final TextView textView = view.findViewById(R.id.opoc_filesystem_item__title);
                             if (data.file.getName().equals(textView.getText().toString())) {
```
harshad1, line 771: I already have code to blink any view. I think you should just use that. If necessary, modify it change the background in this way.

app/src/main/java/net/gsantner/markor/format/asciidoc/AsciidocActionButtons.java
```diff
@@ -29,8 +29,8 @@ public AsciidocActionButtons(@NonNull Context context, Document document) {
     }
 
     @Override
-    public List<ActionItem> getActiveActionList() {
```
harshad1, line 32: I think the diff is broken. Please rebase and re-create this PR

app/src/main/AndroidManifest.xml
guanglinn: I didn't change it, this change from PR #2176.

app/src/main/java/net/gsantner/opoc/frontend/filebrowser/GsFileBrowserListAdapter.java
guanglinn: I think it may be unnecessary to use that complex animation here.

gsantner (approved):
@guanglinn  @harshad1 

Many merge requests - regarding blinking and showing position in the list: Let's bring it all together on master now, and further improve/debloat it afterwards.

When dark theme is used, I find the white blink color bad.

app/src/main/java/net/gsantner/markor/format/asciidoc/AsciidocActionButtons.java
gsantner: It was indeed, showed 77 changed files now 3 changed files

guanglinn:
> When dark theme is used, I find the white blink color bad.

I will improve it.

PR #2213, opened 2024-02-07
Enable fast scroll for file browser

guanglinn:
Enable fast scroll for RecyclerView, the scrollbar will also be revealed in the file browser.


guanglinn:
https://github.com/gsantner/markor/assets/60415467/1d8dded3-7f3f-485c-b6e7-d8b94f742062

harshad1:
So we have these classes already:

https://github.com/gsantner/markor/blob/master/app/src/main/java/net/gsantner/markor/web/DraggableScrollbarWebView.java

Perhaps we should just repurpose this logic?

guanglinn:
Thank you for your prompt. These codes are very useful and I should reuse them. I may need to override RecyclerView.

guanglinn:
But the RecyclerView already has built-in fast scroll. Is it necessary to override this class, will it make the code redundant?

harshad1:
Ok. Just wanted to show you that code in case it makes sense to reuse it. If it does not then it does not

guanglinn:
Thanks for your guidance, which lets me have a better understanding of this project's code.

app/src/main/res/layout/opoc_filesystem_fragment.xml
```diff
@@ -39,8 +39,12 @@
             android:layout_width="match_parent"
             android:layout_height="match_parent"
             android:minHeight="100dp"
+            app:fastScrollEnabled="true"
+            app:fastScrollHorizontalThumbDrawable="@drawable/scrollbar_thumb"
+            app:fastScrollHorizontalTrackDrawable="@drawable/scrollbar_track"
```
gsantner, line 44: I don't any horizontal scrolling is needed? Especially also on the main screen there is horizontal swiping which contradicts horizontal scrolling in subviews.

gsantner (changes requested):

app/src/main/res/layout/opoc_filesystem_fragment.xml
guanglinn: I noticed that if we added the `app:fastScrollEnabled="true"`, then the following 4 related Drawable attributes are all indispensable, even the horizontal scrollbar will not be revealed forever,
`app:fastScrollHorizontalThumbDrawable`;
`app:fastScrollHorizontalTrackDrawable`;
`app:fastScrollVerticalThumbDrawable`;
`app:fastScrollVerticalTrackDrawable`,
otherwise this APP will crash when launched.

guanglinn: So we cannot delete `app:fastScrollHorizontalThumbDrawable` and `app:fastScrollHorizontalTrackDrawable`.

guanglinn:
Hi, @gsantner . Do you prefer hex color values in lowercase or uppercase.

gsantner:
I prefer uppercase for hexdecimal. But there is no need to refactor any predefined colors when they are not upper.

guanglinn:
This PR is ready to be merged now.

app/src/main/res/layout/opoc_filesystem_fragment.xml
guanglinn: @gsantner , I just tried it, there is no contradicts in horizontal swiping and horizontal scrolling in subviews.

gsantner (approved):
Thanks, the scroll track works as expected.

Though - @harshad1  @guanglinn : the scrolling feel s very sluggish now. There were a lot of changes to the scrolling including i.e. threadexecutor, which might be an delay/sync issue. It feels quite bad 😄

gsantner:
@guanglinn 
the scrolling and filebrowser feel very sluggish now. If I didn't know any better, I would say there is some while(true) in the UI thread running.

Also another finding, the file/folder blinks when browsing files. I.e. go to subfolder and up again, and the file is highlighted again and again.

guanglinn:
@gsantner 
It works well on my device. Could you add a screen recording here.

I will try to resolve these issues as soon as possible.

gsantner:
* The total sluggish is not always the case, sometimes it's really slow and also the tab switcher (bottom bar) takes a bit to switch around.
* Upon new creation, there is a short blink of it, and then the text stays unreadable
* Also see the blink often running very quick, so it blinks like 15 times quickly in a second

Some impressions

https://github.com/gsantner/markor/assets/6735650/8d7a2f87-d89c-49f4-9319-3243a7f64298

https://github.com/gsantner/markor/assets/6735650/8161ab58-211a-4e7f-b917-de7c5125d133

gsantner:
Thank you very much for implementing (and wanting to improve it)

PR #2238, opened 2024-03-07
Add action: 'duplicate lines'

Goli4thus:
Hello!

First off, thanks for creating `Markor`!

This pull request adds a new action: "duplicate lines".

### Why was it added?
More than just a couple times I noticed myself wanting to duplicate the current line and making a slight variation to it. Doing so via selecting the line, copying it, moving to new line and pasting it is just too tedious when needed often.

I've searched a bit regarding issues and discussions.
The closest I could find was this: #2012
But even the proposed answer felt too tedious, cause it still requires multiple steps.

### What exactly does the action do?
Based on the currently focused line(s), once the action is triggered, the selected lines (even if only partially selected) will be duplicated downwards.
The selection follows the newly created lines (e.g. for repeated duplication).
Cursor position is maintained.

I very much based this on the existing "move lines down" action.
I've done a bunch of manual testing regarding various scenarios:
- single line
- multiple lines partially selected
- single empty line
- multiple lines, including empty line (start, middle, end of "selection block")

Overall it looked pretty solid by the end.
And I couldn't find any automated tests regarding the "move lines up / down" actions. So I figured this could be enough for testing. 

The icon I created from scratch. If the design is too shabby, feel free to say so. ;)


app/src/main/java/net/gsantner/markor/format/ActionButtonBase.java
```diff
@@ -873,6 +880,37 @@ public static void moveLineSelectionBy1(final HighlightingEditor hlEditor, final
         }
     }
 
+    public static void duplicateLineSelection(final HighlightingEditor hlEditor) {
+        // Duplication is performed downwards, selection is moving alongside it.
+        final Editable text = hlEditor.getText();
+
+        final int[] sel = TextViewUtils.getSelection(hlEditor);
+        final int linesStart = TextViewUtils.getLineStart(text, sel[0]);
+        final int linesEnd = TextViewUtils.getLineEnd(text, sel[1]);
```
harshad1, line 889: look at TextViewUtils.getLineSelection and TextViewUtils.getSelectedLines

app/src/main/java/net/gsantner/markor/format/ActionButtonBase.java
```diff
@@ -873,6 +880,37 @@ public static void moveLineSelectionBy1(final HighlightingEditor hlEditor, final
...
+        hlEditor.withAutoFormatDisabled(() -> {
+            final String lines_final = String.format("%s\n", lines);
+            text.insert(altStart, lines_final);
+        });
+
+        final int sel_offset = selEnd[0] - selStart[0] + 1;
+        selStart[0] += sel_offset;
+        selEnd[0] += sel_offset;
+
+        hlEditor.setSelection(
```
harshad1, line 909: I don't think most of this is necessary. We only need to do this if we are modifying the selected lines. But we are not doing that here.

harshad1: I think this function can be reduced to something like

```
final int sel[] = TextViewUtils.getLineSelection(text);
final CharSequence lines = text.subSequence(sel[0], sel[1]);
hlEditor.withAutoFormatDisabled(() -> text.insert(sel[1] + 1, lines)
```

Please test :)

harshad1:
FWIW I like this :)

app/src/main/java/net/gsantner/markor/format/ActionButtonBase.java
Goli4thus: Thanks for the quick feedback!
I just tried your suggestion, but there are a couple problems with it:
1) it is missing a newline at the end of the duplicated line(s)
2) it doesn't perserve the cursor column position

Please see later comment for more details on what I mean by point 2.

app/src/main/java/net/gsantner/markor/format/ActionButtonBase.java
Goli4thus: Please see later comment on why I still think I need to use this existing approach.

Goli4thus:
I have pushed another commit which removes some needless code.
Still, most remained for now.

Furthermore, I likely didn't explain clear enough what I was going for with this action.
So here we go:

One concrete usecase of mine would be to:
1. Manually position the cursor in an existing line at the position I'd wanted to make a change (i.e. for creating a variation of that very line)
2. Duplicate the line with this new action
3. Apply the change right away at the initially selected column position (be it deleting some text or add some more text)

If the selection wouldn't move with the duplication, then the cursor would not end up in the duplicated line and would
not allow for step 3 above.
But I'd argue that's most likely the intended line for applying a change (at least for me that would be the case the vast majority of the time).
I simply would like to save on manually having to position the cursor after the duplication action myself.


The above was just about a single line scenario.
When it comes to a multi-line selection, they way my current implementation behaves is simply the consequence of how the
single line scenario should behave.

I could agree that it looks bit funny at first for the selection to be moving along.
But for the sake of simplicity, I wouldn't make a distinction for "single vs. multi" line selections when it comes to
where the cursor / selection is located at the end of triggering this action.

At least the current implementation works robustly from what I can see and fulfills the above requirement regarding
cursor positioning.


Is it now clearer what my intention behind this action is?
Do your still think there are more things to optimize at this point?

harshad1:
Ah you want to keep the 'current' selection but move it to the duplicated lines. Ok, that makes sense.

gsantner:
V y p 😄

gsantner (approved):
Thanks! Merging.

Note: with a empty file or only one line of text it seemed to not work, but after adding a second line and reopening the file it worked. Something we can fix later though.

Goli4thus:
> Thanks! Merging.
> 
> Note: with a empty file or only one line of text it seemed to not work, but after adding a second line and reopening the file it worked. Something we can fix later though.

Thanks for the merge and also thanks for doing some additional testing!
I just created a bugfix PR here: #2251

PR #2259, opened 2024-04-05
Shortcut to special files

harshad1:
This PR adds the ability to create shortcuts to special files (Favourite etc on the home screen).

They are created as usual, like any other folder - long press and create shortcut


app/src/main/java/net/gsantner/markor/util/MarkorContextUtils.java
```diff
@@ -46,13 +47,27 @@ public <T extends GsContextUtils> T applySpecialLaunchersVisibility(final Contex
         return thisp();
     }
 
+    private static int getIconResForFile(final File file) {
```
gsantner, line 50: suggestion: add @ nullable or nonnull, if it's nullable then handle if (null) in first condition

app/src/main/java/net/gsantner/opoc/frontend/filebrowser/GsFileBrowserFragment.java
```diff
@@ -37,6 +37,7 @@
 import androidx.recyclerview.widget.LinearLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
 import androidx.swiperefreshlayout.widget.SwipeRefreshLayout;
+import androidx.appcompat.widget.Toolbar;
```
gsantner, line 40: unneeded import (automatic cleanup possible)

gsantner (commented):
thanks

app/src/main/java/net/gsantner/markor/util/MarkorContextUtils.java
harshad1: Fixed

app/src/main/java/net/gsantner/opoc/frontend/filebrowser/GsFileBrowserFragment.java
harshad1: Fixed

gsantner (approved):
thanks, works!

cinnabar777:
When using this feature the home screen shortcut opens a new window every time it's trapped instead of bringing up the already open window. 

I think it's due to having multiple windows option enabled in settings? 

Nonetheless the shortcut does open the file and in the last used state, but will open news window every time.

PR #2265, opened 2024-04-18
Changing to correct action ids for attachments

### LINKED ISSUE #2253: Attachment button does not react
# IMPORTANT
* App version: 2.11.1
* Android version/rom: GrapheneOS
* Device: pixel6a

## Explain the Problem
The clip button in plaintext files or todo list does not show any reaction at all. No error message is displayed.
All access rights to file system have been granted. 

## Steps to Reproduce
Open plain text file, hit clip button.

harshad1:
In the last round of attachment improvements, several formats were left with old action keys. This fixes that. Attachment button on todotxt etc will now work.

This fixes #2253

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/orgmode/OrgmodeActionButtons.java
+++ b/app/src/main/java/net/gsantner/markor/format/orgmode/OrgmodeActionButtons.java
@@ -26,9 +26,11 @@ public class OrgmodeActionButtons extends ActionButtonBase {
                 new ActionItem(R.string.abid_common_checkbox_list, R.drawable.ic_check_box_black_24dp, R.string.check_list),
                 new ActionItem(R.string.abid_common_unordered_list_char, R.drawable.ic_list_black_24dp, R.string.unordered_list),
                 new ActionItem(R.string.abid_common_ordered_list_number, R.drawable.ic_format_list_numbered_black_24dp, R.string.ordered_list),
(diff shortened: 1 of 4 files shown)


harshad1:
This now can add attachments to the todotxt and plain text. But at least for todotxt we do not render or have a way to open the link. This will need more development

app/src/main/java/net/gsantner/markor/format/orgmode/OrgmodeActionButtons.java
```diff
@@ -26,9 +26,11 @@ public List<ActionItem> getFormatActionList() {
                 new ActionItem(R.string.abid_common_checkbox_list, R.drawable.ic_check_box_black_24dp, R.string.check_list),
                 new ActionItem(R.string.abid_common_unordered_list_char, R.drawable.ic_list_black_24dp, R.string.unordered_list),
                 new ActionItem(R.string.abid_common_ordered_list_number, R.drawable.ic_format_list_numbered_black_24dp, R.string.ordered_list),
-                new ActionItem(R.string.abid_common_attach_something, R.drawable.ic_attach_file_black_24dp, R.string.attach),
```
gsantner, line 29: The dialog also used to hold a button to insert a color (html foreground, html background or as simple #hexstring). That ones seems to have gone completely down under

gsantner:
> But at least for todotxt we do not render or have a way to open the link. This will need more development

We make markdown style link on press and allow unformatted https links also as usual, the "open browser" button anyway works outside of syntax usually.

gsantner:
Thanks merged

PR #2280, opened 2024-05-13
WIP EXPERIMENTAL: File access in javascript

harshad1:
In this PR I am experimenting with adding a 'Markor' api to javascript opened in the webview. 

This API will allow reading / writing / navigating the file system from javascript in a webview.

May enable useful extensions to markor by users.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/web/MarkorWebViewClient.java
+++ b/app/src/main/java/net/gsantner/markor/web/MarkorWebViewClient.java
@@ -9,11 +9,14 @@ package net.gsantner.markor.web;
 
 import android.app.Activity;
 import android.content.Context;
+import android.graphics.Bitmap;
+import android.os.Build;
 import android.webkit.WebView;
 
 import net.gsantner.markor.ApplicationObject;
 import net.gsantner.markor.activity.DocumentActivity;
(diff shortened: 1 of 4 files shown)


harshad1:
Not sure if we want to merge this at all. Putting it here to drive discussion.

gsantner:
We should not expose file management features to the web/webview, risky to get misused. Some things might be interesting but it needs to be very very careful if it needs to be.

app/src/main/java/net/gsantner/markor/web/MarkorWebViewClient.java
```diff
@@ -62,4 +79,20 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
         }
         return true;
     }
+
+    @Override
+    public void onPageStarted(final WebView view, final String url, final Bitmap favicon) {
```
harshad1, line 84: For some added safety I added this to disable file access for non-local files.

Most of this PR is to be looked at as an experiment, not to merged anytime soon :)

harshad1:
Closing this PR. I want to explore other avenues for enabling more functionality from JS

gsantner:
As usual idea and improvements are welcome. Only need to make sure it not obviously poses a viable threat

PR #2282, opened 2024-05-14
WIP: BUGFIX: Rework to include shared text in shared links

harshad1:
This addresses the issue brought up in https://github.com/gsantner/markor/discussions/2281

We will now have the link text properly included in the link.

Lines in shared text will be formatted as links iff 
1. They are a file path
2. They end with a web link. In this case the part before the web link will be used as the link text


gsantner:
Thank you, seems to be working good enough as first hotfix. (can be further improved of course if needed)

harshad1:
> One thing I found especially different nowadays is that the todo.txt treatment is gone now. Previously a new task with creationDate was added. Now it's just text added without date.

We check the file format and add the date if it is a todo.txt file:

https://github.com/gsantner/markor/blob/master/app/src/main/java/net/gsantner/markor/activity/DocumentShareIntoFragment.java#L343-L347

```
            if (format == FormatRegistry.FORMAT_TODOTXT) {
                formatted = TodoTxtTask.getToday() + " " + formatted.replaceAll("\\n+", " ");
            } else {
                formatted = "\n" + formatted;
            }
```

PR #2331, opened 2024-06-20
File browser: Make Folder and file icons visually better distinct (outlined icon for file), fixes #2186

### LINKED ISSUE #2186: File browser: Make Folder and file icons visually better distinct (e.g. using color)

### Description

IMO it would be helpful if the folder/file icons were visually more distinct.

When scrolling a folder with many folders and files, it's easy to roughly navigate to the part where folders end and files begin (assuming folders are sorted before files). But since the folder and file icons are so similar (same color grey, very similar shape), the eyes have to scan from top to bottom to find the first file.

An option would be to be able to set the color for folder and file icons.

### Information

Android version: 14 (GrapheneOS)
Device: Pixel 8
App Version: 2.11.0

### Source

F-Droid

### Format / File type

Not specific

### Additional info / Log

_No response_

wshoy:
Ability to easily distinguish between files and directories is essential.
It looks just like in #2186:


harshad1:
Looks nice imo

harshad1:
We tint icons gold to show that they are favorites. Does this interfere with that?

gsantner:
Humbly said, I know some people want a bit more differentiation. At same 
time (including me) want to keep it simple and non-distracting. To me 
the blue looks quite alien to all other colors in the project.

wshoy:
> We tint icons gold to show that they are favorites. Does this interfere with that?

It does not. I have modified only the existing code. It works just like it did before.

wshoy:
> At same time people (including me) want to keep it simple and non-distracting.

I think it's still simple as there are only two things: files and folders. Above all, making the file browser easier to navigate (even when using "folder first") is more important than aesthetics (though I think it looks better).

> To me the blue looks quite alien to all other colors in the project.

The colors are subject to change. The design is taken from #2186.

harshad1:
Perhaps we can make the file icon more different instead? 

Something like this:

image

An outline file icon makes it distinct from the folder icon without looking visually jarring

(From: https://dribbble.com/shots/4404297-Material-Icons-File-And-folder, don't know if this has an appropriate license)

wshoy:
> Perhaps we can make the file icon more different instead?

After looking at other file browsers, for me there are two options:
1. Use color (blue) and keep the icons.
2. Change the icon(s), make folders dark grey and files light grey.

I find the lack of color contrast very unusual. Even most minimal file managers have either colored folders or light files with dark folders.

harshad1:
> light files with dark folders.

This is effectively what I am proposing. Switching to an outline only icon will make them more distinct without changing the appearance otherwise.

wshoy:
> Switching to an outline only icon

I've modified the file SVG and it's like night and day when compared to what we have right now. 

image

gsantner:
Thanks for attaching the screenshot too. It looks very good to me too, while also not distracting.

gsantner:
Need to compile and check it out but looks good code wise.

PR #2351, opened 2024-07-18
Highlight the file when users back

guanglinn:
Add support for highlighting the file/note just opened when users back, this feature can enhance the experience of this editor.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
+++ b/app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
@@ -622,10 +622,8 @@ public class HighlightingEditor extends AppCompatEditText {
          * Reset to the state without line numbers.
          */
         public void reset() {
-            if (_editor.getPaddingLeft() != _defaultPaddingLeft) {
-                _editor.setPadding(_defaultPaddingLeft, _editor.getPaddingTop(), _editor.getPaddingRight(), _editor.getPaddingBottom());
-                _maxNumberDigits = 0;
-            }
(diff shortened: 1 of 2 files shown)


guanglinn:
We can see the obvious advantage, it easily gives us a better sense of location, especially when we have many files/notes. I know we have the "File browser" option, but our intuition is to go back directly instead of clicking that option, so users may just use that option to show the file that opened in other apps in the file browser.

**With this feature:**

https://github.com/user-attachments/assets/5e745e64-6570-4d69-8431-8197c084bb80/

**Without this feature:**

https://github.com/user-attachments/assets/d968525d-3908-4222-b595-b381f0378b18/

gsantner:
Any animation/highlighting of the file should be short amount of time.

Please don't make it constant "if X still open" like elyahw suggested.

guanglinn:
I used the stable method showAndFlash(file) to highlight the file, which makes this feature more stable and simple. So I think the current feature may be better.

harshad1:
This is really nice. Thanks.

I plan to extend showAndFlash() with the ability to optionally highlight the file for longer (Highlight and fade to normal over 5s).

I will pull from some of @guanglinn's previous work to enable this. We can experiment to find a clean approach which makes it really obvious which file the user returned from

gsantner (approved):

guanglinn:
Thanks for your work as well. @harshad1 

I also think there may be a better and cleaner approach to improve this feature, although it works well now. To do these may require more your work (thank you) as you are more familiar with these code of the file browser and others than I am.

PR #2379, opened 2024-07-30
Create launcher widget for todo, closes #1997

### LINKED ISSUE #1997: Document editing launcher widget - Widget Show contents of 1 files

### Description

1. Can a new widget type be considered: a widget type that shows the internal content of one file (chosen by the user).
2. This way one could have multiple widget instances, each showing the internal contents of one file.
3. Double tapping the widget will open Markor on that file in editing mode
4. The widget has a settings button on the widget that lets you edit the widget config (which file it shows + what action to take on tapping the widget - editor or view mode). That way a change doesn't require the lengthy process of removing the widget and creating a new one.
5. Pat yourself on the back it's an awesome little app THANK YOU!!

### Information

Android version: 
Device: 
App Version: 

### Source

F-Droid

### Format / File type

Not specific

### Additional info / Log

```shell
-Just about to donate. This is a great little app. I love that if I put a - or * it will auto enter unstructured list mode and help me by populating that char on the next line.
```

wshoy:
Introduces a Todo widget.

Todo:
- [x] widget color scheme
- [x] place cursor on the clicked line
- [x] add widget preview image
- [ ] highlighting

screenshot


gsantner:
Thanks, let me know when you want a check

as you mentioned "widget color scheme settings": please don't make separate colors for that. You can use the same foreground (text) and background color like at the normal editor ("" basic color scheme"")

harshad1:
Suggestion: reuse our existing widget but replace the provider with one which supplies lines from a todo.txt file.

You'll get 70% of the widget for free with good code reuse.

app/src/main/java/net/gsantner/markor/widget/TodoWidgetRemoteViewsFactory.java
```diff
@@ -0,0 +1,96 @@
...
+    public long getItemId(int position) {
+        return position;
+    }
+
+    @Override
+    public boolean hasStableIds() {
+        return false;
+    }
+
+    private List<String> readFileContent(File file) {
```
harshad1, line 83: I would use the reader functionality in the Document class and then use [TodoTxtTask](https://github.com/gsantner/markor/blob/master/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtTask.java#L85) to parse it

You get a lot of parsing functionality that way.

wshoy:
In these commits:
- used the editor color scheme
- used Document and TodoTxtTask classes
- parse the task text (removed patterns such as creation dates)
- made a temporary workaround for placing the cursor on the clicked line

> Suggestion: reuse our existing widget but replace the provider with one which supplies lines from a todo.txt file.

Which parts can be reused?

harshad1:
> Which parts can be reused?

Our current widget displays a list of files. Clicking on a file opens it.

What you want is a list of todo text lines where clicking on a line opens the todo file at the appropriate line.

We can achieve this by:
1. Changing the provider to provide lines
2. Changing the intent to open the file at the correct line

For 2 we just need to add an EXTRA_FILE with the file and EXTRA_FILE_LINE_NUMBER with the correct line number to each intent.

Both can be done by creating a new factory and reusing the rest of the widget. You will need to modify the other parts of the widget to use the correct factory though.

I may be missing something here, but I don't think so

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
```diff
@@ -287,6 +288,11 @@ public void onPause() {
         _appSettings.addRecentFile(_document.getFile());
         _appSettings.setDocumentPreviewState(_document.getPath(), _isPreviewVisible);
         _appSettings.setLastEditPosition(_document.getPath(), _hlEditor.getSelectionStart());
+
+        if(_document.getPath().equals(_appSettings.getTodoFile().getPath()))
```
harshad1, line 292: I would make an appSettings function called isTodoFile or something

I would also make sure this does not slow down opening / closing the todo file

gsantner: I didn't find any other usage like this (""istodo file check"), so suggest to keep it like that. Just bloats a new method which single usage then.

Opening we should be fine by usual file changed detection rules, as long widget writes when not editing anymore.

Closing not sure maybe the call can be called in a new started thread

harshad1: Ok. Fair.

Either way one should absolute paths

gsantner: Roughly right or do you have something in mind which needs special attention?

guess the most important part is that the widget actually writes & reads filesystem at the right times and not resets it to older state

gsantner:
@wshoy 
What do you think about that, possibly also in a later PR?: To show a file selection dialog when making the widget, to select the file to create the widget for. And in AppSettings we can store what files are currently widget-used (to trigger the widget update)

gsantner:
@wshoy
What do you think about that, possibly also in a later PR?:    
To show a file selection dialog when making the widget, to select the file to create the widget for.   
In AppSettings we can store what files are currently widget-used (to trigger the widget update)


At the end the most important part is that users don't lose data..so the file management part. Years of changes of Markor told..it's very difficult to get that somewhat 99% reliable working for most users.

wshoy:
> To show a file selection dialog when making the widget

It is a good idea. If that happens, I think, that the ToDo functionality should be prioritized in some way (maybe making it the default file or making a separate ToDo widget).
I assume, that the widget would be mostly used for the ToDo file, so it should be made easy for the user to do that.

gsantner:
I assume that there is a split interest of todo vs notes/normal text editing. (so basically resembling why ToDo + QuickNote have their prominent existance)

gsantner:
I assume that there is a split interest of todo vs notes/normal text editing. (so basically resembling why ToDo + QuickNote have their prominent separated existance)

Just to note again - feel free to make in the PR what you have time and interest on. I can only tell what I personally think is important. At the end most important need is that the file handling does not spread havoc and users lose recent modifications 😄

gsantner:
Is the WIP marking still the case?

harshad1:
I want to track down the issues you mentioned about virtual folders not being updated and see if there are any related issues with this approach.

harshad1:
Aside from that this is done

wshoy:
> Is the WIP marking still the case?

It's finished.

app/src/main/java/net/gsantner/markor/widget/TodoWidgetRemoteViewsFactory.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public int getCount() {
+        return _tasks.size();
+    }
+
+    @Override
+    public RemoteViews getViewAt(int position) {
+        RemoteViews views = new RemoteViews(_context.getPackageName(), R.layout.todo_widget_list_item);
+        views.setTextViewText(R.id.todo_widget_item_text, _tasks.get(position).getDescription());
```
harshad1, line 60: It would be really good to also highlight this.

Can do this pretty easily. See getSttHighlighter in MarkorDialogFactory.

gsantner: Plaintext/no highlight would be enough for the first version. Especially if you anyway think about further changes

wshoy: >Plaintext/no highlight would be enough for the first version.
Especially if you anyway think about further changes

Do you mean, that I should merge this PR and make further changes in another?
I also plan on making another widget for regular notes which would take some time.

gsantner: My plan would be to get a initial version that works well enough / not makes crashes. Prefered over full featured and half stuff bare works 😄

wshoy: Well... It works well on my machine 😅
I cannot find other ways to improve it beyond what I mentioned. Maybe you might find something?

gsantner:
has conflicts now unfortunatley

gsantner (approved):

gsantner:
Thanks you, I merged the widget! 😄 

Also tried it: Seems currently limited to just displaying ToDo, and that readonly.

I think editing is not a all too good idea, but selecting the file for each widget would make sense.

ssebs:
Awesome to see this merged! Just curious, when will this be added to a release?

gsantner:
Whenever the next release will be.

enterprisey:
If I may be permitted an off-topic message, thank you so much for working on this feature! It'll be really helpful. I really appreciate the work everyone does!

PR #2384, opened 2024-08-04
(Deprecated) Update flexmark from 0.42.14 to 0.64.8, minSdk to 21

guanglinn:
1. Update flexmark from 0.42.14 to 0.64.8;
2. Update other libraries;
3. Other bug fixes and improvements.


guanglinn:
Since version 0.42.14, `flexmark-java` has obtained a large number of improvements and bug fixes, I think it is worth upgrading it.
After upgrading `flexmark` and other libraries, Markor works well, and no issues occurred yet.

`browser 1.8.0` requires `minSdkVersion` 18 cannot be smaller than version 19.
`appcompat 1.7.0` requires `minSdkVersion` cannot be smaller than version 21.
So now minSdkVersion is 21. But even if we choose API 21 (Android 5.0, in 2014) as minimum SDK, Markor can still run on approximately 99% of devices, so I don't think we need to worry about this.

2024-08-04-min-sdk-version-21

gsantner:
I don't think of increasing minSdk again. Already pushed it from 16 to 18 the last days.

guanglinn:
If there are any issues, I will revert it to 18. 😊

guanglinn:
This PR involves a lot commits unrelated to libraries update, it seems not clean, so I closed this PR and created a new PR #2407.

PR #2388, opened 2024-08-07
Improvements for the action buttons, closes #2386

### LINKED ISSUE #2386: Checkboxes in "Date and time" action button dialog not saved

- [X] The bug is still present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor settings are not touched. In case the issue is resolved there, you don't need to create a bug report. The change will be part of the next Markor update.)

### Description

"Get format instead date or time" and "**Always** use current date &amp; time" checkbox values are not saved. 

The latter checkbox always needs to be unchecked first to pick the date or time. This is not the implied behavior.

### Steps to reproduce

1. Click on the "Date and time" button
2. Check/Uncheck either checkbox
3. Click "OK"
4. Click on the "Date and time" button again

### Information

Android version: 13
Device: 
App Version: 2.12.5

### Source

F-Droid

### Format / File type

Not specific

### Additional info / Log

```shell
-
```

wshoy:
In this PR:
- Date and time
    - Made setting dates easier by not disabling the "Pick Date/Time" buttons and changing the current date checkbox behavior, fixes #2386
    - The format checkbox also disables the "Pick Date/Time" buttons
    - The current date checkbox updates the date preview
    - Removed the word "Always" from the current date checkbox description (retranslation needed) https://github.com/gsantner/markor/issues/2386#issuecomment-2271733903_
- Due date
    - Used "Due date" instead of "Current date" for the due date action button text
- Priority
    - Added long press behavior to set/unset priority from the previous line
    - Made the highlighting colors equally bright and replaced yellow with magenta
- Bold, Italic, etc.
    - Fixed a bug where the button didn't remove characters in the first word


app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtActionButtons.java
```diff
@@ -208,7 +199,36 @@ public boolean onActionLongClick(final @StringRes int action) {
                 }
                 return true;
             }
-            case R.string.abid_todotxt_current_date: {
+            case R.string.abid_todotxt_priority: {
```
harshad1, line 202: Could you describe the behaviour you are trying to achieve here?

wshoy: In my experience, when adding a new task below another, the task's priority would usually be the same.
The goal is to set/unset the priority to be the same as the top one.

This is what felt right to me for multi line:
- if at least one priority is different than the top one - set all to the top priority
- if they are the same and same as the top one - remove priorities as there is nothing else to do (another press restores it)

That also means, that any priorities can be removed with at most two presses.
Might also add a check for the bottom line as well.

harshad1: So long pressing the priority button copies the priority from the first task to all the others? 

Does this really need a long press? Currently one can select all the tasks and assign them a priority pretty easily.

You can select multiple tasks by simply creating a selection covering multiple tasks in the editor or by using the search dialog.

So search -> select all -> priority action -> select priority to assign to all selected tasks.

wshoy: > So long pressing the priority button copies the priority from the first task to all the others?

By "top" I mean the previous line from selection. Priority is copied to the selected lines. 
The exact behavior is not set in stone.

> Does this really need a long press? Currently one can select all the tasks and assign them a priority pretty easily.

In my experience, it feels much easier and responsive in daily use, as it makes creating another task a little bit faster than the other methods.

harshad1: Ah.

So the current workflow is:

select -> priority action -> choose

And the new workflow for this use case is

select -> long press priority action 

We're replacing 2 short presses with one long press.

wshoy: Well said

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtBasicSyntaxHighlighter.java
```diff
@@ -11,11 +11,11 @@ public class TodoTxtBasicSyntaxHighlighter extends SyntaxHighlighterBase {
     private final static int COLOR_CONTEXT = 0xff88b04b;
 
     private final static int COLOR_PRIORITY_A = 0xffEF2929;
-    private final static int COLOR_PRIORITY_B = 0xffF57900;
-    private final static int COLOR_PRIORITY_C = 0xff73D216;
-    private final static int COLOR_PRIORITY_D = 0xff0099CC;
-    private final static int COLOR_PRIORITY_E = 0xffEDD400;
-    private final static int COLOR_PRIORITY_F = 0xff888A85;
+    private final static int COLOR_PRIORITY_B = 0xffd16900;
```
gsantner, line 14: Is there a reason why to change the priority colors suddenly?

gsantner, line 14: Why did you change those colors?

wshoy: The B and C colors were disproportionately bright compared to the red color of A. The A priority, being most important, should be at least as much noticeable as the other ones.
The yellow E color was much brighter than all the other colors and didn't match the hue sequence.

I used the [perceived brightness](https://alienryderflex.com/hsp.html) formula to match the colors:

Before:
EF2929 - 0.53
F57900 - 0.64
73D216 - 0.68
0099CC - 0.53
EDD400 - 0.81
888A85 - 0.54

After:
EF2929 - 0.53
d16900 - 0.55
59a112 - 0.52
0091c2 - 0.51
a952cb - 0.51
878986 - 0.53

gsantner (approved):
Thank you!

Issue #2373, opened 2024-07-27
Markdown scroll sync in preview and src


guanglinn:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **enhancement/feature request**. Not a [bug or question](https://github.com/gsantner/markor/issues/new/choose).
- [X] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [X] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [X] The wanted feature/enhancement is not present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor's settings are not touched. If your desired feature is present, you don't need to open this issue. The change will be part of the next Markor update.)

### Description

This is a feature request.  
Hope Markor supports the feature that scroll sync in preview/view mode and src/edit mode.

### Information

Android version: 2.12.4 (code 153)
Device: Android 12/HarmonyOS 4.2.0
App Version: Huawei Nova 8

### Source

F-Droid

### Format / File type

Markdown

### Additional info / Log

```shell
-
```

gsantner:
#2189 #679 #1070

guanglinn:
I am sorry for not noticing those closed issues.

guanglinn:
This should be an advanced feature, but I noticed that [Visual Studio Code](https://github.com/microsoft/vscode/) is doing well in Markdown scroll sync, the code of this feature is here [vscode/extensions/markdown-language-features/preview-src/scroll-sync.ts](https://github.com/microsoft/vscode/blob/main/extensions/markdown-language-features/preview-src/scroll-sync.ts). I think it may be a valuable reference for us. @gsantner @harshad1 

https://github.com/user-attachments/assets/c3f10263-26f9-4584-b5f8-56afb9a95d14/

harshad1:
To this end every node in the markdown html now has a line=# attribute. This should be sufficient to jump when going back and forth if handled well, I think

gsantner:
Unfortunatley it's though markdown specific and next up people cry out for it supported for other formats. It's not generic solveable in easy way

guanglinn:
[...]

Issue #2189, opened 2023-12-27
Added an option to synchronize preview and edit positions


naosense:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **enhancement/feature request**. Not a [bug or question](https://github.com/gsantner/markor/issues/new/choose).
- [X] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [X] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [X] The wanted feature/enhancement is not present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor's settings are not touched. If your desired feature is present, you don't need to open this issue. The change will be part of the next Markor update.)

### Description

 Currently, the preview window scrolling and editing position are separated. This is especially inconvenient when you're tweaking the text while you're skimming through the article. I hope adding an option to synchronize the preview and edit positions. Thanks in advance!

### Information

Android version: 
Device: 
App Version: 

### Source

GitHub Releases

### Format / File type

Markdown

### Additional info / Log

```shell
-
```

gsantner:
Please browse existing issues before creating a new one. It has been asked a few times.

Thank you.

naosense:
> Please browse existing issues before creating a new one. It has been asked a few times. Thank you.

sry, i do searched issues before i created this, maybe the keyword i choose is not right...

naosense:
after i browse existing issues, i dont find any related ones.

gsantner:
#679 #1070 for example

Issue #1070, opened 2020-09-29
switching from edit to view mode doesn't remember cursor position


seelaman:
#### General information

* **App version:**
2.3.3 Playstore

* **System:** 
Stock Android 10, Samsung Note 9

#### Description
When I'm in a long markdown file and I switch from edit mode to view mode (for example to jump to a link), the document is displayed scrolled to the top. It would be nice if the view was centered around the cursor position from edit mode so I don't have to use the search to jump back to where I was.

It's similar to [#205](https://github.com/gsantner/markor/issues/205) but that one only applied to "default to view mode" which I'm not using.

Implementation wise it could also be similar to [#1025](https://github.com/gsantner/markor/issues/1025)

gsantner:
This has been already requested, and I marked it as a won't fix (#679). It's due it's not feasible to implement, the view content (/HTML) does not fully correlate to the original text (otherwise it wouldn't have been converted :P).

Issue #1060, opened 2020-09-14
md: Disable edit-mode ToC in view-mode


contrapunctus-1:
Thanks for adding the ToC - it makes it easy to read large files. I have a request for a behaviour which would greatly ease their editing as well.

## Situation
1. View mode. User taps a ToC heading.
2. They switch to edit mode.

## Current  behavior
The cursor is placed at the beginning of the document (even with the 'open at end' option enabled? Might be a bug.)

## Suggested behaviour
Place the cursor at or below the heading.

If there is a concern about a case where the user -

1. taps a ToC heading
2. scrolls around, possibly going to another section
3. switches to edit mode

...then the behavior could be to fall back to usual behavior (cursor at start or end of document) if the user scrolls after tapping the ToC heading.

This would make editing large files a lot easier.

gsantner:
Hi, that ToC is only there for edit mode, hence it should be only available in edit mode.

Thus, resolve will be that the edit mode ToC gets disabled/unavailable on view mode.

contrapunctus-1:
Hey, @gsantner . I'm not sure what you mean by "edit mode ToC" - I haven't seen any ToC in the edit mode. 🤔 I'm running v2.3.2 from F-Droid.

gsantner:
Please share a screenshot **what** you actually tap, thats clears things.

contrapunctus-1:
@gsantner Here is a video demonstrating the problem - https://upload.disroot.org/r/bPbuFGWV#ukUtez62aWnV3AvNGsVrBHgvRjpjBNGR8Ogomo21Og8=

gsantner:
OK, so your request is to synchronize view mode scroll state with edit mode scroll state.

This has been already requested, and I marked it as a won't fix. It's due it's not feasible to implement, the view content (/HTML) does not fully correlate to the original text (otherwise it wouldn't have been converted :P).

gsantner:
OK, so your request is to synchronize view mode scroll state with edit mode scroll state.

This has been already requested, and I marked it as a won't fix (#679). It's due it's not feasible to implement, the view content (/HTML) does not fully correlate to the original text (otherwise it wouldn't have been converted :P).

contrapunctus-1:
@gsanter I'm well aware of that issue. This is not a duplicate; note the title I had set - "Use the ToC heading we tapped on". The program just derives the information of where to open edit mode from the ToC heading the user tapped.

gsantner:
[...]

PR #2404, opened 2024-09-03
Add support for scroll position sync

guanglinn:
Add basic support for scroll position sync between Markdown preview (view-mode) and its source code (edit-mode).

This feature supports:
* Sync position from edit to preview;
* Sync position from preview to edit.

The related issues:  
#2373 #2189 #1070 #1060 #679.


guanglinn:
You can see the demonstration,

https://github.com/user-attachments/assets/e2551f0e-7449-49d0-8a70-3ac22d22cf5f/

gsantner:
conflicts

guanglinn:
Resolved conflicts :)

app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
```diff
@@ -2538,7 +2538,7 @@ public <T extends GsContextUtils> T showSoftKeyboard(final Activity activity, fi
                 if (show) {
                     imm.showSoftInput(focus, InputMethodManager.SHOW_IMPLICIT);
                 } else if (token != null) {
-                    imm.hideSoftInputFromWindow(token, InputMethodManager.HIDE_NOT_ALWAYS);
+                    imm.hideSoftInputFromWindow(token, 0);
```
gsantner, line 2541: This might make issues.

gsantner (changes requested):
Tried locally,

unfortunately not works quite well for both directions. From View-Mode to Edit-Mode it seems to works good enough.  
But from Edit to View mode it seems to not work, wlways jumps to top of document.   
This is the builtin Markdown reference.


[screenrecord-2024-09-25_00.59.29.webm](https://github.com/user-attachments/assets/9de9448e-c9d1-484c-85fb-a25e3fc1d34b)

gsantner:
I've merged your other PR (and others) to master as well, due your other one this one anyway has a conflict again

gsantner:
I've merged your other PR to master as well, made conflicts again

app/src/main/java/net/gsantner/opoc/util/GsContextUtils.java
guanglinn: OK, I reverted it.

guanglinn:
Conflicts are resolved now.

guanglinn:
Sure, from Edit to View, it works well for quite a lot common text blocks such as Headings, Bold, Italic, Blockquotes, Lists, Links, Images, ..., but it works not quite well for some special text blocks.
I have tried to solve this, it's quite difficult for myself, and I think this is already a big improvement for Markor to provide a rough position synch for Markdown.
As mentioned above, you can merge it first, it is stable now, and I will improve it if I have some better solutions later, or other contributors will make it perfect on this basis.

gsantner:
What I was suggest for now: Save/Restore last webview scroll position after short delay (when blocks have the right rendered size).

As far I recall this has been before like that, but somehow it doesn't work currently on master either

guanglinn:
Can an option be added? Allow users to choose to make Markor restore last WebView scroll position, or sync WebView scroll position from Edit.

gsantner:
I don't think it's neccassary. When scroll sync work good enough in praxis there is also the editor position saved/restored - so will be view-mode position when switiching to.

gsantner:
Unfortunatley also keep in mind what I tried to say always - Markor is not only Markdown. This all relies on Markdown specific injections/adjusting the original.

This won't work for stuff that doesn't base on Markdown converter.

guanglinn:
Improvements: If we want to go back to the last view position, just long click toolbar.

gsantner:
@harshad1 may I ask about your thoughts, how good does it work on your device?

I watched guanglinns video again, but it just not works as good here at least in one way. In such cases I don't feel too well either with bearing that by default to users.

gsantner:
@harshad1  @wshoy @elyahw 
may I ask about your thoughts, how good does it work on your device?

I watched guanglinns video again, but it just not works as good here at least in one way, which overall seems also worse then without the PR. In such case I don't feel too well either with bearing that by default to users.

wshoy:
This feature feels great on an emulator, but unfortunately, it mostly doesn't work on my phone.
When it does, it's only a couple times after reloading the file, and only from view mode to edit mode. Then it doesn't work at all.

On an emulator, when going from view mode to edit mode, it often jumps to the next line. Probably because of the word wrapping.

app/src/main/java/net/gsantner/markor/frontend/textview/TextViewUtils.java
```diff
@@ -352,7 +345,11 @@ public static void showSelection(final TextView text, final int start, final int
         text.requestRectangleOnScreen(region);
     }
 
-    public static void setSelectionAndShow(final EditText edit, final int... sel) {
+    public static void showSelection(final TextView text) {
+        showSelection(text, text.getSelectionStart(), text.getSelectionEnd());
+    }
+
+    public static void setSelectionAndShow(final EditText edit, boolean setSelection, final int... sel) {
```
harshad1, line 352: We have `showSelection` to show the selection and `setSelectionAndShow` to set selection and show. Having a boolean flag here to setSelection in a function called `setSelectionAndShow` is poor design imo. Just call the right function

app/src/main/java/net/gsantner/markor/frontend/textview/TextViewUtils.java
```diff
@@ -364,12 +361,17 @@ public static void setSelectionAndShow(final EditText edit, final int... sel) {
...
-
-            edit.setSelection(start, end);
+            if (setSelection) {
+                edit.setSelection(start, end);
+            }
             showSelection(edit, start, end);
         }
     }
 
+    public static void setSelectionAndShow(final EditText edit, final int... sel) {
```
harshad1, line 371: Separate functions are better than flags unless there is a good reason.

app/src/main/java/net/gsantner/markor/frontend/textview/TextViewUtils.java
guanglinn: Okay, I have made some improvements for this.

gsantner:
PR has many conflicts (and some open discussionsI, converting it to draft for now.
Please let me know when want to move on.

Thanks!

guanglinn:
> PR has many conflicts (and some open discussionsI, converting it to draft for now. Please let me know when want to move on.
> 
> Thanks!

Thanks,
Maybe it takes some time, maybe I need some help 😕

gsantner:
No problem and understandable.

I also have quite little time available recently, and have to take care of my physical health.

guanglinn:
Yes, time and physical health are the most precious.
This is not a smooth or successful PR for me, I closed it now. Maybe I will reopen it after sufficient improvement or open a new one.

PR #2407, opened 2024-09-04
Update flexmark from 0.42.14 to 0.64.8, minSdk to 21

guanglinn:
1. Update flexmark from 0.42.14 to 0.64.8;
2. Update some dependency libraries;
3. Update minSdk from 18 to 21.


guanglinn:
Since version `0.42.14`, `flexmark-java` has got a large number of improvements and bug fixes, so it is worth upgrading it.

Marder works well and no issues occurred yet after updating `flexmark` and other libraries.

`browser 1.8.0` requires `minSdkVersion` cannot be smaller than version 19.
`appcompat 1.7.0` requires `minSdkVersion` cannot be smaller than version **21**.

Even if we choose API 21 (Android 5.0, in 2014) as minimum SDK, Markor can still run on approximately **99%** of devices, so we don't need to worry too much.

2024-08-04-min-sdk-version-21

gsantner:
For sure there are improvements at every release. Though is there something specific that gave you reason to get a newer version?

JSchmerling:
From what I've seen so far, the benefits of upgrading are improved performance (https://github.com/vsch/flexmark-java/wiki/Version-0.60.0-Changes) and better handling of MarkDown extensions (flexmark-all).

gsantner:
I think the time is now right to update to a newer minimum SDK version. Let's merge this after the next update.

guanglinn:
I will implement rich text pasting after this PR merged.

PR #2377, opened 2024-07-29
Navigation and highlighting improvements

harshad1:
This adds several improvements to file browsing and navigation 

1. Better handling + fixes to handling of special / virtual folders
2. back goes to previous folder (needs to be refined more). Addresses #2337 
3. Fixed activity back-stack and link navigation

Todo:
- [x] Fix file dialog subtitle padding and truncation
- [x] Back should ask before closing markor

I have also added other things to this PR now

1. Highlighting happens in a separate thread making it less janky when typing fast (especially in very long todo lists)
2. We had a lot of code which did not check if selection was invalid - added defensive logic everywhere
3. Several cleanups
4. Remember share into format as link state from last time

PR #2408, opened 2024-09-04
WIP: Navigation improvements

harshad1:
This is basically a re-creation of #2377 

For some reason that PR is constantly failing CI. Seeing if a new PR works.


harshad1:
So the issue was jcenter dying. Have fixed. Merging back

PR #2411, opened 2024-09-10
Updated README with dark/light images

AntoninDelFabbro:
Real minor update for reading confort.
I wanted to test that GitHub option… And to thank you for this masterpiece of application! Thank you! It's now a core app in my routine!


gsantner:
Please keep the original image under the same URL. Some things do/might link to that filepath/full URL. The new one can have dark postfix or alike.

gsantner (changes requested):
Thank you

Please keep the original image under the same URL. Some things do/might link to that filepath/full URL. The new one can have dark postfix or alike.

gsantner:
Thank you!

PR #2416, opened 2024-09-16
GitHub Actions CI configuration - upgrade upload step, fixes CI build error

gsantner:
GitHub Actions CI configuration - upgrade upload step, fixes CI build error

### DIFF
--- a/.github/workflows/build-android-project.yml
+++ b/.github/workflows/build-android-project.yml
@@ -57,14 +57,14 @@ jobs:
 
     - name: "Artifacts: All"
       if: always()
-      uses: actions/upload-artifact@v2.2.1
+      uses: actions/upload-artifact@v4
       with:
         name: "all"
         path: dist
         retention-days: 5
 
     - name: "Artifacts: Android APK"
-      uses: actions/upload-artifact@v2.2.1
+      uses: actions/upload-artifact@v4
       with:
         name: "android-apk"
         path: |


TotallyAvailable:
@gsantner  ***Please*** remember to set the GITHUB_TOKEN permissions (& 'Secret Source') for **all** workflows across ***all*** your repos to the ***absolute minimum*** necessary to avoid further use of **default write-all** tokens.

This message has **nothing** to do with this PR and should only be seen as a heads-up to reduce the risk of leaking highly privileged tokens in the future. Should this scenario not be of concern, ignore this message. Apologies again for using this way of reaching out.

gsantner:
Hello @TotallyAvailable ,
I'm not aware of setting, configuring or exposing a GH_TOKEN or GITHUB_TOKEN anywhere in my projects.  
Do you know of something than I'm not aware off, or just some general hint?

TotallyAvailable:
May I dare to say that this is indeed the problem behind the initially meant "as simple reminder" heads-up. You appear to be among the ***absolute*** majority of users GitHub ***did not*** account for when they made certain "decisions". ***Not*** meant to be seen as any kind of insult towards you (!)

- Your best friend is going to be ```permissions: {}``` from now on followed by researching the suggested read/write declarations ***per*** used action in ***every step*** within every workflow file in ***every*** project you have control over. ***Do not rely*** on "undefined behavior".
- Consider if 'persist-credentials' *is* necessary for any of your workflows. 
- Try to set 'Secret Source' to None unless necessary.
- Remove 'pull_request' as trigger unless necessary. 
- Final releases should only ***ever*** be triggered manually.
- Consider setting up  SECURITY.md in case of any other unrelated issue in the future. 
- Be careful of what ends up in an Artifact.
- Do not rely on the "integrity" of mutable release(-notes/-assets)
- Consider posting checksums (including from the used signing cert) through commits to the readme. (Checksums for Release vX.Y.Z - DD.MM.YY) <- as a commit message example.

Some of the mentioned points may not apply to this or any other project you are involved in. You'll be far from the only one "surprised" when all of this starts to make sense. 

Again not meant to be "rude" in any way ***at all*** ...the blame here falls somewhere else completely. ***I*** should not be going around a not further specified number of repos to point this out due to the "it's fine" attitude coming from above, despite several "incidents" in the past.

Should something feel or sound "too vague", please consider it as intended. Thank you for your attention and eventual steps taken to deal with this issue. Resolving the actual underlying issue is unfortunately not within our power.

How big this whole thing actually is ?
```https://github.com/dependabot/dependabot-core/commit/c69174a5a3ca14c953e2ae96a05a5d3b725a1d58```
Goes well beyond that. Been like that for ***years***, still relevant.

PR #2419, opened 2024-09-18
Fix how multiple windows are launched

harshad1:
Yet more fixes for multiple windows. This makes the behaviour pretty clean IMO


gsantner (approved):
Thanks!

Issue #2410, opened 2024-09-09
Use a single process for the whole app


elyahw:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **enhancement/feature request**. Not a [bug or question](https://github.com/gsantner/markor/issues/new/choose).
- [X] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [X] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [X] The wanted feature/enhancement is not present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor's settings are not touched. If your desired feature is present, you don't need to open this issue. The change will be part of the next Markor update.)

### Description

I noticed that there are 3 different processes being launched (one for the settings, one for file browser, and one for editor).

This is a little bit confusing when task switching on Android, as most of the time I have 2 open instances of the Markor (which are the file browser and the editor).

I think it would useful if all of them were merged into a single process.

### Information

Android version: 13
Device: Pixel 5a
App Version: Latest v2.12.5

### Source

GitHub Releases

### Format / File type

Not specific

harshad1:
I have made some improvements to this in #2377

harshad1:
Do you have this option unchecked?

![Screenshot_20240918-072621](https://github.com/user-attachments/assets/75d17bd9-fb8c-47d1-ae3e-97404ee7ea0b)

harshad1:
Hmmm.. it should be unchecked by default. I'll fix it if it isn't.

gsantner:
It was default multiple windows when I implemented it first time, as far 
I recall.

gsantner:
@elyahw 
Don't get me wrong, but it seems you know the code base somewhat good. As you are developing on a crippled down fork of Markor.  In the recent months saw a increased amount of things you are asking for.

May I politely ask you to contribute to this project, especially if you want something for your fork. Don't let the others do the stuff you want and take their time to answer your stuff, please start contributing. I checked and didn't see any yet: <https://github.com/gsantner/markor/pulls?q=is%3Apr+author%3Aelyahw>.

[...]

PR #2420, opened 2024-09-19
Disable multi-window settings option by default

elyahw:
Change the default option described in this comment:
https://github.com/gsantner/markor/issues/2410#issuecomment-2359629336


gsantner:
Thanks, but to change defaults it needs to be changed also in the settings xml. This returns default value for java callers. UI has a own XML based value.

gsantner (approved):
thanks, merging

PR #2421, opened 2024-09-20
Fixed checking for indices in getSelectedLines

harshad1:
Handle cases where we pass in a single index


harshad1:
This bug caused navigation to links to fail. Fixed now.

gsantner (approved):
Thanks!

Issue #2382, opened 2024-08-01
Touch parent folders when editing a file


ghost:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **enhancement/feature request**. Not a [bug or question](https://github.com/gsantner/markor/issues/new/choose).
- [X] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [X] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [X] The wanted feature/enhancement is not present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor's settings are not touched. If your desired feature is present, you don't need to open this issue. The change will be part of the next Markor update.)

### Description

1. Currently, when  _sorting folders first_ option is turned off, when opening one of the folders and editing one of the notes inside:
2. The _file_ is touched (modification date is updated). But the _folder_ is **not touched**, its modification date remains in the past.
3. This is a problem because from the main menu, it is not possible to tell which folder was edited last.

### Information

Android version: 13
Device: Pixel a5
App Version: 2.12.5

### Source

GitHub Releases

### Format / File type

Not specific

### Additional info / Log

```shell
-
```

gsantner:
I'm not sure if I understand rught, but it might be just how the filesystem used for that folder works.

Changing a file in a folder not always leads to a timestamp change in the folder itself. And Markor only reads out filesystem metadata.

gsantner:
Does any other editor or alike do that?

gsantner:
Does any other editor or alike do that - touching all parent directories?

I'm not aware of that at least and it may have unwanted side effects (i.e. with sync clients, permissions etc).

gsantner:
I'm not sure if I understand right, but it might be just how the filesystem used for that folder works.

Changing a file in a folder not always leads to a timestamp change in the parent director(ies). And Markor only reads out filesystem metadata.

> This is a problem because from the main menu, it is not possible to tell which folder was edited last.

Correct, you may not put too much relevance in what timestamps directories have. But thats regardless of what software you use.

[...]

PR #2422, opened 2024-09-21
Touch parent folder on file modification

ghost:
This solves https://github.com/gsantner/markor/issues/2382


harshad1:
I still oppose including this in mainline markor (we will have to check the effect on every sync solution etc as this is nonstandard behavior for a file manager)

That said, if the author would like to merge this into their personal branch, I would make the following changes:

1. Touch the parent folder in the Document class
2. Don't use currentTimeMillis(), instead copy the stamp from the just written file

gsantner:
I can also suggest to guard it with try catch.

This change is at pretty much the most dangerous point of the project. You don't want to lose any file contents, just because your filesystem throws some error by touching.

Issue #2435, opened 2024-10-06
Per directory sort settings


harshad1:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **enhancement/feature request**. Not a [bug or question](https://github.com/gsantner/markor/issues/new/choose).
- [X] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [X] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [X] The wanted feature/enhancement is not present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor's settings are not touched. If your desired feature is present, you don't need to open this issue. The change will be part of the next Markor update.)

### Description

It would be good to have per-directory file sort settings

For example, I generally like to sort by name with folder first, but would like the Downloads folder to be sorted by time.

### Information

Android version: 
Device: 
App Version: 

### Source

F-Droid

### Format / File type

Not specific

VARTIK-ARSH:
@gsantner can i work on it.

gsantner:
Yes.

VARTIK-ARSH:
@gsantner please assign the project to me.

PR #2437, opened 2024-10-11
#2435 solved

VARTIK-ARSH:
@gsantner **The problem has been resolved; it will now sort by date first, then alphabetically.**


gsantner:
#2435 is about having possibility to have sort settings directory specific saved/restored i.e. by additional checkbox to enable it.

This PR toggles some boolean flags, which is quite something different.

gsantner:
Check out the AppSettings.java file and you will notice some per-file options, like `setDocumentWrapState` is.

PR #2440, opened 2024-10-13
FileManager: Remove top submenu to open special folders, load storage overview in filemanager instead

gsantner:
* Remove top menu submenu  to open special folders, go to virtual storage directory folder instead
* There all special folders/devices/lists are available too, and only needs to be maintained at one location then
* this way less duplication of those options, and all of them are available outside mainactivity too

screenshot-2024-10-14__01-22-56-append


harshad1:
This is nice. Makes navigation cleaner imo.

Should we also remove long press fab to jump folders?

gsantner:
Would make that jump to storage too, instead of the current target

PR #2443, opened 2024-10-14
Select lines

harshad1:
This adds an option to select lines which pops up in the long press menu in the text editor. The menu remains open so we can easily copy or replace selections.


harshad1:
image

app/src/main/java/net/gsantner/markor/model/AppSettings.java
```diff
@@ -746,57 +746,6 @@ public String getNavigationBarColor() {
         return getString(R.string.pref_key__navigationbar_color, "#000000");
     }
 
-    public String getAppStartupFolderMenuId() {
```
harshad1, line 749: These aren't really needed any more

app/src/main/java/net/gsantner/markor/activity/MainActivity.java
```diff
@@ -411,8 +411,7 @@ public GsFileBrowserOptions.Options getFilesystemFragmentOptions(GsFileBrowserOp
                 public void onFsViewerConfig(GsFileBrowserOptions.Options dopt) {
                     dopt.descModtimeInsteadOfParent = true;
                     dopt.rootFolder = _appSettings.getNotebookDirectory();
-                    final File fallback = _appSettings.getFolderToLoadByMenuId(_appSettings.getAppStartupFolderMenuId());
-                    final File file = MarkorContextUtils.getValidIntentFile(getIntent(), fallback);
+                    final File file = MarkorContextUtils.getValidIntentFile(getIntent(), dopt.rootFolder);
```
gsantner, line 414: Does Markor still startup with i.e. Favorites or storage if you select that in the filebrowser app settings?

gsantner:
Might I also suggest a unicode symbol instead of text? I know already that translating those two words will result in what you consider a sentence. And you want to have more then just this option 😄

app/src/main/java/net/gsantner/opoc/frontend/filebrowser/GsFileBrowserListAdapter.java
```diff
@@ -840,25 +831,28 @@ private StringFilter(GsFileBrowserListAdapter adapter, List<File> adapterData) {
...
-                _filteredList.addAll(_originalList);
-            } else {
-                for (File file : _originalList) {
-                    if (file.getName().toLowerCase(Locale.getDefault()).contains(constraint)) {
-                        _filteredList.add(file);
-                    }
-                }
-            }
+            _lastFilter = constraint.toString().toLowerCase().trim();
+            _filter(_originalList, _filteredList);
```
harshad1, line 836: we don't want to run an async filter inside load folder as it is already async. So we break the internal processing out into a stand-alone function

harshad1:
Changed from text to icon

Screenshot_20241109-103703

harshad1:
Resolved merge conflict. This is ready to go.

gsantner:
Thank you, merged

PR #2450, opened 2024-10-17
orgmode: add more text styling syntax and buttons

nyxkn:
This adds syntax highlighting for bold, italic, strikeout, underline, and inline code.

For the matching regex I've used the same one used for markdown bold. To my testing this matches the expected behaviour. The only change is that we do not want to allow square brackets as a surround.


gsantner:
Are there any exact same regex used? In that case the Markdown one can be referenced/variable copied instead of copypaste the regex string

nyxkn:
Markdown bold is almost the same but I've removed the square brackets from the lookbehind and lookahead. So maybe not straightforward to reuse?

```
markdown bold: "(?<=(\\n|^|\\s|\\[|\\{|\\())(([*_]){2,3})(?=\\S)(.*?)\\S\\2(?=(\\n|$|\\s|\\.|,|:|;|-|\\]|\\}|\\)))"

orgmode: "(?<=(\\n|^|\\s|\\{|\\())([*])(?=\\S)(.*?)\\S\\2(?=(\\n|$|\\s|\\.|,|:|;|-|\\}|\\)))"
```

But all the regex I use for the 5 styles in orgmode are identical to each other, just matching a different symbol. So I can definitely reuse the common parts there!

gsantner:
Yes read that, so asked regarding others

gsantner:
Thanks, are you happy with the changes of the PR now or something still in progress?

gsantner (approved):

nyxkn:
All good with me, thanks!

gsantner:
thank you, merged

PR #2452, opened 2024-10-18
WIP: ToDo widget text highlighting

### LINKED ISSUE #1997: Document editing launcher widget - Widget Show contents of 1 files

wshoy:
Where should I get color from?

- [ ] fix color


harshad1:
See how the text in the Todo txt dialogs are highlighted. This will be a 1 function call

app/src/main/java/net/gsantner/markor/widget/TodoWidgetRemoteViewsFactory.java
```diff
@@ -86,4 +94,84 @@ public long getItemId(int position) {
     public boolean hasStableIds() {
         return false;
     }
+
+    private Spannable getTaskSpannable(TodoTxtTask task) {
```
harshad1, line 98: This static function will return a highlighter callable which can apply todotxt highlighting to a spannable.

https://github.com/gsantner/markor/blob/master/app/src/main/java/net/gsantner/markor/frontend/MarkorDialogFactory.java#L758

So simply 
```
final GsCallback.a1<Spannable> hl = getSttHighlighter();
for (Spannable s : lines) {
    hl.callback(s)
}
```

harshad1: This will highlight every line in `lines` 

You will need to make getSttHighlighter public and move it somewhere appropriate.

gsantner:
from me:

* best reuse as much possible from existing highlighting stuff and colors
* try to keep todo.txt specific a bit separated, with that in mind that it is to be expanded to other filetypes / other textedit widget in future

app/src/main/java/net/gsantner/markor/widget/TodoWidgetRemoteViewsFactory.java
wshoy:
```
final GsCallback.a1<Spannable> hl = getSttHighlighter();
Spannable s = new SpannableString(_tasks.get(position).getLine());
hl.callback(s);
views.setTextViewText(R.id.todo_widget_item_text, s);
```
I couldn't make it work. What is wrong here?

It's also very important to remove creation dates, priority parentheses, etc. as they take up valuable space especially on narrower screens. How would that be achieved?,

gsantner:
Hello @wshoy
do you want to make further changes (as I understood from the discussion it's not finished)

wshoy:
The highlighter doesn't seem to work, then I guess that's all. Just not sure about the duplicated colors.

Somewhere in the future I plan to make a separate plain text widget. #1997

PR #2455, opened 2024-10-21
Override .org file extension to use text mimetype (*/* unrecognized on older devices)

antler5:
Prevents `QuickNote.org` from being filtered out of the QuickNote file-picking dialog.

### DIFF
--- a/app/src/main/java/net/gsantner/opoc/util/GsFileUtils.java
+++ b/app/src/main/java/net/gsantner/opoc/util/GsFileUtils.java
@@ -461,6 +461,8 @@ public class GsFileUtils {
             return "text/markdown";
         } else if (ext.matches("(te?xt)|(taskpaper)")) {
             return "text/plain";
+        } else if (ext.matches("org")) {
+            return "text/org";
         } else if (ext.matches("webp")) {
             return "image/webp";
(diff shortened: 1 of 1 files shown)


gsantner:
Thank you, but if we explicitly make some remapping here..better to somewhat the right type

text/org https://orgmode.org/worg/org-media-type.html

antler5:
Ah, I haven't filled out the device information box that would have been in the template if i'd opened an issue -- i stopped getting android updates many years ago

It's up to you whether that's worth merging a fallback for, but ime there's (evidently) no way for a user who wants this to work around their android's poor mimetype api without forking to extend this list. with that in mind, i don't care what the mimetype is as long as it starts with 'text/' >u<

i think i'm against filtered file-pickers without user-facing toggles in general, but idk java so i stopped when i'd fixed my problem -- maybe another day if that'd be a perferred solution

gsantner:
Does it show up with text/org mime type now on your device, and it properly opens the files in editor on click?

gsantner (approved):

antler5:
> Does it show up with text/org mime type now on your device, and it properly opens the files in editor on click?

Yep, opens as-expected (in the correct mode)
~I do wonder what would happen if QuickNote was set to a `.jenc` file~

PR #2466, opened 2024-10-31
Update links

leoheitmannruiz:
The fdroid.app: thingies don't seem to render correctly https://f-droid.org/packages/net.gsantner.markor/

Also, it seems Markor is the only app in F-Droid that uses them:

```
$ grep "fdroid.app:" index-v2.json | wc -l
       1
```

Perhaps it makes sense to update the links here and deprecate the fdroid.app: thingies over at F-Droid. I'll ask them.

### DIFF
--- a/metadata/en-US/full_description.txt
+++ b/metadata/en-US/full_description.txt
@@ -17,9 +17,9 @@
 👌 No ads or unnecessary permissions
 🌎 Language selection -- use other language than on the system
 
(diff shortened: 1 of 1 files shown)


gsantner:
Hello, Does the new link open in F-Droid app if you have it installed? Thats what that prefix is about

gsantner:
it would just be good to use links that lead to Markor page inside it. 

As far I see www.f-droid.org, but not without www is registered.

gsantner:
[](https://postimg.cc/WFCpY20R)

gsantner (approved):

PR #2470, opened 2024-11-06
Improve line numbers

guanglinn:
Refactor this feature:
- Improve line numbers experience
- Optimize algorithm of drawing line numbers


guanglinn:
The line numbers is fixed to the left

Before
https://github.com/user-attachments/assets/4ed36da0-4be5-4120-a99e-917d97ea5be6/

Now
https://github.com/user-attachments/assets/09db6426-5569-48da-974e-bc3c6dc9304a/

app/src/main/java/net/gsantner/markor/frontend/textview/LineNumbersTextView.java
```diff
@@ -0,0 +1,302 @@
...
+import android.text.TextWatcher;
+import android.util.AttributeSet;
+import android.widget.EditText;
+
+import androidx.appcompat.widget.AppCompatTextView;
+
+/**
+ * @author Li Guanglin
+ */
+public class LineNumbersTextView extends AppCompatTextView {
```
harshad1, line 18: We disable some accessibility things in Highlighting editor for performance. May be good to disable them here too

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
```diff
@@ -132,9 +137,23 @@ public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
...
+            if (_lineNumbersView.isLineNumbersEnabled()) {
+                final long time = System.currentTimeMillis();
+                if (time - _lineNumbersRefreshTime > 125) {
+                    _lineNumbersRefreshTime = time;
+                    _lineNumbersView.forceRefresh();
+                }
+            }
+        });
+        _lineNumbersView = view.findViewById(R.id.document__fragment__edit__line_numbers_view);
+        _lineNumbersView.setEditText(_hlEditor);
```
harshad1, line 155: IMO we should add the scroll change listener inside setEditText.

setEditText() should remove the listener from an the existing edit text when this is done

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
```diff
@@ -132,9 +137,23 @@ public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
...
         _hlEditor = view.findViewById(R.id.document__fragment__edit__highlighting_editor);
+        _editorHolder = view.findViewById(R.id.document__fragment__edit__editor_holder);
         _textActionsBar = view.findViewById(R.id.document__fragment__edit__text_actions_bar);
         _webView = view.findViewById(R.id.document__fragment_view_webview);
         _primaryScrollView = view.findViewById(R.id.document__fragment__edit__content_editor__scrolling_parent);
+        _primaryScrollView.getViewTreeObserver().addOnScrollChangedListener(() ->
+        {
+            if (_lineNumbersView.isLineNumbersEnabled()) {
+                final long time = System.currentTimeMillis();
+                if (time - _lineNumbersRefreshTime > 125) {
```
harshad1, line 148: Why a time based vs a scroll based refresh like we do now?

guanglinn: Add a time based way just reduces the refresh rate of the line numbers when scrolling

harshad1: But we already have a system where the line numbering only redraws after the user has scrolled by some delta (1 page iirc). IMO that is superior

app/src/main/java/net/gsantner/markor/frontend/textview/LineNumbersTextView.java
```diff
@@ -0,0 +1,338 @@
...
+        this.scrollView = scrollView;
+        lineNumbersDrawer = new LineNumbersDrawer(editText, this);
+    }
+
+    public void setLineNumbersEnabled(final boolean enabled) {
+        lineNumbersEnabled = enabled;
+        if (lineNumbersEnabled) {
+            lineNumbersDrawer.prepare();
+            if (scrollView != null) {
+                scrollView.getViewTreeObserver().addOnScrollChangedListener(onScrollChangedListener);
```
harshad1, line 81: I think you can do `getViewTreeObserver().addOnScrollChangedListener` on the EditText, so you don't have to get and pass _primaryScrollView from the DocumentEditAndViewFragment

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
guanglinn: That system is called passively when the view is changed, it works well when the view is EditText, but now this view is TextView. So this code is designed to actively trigger that system to check and determine whether to draw new line numbers.

gsantner: How well does it work on bigger files (old and new)? Redrawing partial should work better, isn't it?

guanglinn: Yes, I think it's better than before.

https://github.com/user-attachments/assets/ca05b621-8298-4823-a3d3-0682ccd54a07/

gsantner (approved):
Thank you!

Merging

PR #2481, opened 2024-12-08
Restore audio recording 

harshad1:
In #2468, audio recording with the stock built in audio recorder was also removed. In this PR I restore recording audio with the built in audio recorder.


harshad1:
Good catch. Apparently we don't need the permission. Dropped it.

gsantner (approved):
Thanks

PR #2504, opened 2025-01-09
Orgmode fixes and improvements, closes #2405

### LINKED ISSUE #2405: Org Mode check list button inserts lowercase X

- [X] The bug is still present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor settings are not touched. In case the issue is resolved there, you don't need to create a bug report. The change will be part of the next Markor update.)

### Description

Tapping the Org Mode "Check list" button with an empty check box inserts lower-case x, when it should insert uppercase X

i.e. [ ] Item => [x] Item, rather than [X] Item

Both emacs and the markor view mode for org mode seem to expect uppercase for correct behaviour.

### Steps to reproduce

1. Create an Org Mode file and add "Item" as first line
2. Click check list button => - [ ] Item
3. Click check list button again => - [x] Item
4. Click view button
5. Click pencil button and change to - [X] Item
6. Click view button

### Information

Android version: 
Device: 
App Version: 

### Source

F-Droid

### Format / File type

Not specific

### Additional info / Log

```shell
-
```

nyxkn:
Adding ReplacePatternGenerator file for orgmode.
We need this to set more correct patterns for orgmode lists and headings, which are different to markdown ones.

Fixes action buttons behaviour for ordered/unordered/checkbox lists, which was previously incorrect.
Fixes checkbox list to only use uppercase X as per orgmode specs (closes #2405 )
Fixes newlines autoindent. It now follow the correct pattern for lists, instead of following markdown patterns
Fixes headings being bolded from level 3 onwards, i.e. `*** heading` was having `***` become bold since it's the same delimiter for bold, but we don't want that. More generically, we prevent emphasizing a pattern if it's made entirely of the delimiter character, e.g. *** or ///
Fixes heading being highlighted across newlines. i.e. `* \n heading` was being treated as heading, but only whitespace should be allowed between asterisks and heading text, not newline.

We're also adding heading action buttons.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/FormatRegistry.java
+++ b/app/src/main/java/net/gsantner/markor/format/FormatRegistry.java
@@ -29,6 +29,7 @@ import net.gsantner.markor.format.markdown.MarkdownReplacePatternGenerator;
 import net.gsantner.markor.format.markdown.MarkdownSyntaxHighlighter;
 import net.gsantner.markor.format.markdown.MarkdownTextConverter;
 import net.gsantner.markor.format.orgmode.OrgmodeActionButtons;
+import net.gsantner.markor.format.orgmode.OrgmodeReplacePatternGenerator;
 import net.gsantner.markor.format.orgmode.OrgmodeSyntaxHighlighter;
 import net.gsantner.markor.format.orgmode.OrgmodeTextConverter;
 import net.gsantner.markor.format.plaintext.PlaintextActionButtons;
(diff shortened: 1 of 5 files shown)


gsantner (approved):
Thank you for the changes!

PR #2566, opened 2025-04-20
Improving view mode and rendering

harshad1:
* Can now set view mode font size in view mode
* Much faster todo text highlighting
* Getting AppSettings with a proper context where ever possible (using applicationContext will cause theme to resolve incorrectly)
* Moved case change to an action
* Properly close the keyboard when changing to file browser


app/src/main/java/net/gsantner/markor/frontend/AttachLinkOrFileDialog.java
```diff
@@ -462,6 +464,20 @@ public void onFsViewerSelected(final String request, final File file, final Inte
...
+
+                            if (action == InsertType.LINK_BROWSE) {
+                                dopt.neutralButtonText = R.string.folder;
+                            }
+
+                            _dopt = dopt;
+                        }
+
+                        @Override
+                        public void onFsViewerNeutralButtonPressed(File currentFolder) {
```
harshad1, line 476: Select folder

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
```diff
@@ -76,6 +76,8 @@ public class DocumentEditAndViewFragment extends MarkorBaseFragment implements F
     public static final String SAVESTATE_DOCUMENT = "DOCUMENT";
     public static final String START_PREVIEW = "START_PREVIEW";
 
+    public static float VIEW_FONT_SCALE = 100f / 15.7f;
```
gsantner, line 79: Thanks!

Do you see that scale still fitting quite good?

harshad1: Looks good to me. But I do all my testing just my own primary device - a pixel 7 pro

app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtSyntaxHighlighter.java
```diff
@@ -23,54 +24,71 @@
...
+        _paragraphSpan = new ParagraphDividerSpan(paint);
+
+        return this;
     }
 
     @Override
     public void generateSpans() {
 
-        super.generateSpans();
+        // Single span for the whole text - highly performant
```
harshad1, line 46: Single span for the whole file. 

Much faster! Much less highlighting jank!

app/src/main/java/net/gsantner/markor/ApplicationObject.java
```diff
@@ -30,15 +30,15 @@ public static ApplicationObject get() {
         return _app;
     }
 
-    public static AppSettings settings() {
+    public static AppSettings backupSettings() {
```
harshad1, line 33: Making it clear that these settings are to be used as a last resort

app/src/main/java/net/gsantner/opoc/frontend/base/GsActivityBase.java
```diff
@@ -63,19 +66,9 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
         }
     }
 
-    protected void onPreCreate(@Nullable Bundle savedInstanceState) {
```
harshad1, line 66: Just moved to onCreate()

Children ought to call this first

app/src/main/java/net/gsantner/markor/frontend/textview/HighlightingEditor.java
```diff
@@ -293,6 +290,25 @@ private int rowEnd(final int y) {
         return layout == null ? 0 : layout.getLineEnd(layout.getLineForVertical(y));
     }
 
+    // Additional selections for search / replace etc
+    // ---------------------------------------------------------------------------------------------
+
+    private final List<SyntaxHighlighterBase.SpanGroup> _selections = new ArrayList<>();
+
+    public void addAdditionalSelection(final int start, final int end, final @ColorInt int color) {
```
harshad1, line 298: @guanglinn Here is an API to add highlighting to the Highlighter. Haven't tested it. May need some fixes / improvement.

guanglinn: Thank you, I will use it in the next commit.

gsantner (approved):
Thank you, I checked the changes and tried it. It's noticeable faster.

Thank you very much! Merging

PR #2597, opened 2025-06-12
Some fixes for AsciiDoc

JSchmerling:
Dark mode fixes:
- tables have now the same size in dark and light mode
- caution-admonition has now the right color
- the color of the font of table head and body is now swapped to optimize the readability

ActionButton fixes:
- one line instead of two when the button was clicked
- one page break instead of two when the button was clicked
- the quote button works now


gsantner:
Thank you! merging it

gsantner (approved):

PR #2603, opened 2025-06-22
Draft: new_file_dialog

JSchmerling:
This is early-stage draft for the new file dialog. Right now the functionality is stripped away. 
Deprecated stuff like RelativeLayout was thrown out and the dialog was simplified (no duplication of format and type). 

Feedback is welcome

### DIFF
--- a/app/src/main/res/layout/new_file_dialog.xml
+++ b/app/src/main/res/layout/new_file_dialog.xml
@@ -46,34 +46,16 @@
                     android:text="@string/name"
                     android:textAppearance="@style/TextAppearance.AppCompat.Caption"/>
 
-                <FrameLayout
+                <AutoCompleteTextView
+                    android:id="@+id/new_file_dialog__name"
                     android:layout_width="match_parent"
-                    android:layout_height="wrap_content">
-
-                    <EditText
-                        android:id="@+id/new_file_dialog__name"
-                        android:layout_width="match_parent"
-                        android:layout_height="48dp"
-                        android:layout_gravity="center_vertical"
-                        android:hint="@string/title"
-                        android:importantForAutofill="no"
-                        android:lines="1"
-                        android:textAppearance="@style/TextAppearance.AppCompat.Caption"
-                        android:textSize="15sp"
-                        tools:ignore="HardcodedText,TextFields"/>
-
-                    <TextView
-                        android:id="@+id/new_file_dialog__name_format_spinner"
-                        style="@style/Base.Widget.AppCompat.Spinner"
-                        android:layout_width="48dp"
-                        android:layout_height="48dp"
-                        android:layout_gravity="end"
-                        android:contentDescription="@string/choose_format"
-                        android:inputType="none"
-                        android:paddingEnd="0dp"
-                        android:paddingStart="24dp"/>
-
-                </FrameLayout>
+                    android:layout_height="48dp"
+                    android:layout_gravity="center_vertical"
+                    android:hint="@string/title"
+                    android:lines="1"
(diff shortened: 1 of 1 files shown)


harshad1:
What do you mean by 

> no duplication of format and type

Type is the file type (markdown, asciidoc etc)
Format is a template for the title (so you can insert a date or UUID etc)

JSchmerling:
I meant type and file extension.

gsantner:
Not enterly sure either, one is for filename (.txt) the other how to treat the file (.txt can be a todo.txt, plaintext without any format specific features, ...)

JSchmerling:
Well, in the vast majority of cases the type can be derived from the file extension. And this would initially only be the type with which the file would be created.

You could then proceed according to the following grid:
.md, .markdown, ... .md -> MarkDown
.adoc,, .asciidoc -> AsciiDoc
.org -> Orgmode
.csv -> CSV
.wiki -> Wikitext
.todo.txt -> ToDo
Default

gsantner:
zim/wikitext is .txt too normally

JSchmerling:
But then you could add `.txt (wikitext)` as an option in the dropdown menu. Then a .txt file would be created but with the type Zim/Wikitext

JSchmerling:
This is what the dropdown menu for the file extension than would look like:

- `.md (MarkDown)`
- `.adoc (AsciiDoc)`
- `.org (OrgMode)`
- `.csv (CSV)`
- `.txt (Text)`
- `.txt (Zim/Wikitext)`
- `.todo.txt (ToDo)`

gsantner:
To me would sound fine, but you can expect already that users will ask if they can use .markdown for example and still use Markdown format, and if there is a settings option for that to use it. Or how to use a custom file extension with no filetype.
Currently you can just put your mouse there and write any filetype you like.

gsantner:
To me personally would sound fine as I use the default extension for the formats, but you can expect already that users will ask if they can use .txt for example and still use Markdown format, and if there is a settings option for that to use it. Or how to use a custom file extension with no Format.
Currently you can just put your mouse there and write any filetype you like.

Merging filename excluding ext + ext to one field (filename with ext), and changing the text there when changing the selection is a possibility. But having only these options and no way to type a custom file extension is bad.

What we also have currently is to remember the last used file extension + format combo, also something that is to be gone with this.

I don't see hiding the extension is the better way. For sure we can make it better, but don't think this one is it (usage wise, changing up layouts is performance wise sure better).

JSchmerling:
Perhaps there has been a slight misunderstanding:
- The merging of file extension and file type is only intended for the new_file_dialog
- The dropdown menu for the file extensions should only give a selection of supported file extension+types (as before)

Nevertheless, for the time being I will only modernize the existing dialog and make it more accessible.

harshad1:
The intended behavior is that you generally just set type and the extension is updated. Users don't have to touch the extension edit box unless they need to. But it is there if they need it.

If you want a cleaner system here is my suggestion:

- Remove the extension edit box
- Render the full file name (red if it is taken) as the user edits the title (include format and extension)
- if the user types in an extension then use that (we have routines to extract extension)

For example:

Type: markdown, filename: foo = foo.md
Type: markdown, filename: foo.txt = foo.txt (type still set to markdown on creation)

This is potentially cleaner but still gives the user full flexibility

JSchmerling:
> * Remove the extension edit box
> * Render the full file name (red if it is taken) as the user edits the title (include format and extension)
> * if the user types in an extension then use that (we have routines to extract extension)

Good idea. I'll make it on the next iteration.

gsantner:
Hello, did you intentionally close this?

PR #2666, opened 2025-10-11
Fix: Connect 'Prefer view' setting to actual file opening behavior

ullas-n-arc:
# Make "Prefer view" setting functional for Markdown files

## Problem Solved
The **"Prefer view"** setting existed in Markor's preferences but was completely non-functional. Users who enabled this setting expecting Markdown files to open in preview mode by default were disappointed when files continued opening in edit mode, requiring manual switching every time.

## 🔧 What This PR Fixes
This PR makes the existing "Prefer view" preference actually work as intended:

### Before (Broken):
- "Prefer view" setting existed but was ignored
- All files opened in edit mode regardless of user preference
- Required extra tap to switch to preview for every file
- Poor UX for users who primarily read/view Markdown files

### After (Fixed):
- "Prefer view" setting is properly respected
- Markdown files open in preview mode when setting is enabled
- Works for both existing files AND newly created files
- Improved UX for users who use Markor as a viewer

### Files Modified:
- **`DocumentActivity.java`**: Added missing logic to check `isPreferViewMode()` when deciding file opening mode
- **`MainActivity.java`**: Extended functionality to respect preference for newly created Markdown files

### Key Implementation Details:
```java
// NEW: Check user preference for preview mode
else if (_appSettings.isPreferViewMode()) {
    startInPreview = true;
    Log.d("DocumentActivity", "Opening in preview mode: user preference enabled");
} 


// NEW: Helper method to detect Markdown files
private boolean isMarkdownFile(final File file) {
    final String name = file.getName().toLowerCase();
    return name.endsWith(".md") || name.endsWith(".markdown") || 
           name.endsWith(".mdown") || name.endsWith(".mkd");
}
```
## Impact & User Benefit
- Fixes broken functionality that users expected to work
- Improves accessibility for users who primarily view rather than edit
- Reduces friction by eliminating extra taps for preview mode
## Testing Instructions
1. Enable Settings → View mode → Prefer view
2. Open any existing `.md` file → Should open in preview mode
3. Create a new `.md` file → Should also open in preview mode
4. Disable the setting → Files should open in edit mode (default behavior)


gsantner (changes requested):
Changes of this PR will only work / further reduce it down to only work in Markdown - however theres more than Markdown where View-Mode is supported.

harshad1:
What is the objective here? i.e. the desired behaviour?

The current behavior is:
1. New files will respect the main setting
2. Files will remember the last used setting

A small but significant change we could have:
Files will remember the last used setting iff the view mode has been explicitly changed

In the current behavior:
1. Set default mode to edit
2. Create new file
3. Open file <- opens in edit mode
4. Set default mode to view
5. Open file <- opens in edit mode because it remembers the last open

After my proposed change:
1. Set default mode to edit
2. Create new file
3. Open file <- opens in edit mode
4. Set default mode to view
5. Open file <- opens in view mode because it was never explicitly changed
6. Change mode to view
7. Close file
8. Open file again <- opens in view mode only now because it was explicitly changed

This is an easy change to make. What does everyone think?

ullas-n-arc:
Great idea

harshad1:
I looked at the current preview behavior again and actually I think it is already what we want.

- If global preview mode is enabled files _always_ open in preview mode
- If global preview mode is disabled (i.e we default to edit mode) we open in the last used mode for that document (as seen when the document closed

Could you describe your requirement in a little more detail?

Issue #2697, opened 2025-11-25
VBA code is not being syntax-highlighted properly


xingbanyelan-51:
### ⚠️ This issue respects the following points: ⚠️

- [x] This is a **bug**. Not a [question or feature request](https://github.com/gsantner/markor/issues/new/choose).
- [x] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [x] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [x] The bug is still present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor settings are not touched. In case the issue is resolved there, you don't need to create a bug report. The change will be part of the next Markor update.)

### Description

VBA code is not being syntax-highlighted properly

![Image](https://github.com/user-attachments/assets/5cf7046e-9098-4428-96d5-a7dd2a42976d)
![Image](https://github.com/user-attachments/assets/ea2f7a60-7e0e-472c-bcb0-c31304e76f2d)
![Image](https://github.com/user-attachments/assets/32aa3c33-8fc9-4535-8d03-be9b0cbb4109)

### Steps to reproduce

1. Copy the VBA script into an MD (Markdown) file
2. Switch to Reading View (or Preview mode)
3. The VBA code appears only in black text — variables and keywords are not colored (no syntax highlighting)

### Information

Android version: 11
Device: vivo Y76s
App Version: 2.15.2

### Source

GitHub Releases

### Format / File type

Markdown

### Additional info / Log

```shell
Subject: Feature Request: Add Syntax Highlighting Support for VBA Code Blocks in Markdown Preview

Dear gsantner (or Markor development team),

I hope this message finds you well. First off, thank you for creating and maintaining Markor—it's an incredibly lightweight and versatile Markdown editor for Android, and I've been using it daily for note-taking and code snippets. The offline capabilities and clean interface are standout features.

I'm writing to report a small but noticeable limitation in the Markdown preview mode (Reading View) and suggest a practical solution. When embedding VBA (Visual Basic for Applications) scripts in fenced code blocks within an .md file—like this:



Sub Example()
[...]

PR #2698, opened 2025-11-28
View-Mode: Improve user experience of code blocks

guanglinn:
Improve user experience of code blocks in view mode:

-   Fix line numbers display error in long wrapping words
-   Auto-refresh line numbers when font size was changed
-   Switch line numbers is available
-   Switch wrap words is available
-   Tweak paddings
-   Add Visual Basic syntax highlight
-   Add Zig syntax highlight

#2697

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
@@ -104,7 +104,7 @@ public class DocumentEditAndViewFragment extends MarkorBaseFragment implements F
 
     private DraggableScrollbarScrollView _verticalScrollView;
     private HorizontalScrollView _horizontalScrollView;
-    private LineNumbersTextView _lineNumbersView;
+    private LineNumbersView _lineNumbersView;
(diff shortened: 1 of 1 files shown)


guanglinn:
<table>
    <tr>
        <td>2025-11-28-prism-line-numbers-layout-error</td>
        <td>2025-11-28-prism-line-numbers-layout-fixed</td>
    </tr>
</table>

**Left:** line numbers display errors in long wrapping words.
**Right:** fixed.

gsantner:
Are there many prismjs (official) definitions not in Markor?
I initially wanted to keep it slim and most important ones, but at this point seems like we have most of the available ones already.

guanglinn:
Yes, there are too many official supported languages. https://prismjs.com/download

I also think it may be already enough for the languages we have added. [app/thirdparty/assets/prism/components](https://github.com/gsantner/markor/tree/master/app%2Fthirdparty%2Fassets%2Fprism%2Fcomponents)

Fortunately, these files are very small, about 1.5KB per file.

gsantner (approved):
Thanks, I tried it and it works, though no noticeable difference here

PR #2703, opened 2025-12-06
More modern material UI

harshad1:
In this PR I have adjusted the colours and some UI elements in Markor to more closely match the material guidelines.

image
image
image
image

Per @gsantner's request I have reverted to using 'primaryColor' for the light mode top bar. Material design guidelines say that dark mode should have a dark bar and I also think this looks way better so I have kept this


harshad1:
In a subsequent PR, I may add an option to make the UI colours follow the editor colours. (so fully solarized will be possible)

app/src/main/res/drawable/popup_menu_background.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<inset xmlns:android="http://schemas.android.com/apk/res/android"
+    android:insetLeft="4dp"
+    android:insetRight="4dp"
+    android:insetTop="4dp"
+    android:insetBottom="4dp">
```
harshad1: Slightly rounded corners

gsantner:
Thank you, I will try to find time on sunday to review

harshad1:
@gsantner This is now ready for resting and review. 

All feedback is appreciated.

gsantner:
Hello @harshad1 

Thank you, overall the PR is fine and a good improvement!

However - 

in dark mode, please restore the blueish-black color at toolbar and bottom navigation.
Humbly said, also like the previous background color better [softer background/text, less straining to my eyes], but it's little difference.

Screenshot_20260108-044828_Marder-append


and in light mode, please restore
*  the soft white instead of the bright white
* dark system navigation
* same as for dark mode, blueish-black toolbar and bottom navigation

Screenshot_20260108-044903_Markor-append

I prefer Markor to stay and not to be pure white / black / gray based UI.

harshad1:
Thanks for the review. I'll make changes.

For reference I picked the lightest grey and the darkest white which are still considered compatible with the material design spec. Even with these changes Markor is still considerably softer than other apps I referred to (thunderbird, gmail etc etc). 

I will make it softer still.

harshad1:
This branch ended up being too hard to work in. I created a new branch with more streamlined changes. Opening a new PR with the same intent

Issue #2701, opened 2025-12-01
Cannot remove priority in todo.txt under language (中文,中国)


li-zyang:
### ⚠️ This issue respects the following points: ⚠️

- [x] This is a **bug**. Not a [question or feature request](https://github.com/gsantner/markor/issues/new/choose).
- [x] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [x] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [x] The bug is still present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor settings are not touched. In case the issue is resolved there, you don't need to create a bug report. The change will be part of the next Markor update.)

### Description

In todo.txt, when I remove the priority tag of a task using priority selector, it is replaced by `(无)` instead of being removed. Under other language including similar language (中文,臺灣) and (日本語) the whole tag is removed. If I change the priority tag from `(无)` to `(A)`, the program adds a new tag.

https://github.com/user-attachments/assets/fda27426-e6d5-4a3e-8ab7-0e1ed40424aa

### Steps to reproduce

1. Change language setting to "Chinese (中文,中国)" / "Chinese Simplified (中文,中国)"
2. Write arbitrary task in todo.txt
3. Add a priority
4. Remove that priority

### Information

Android version: any
Device: any
App Version: 2.15.2 (161) and 2.15.2-0027 (251112)

### Source

GitHub Releases

### Format / File type

todo.txt

### Additional info / Log

```shell
-
```

gsantner:
Is this OK in 2.16 @li-zyang ?

li-zyang:
It's solved. Thanks!

PR #2704, opened 2025-12-15
To-Do: Fix priority can not be removed when app language is Simplified Chinese

guanglinn:
Fix this issue #2701 - Priority can not be removed but set to **(无)** on priority **None** is selected when application language is Simplified Chinese.


guanglinn:
There is a more robust solution by @harshad1.
See [pull/2703#discussion_r2625655435](https://github.com/gsantner/markor/pull/2703#discussion_r2625655435)

PR #2709, opened 2025-12-22
View-Mode: Improve text search

guanglinn:
Add following features based on the simple text search in view mode,

-   Move to the previous match
-   Show number of matches
-   Show current index of active match

These functions make it a complete text search.


guanglinn:
## Before



## Now

https://github.com/user-attachments/assets/2323b7be-59b2-40ea-ac0b-de6de51307c2/

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
```diff
@@ -353,25 +357,88 @@ public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
...
+                        _webView.findNext(true);
+                        return true;
+                    }
+                    return false;
+                }
+            });
+
+            ViewGroup mSearchPlate = null;
+            try {
+                Field mSearchPlateField = SearchView.class.getDeclaredField("mSearchPlate");
```
harshad1: Could you add some comments for why reflection is needed here?

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
```diff
@@ -780,6 +690,86 @@ private void showHideActionBar() {
...
+        });
+
+        // Because SearchView doesn't provide a public API to add custom buttons
+        // We must get the searchPlate (the layout containing the text field and close button) from SearchView
+        // This approach is more robust than reflection
+        int searchPlateId = searchView.getContext().getResources().getIdentifier("android:id/search_plate", null, null);
+        ViewGroup searchPlate = searchView.findViewById(searchPlateId);
+        if (searchPlate == null) {
+            // Ensure that SearchView is always available even if getting searchPlate fails
+            _menuSearchViewForViewMode.setSubmitButtonEnabled(true);
```
guanglinn: Ensure that SearchView is always available even if getting searchPlate fails

PR #2713, opened 2026-01-03
Select date format for todo tasks

AlexZ005:
This PR adds a new setting for configuring date format used in 'To-Do' for newly created tasks.

The new setting provides three predefined date formats:

- yyyy-MM-dd
- yyyy-MM-dd HH:mm
- yyyy-MM-dd HH:mm:ss

These formats have space between date and time. When a task marked as complete, the same logic is applied to completed date/time. With this change, due date can include time component and sorting continues to work correctly.

These options can be adjusted if needed, depending on the maintainers’ preferences.

It would also be useful to add the ability to define a custom date format as an additional option. This is possible, but would require more changes, including rewrite of how the application handles date parsing (DATEF_YYYY_MM_DD and PT_DATE). Please add to a comment if such a modification would be appropritate as follow-up change.

This PR can be merged as is if the feature is considered useful and no further modifications are required.

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtActionButtons.java
+++ b/app/src/main/java/net/gsantner/markor/format/todotxt/TodoTxtActionButtons.java
@@ -359,14 +359,13 @@ public class TodoTxtActionButtons extends ActionButtonBase {
     }
 
     private static Calendar parseDateString(final String dateString, final Calendar fallback) {
-        final String dateFormat = TodoTxtTask.getDateFormat().toPattern();
-        if (dateString == null || dateString.length() != dateFormat.length()) {
+        if (dateString == null || dateString.length() != TodoTxtTask.DATEF_YYYY_MM_DD_LEN) {
             return fallback;
         }
 
         try {
             Calendar calendar = Calendar.getInstance();
-            calendar.setTime(TodoTxtTask.getDateFormat().parse(dateString));
+            calendar.setTime(TodoTxtTask.DATEF_YYYY_MM_DD.parse(dateString));
             return calendar;
         } catch (ParseException e) {
             return fallback;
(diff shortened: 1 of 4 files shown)


harshad1:
Do we really need this? We have a snippet engine with time support and a dedicated time dialog too

You can define a template as

```
`HH:mm` {{sel}} @inbox
 ```
 
* here it has a default context
* The cursor will start at {{sel}}

Then you simply create a new task and insert the snippet to get your chosen task format

You can also make something similar your 'share into' format template. Mine inserts a full iso date time.

AlexZ005:
Thanks for the explanation, I have tried your suggestion, it works.

However, this PR adds time automatically at the moment a new line/task is created, using the configured date format. Currently, it’s not possible to change the built-in date format that is inserted automatically on a new line — snippets require an explicit user action and don’t affect the default behavior.

So this change is about:

- Making the default date configurable
- Avoiding manual snippet insertion for every task
- Keeping task and due dates sortable

Snippets and templates are still very powerful, but they don’t replace the need for configuring the app’s automatic date format, which is what this PR focuses on. To make it even more customizable, it would be useful to have an option for set 'custom date format', similar to your template suggestion (instead of predefined ones as in this PR), but this type of change would require rework on how app handles date parsing.

gsantner:
It sounds more like you are after to customize not the normal todo.txt date, but what comes after it. At todo.txt the format is simple, and it always is either yyyy-mm-dd, or nothing when you want to leave it out. 
Making this format customizable is a bad idea as it can break other functionality.

AlexZ005:
Thanks for clarifying, that makes sense, and I agree that the core todo.txt date format should remain fixed to avoid requiring changes across many related components.

To clarify my intent: the original goal was to make the todo.txt date format configurable at its source by allowing custom patterns via [SimpleDateFormat](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html), as it initially appeared to be a simple text value added to each new line. However, after working with the code, it became clear that more functionality is tied to this format than initially expected. As a result, I tried to start with the least disruptive approach by introducing a small set of predefined patterns.

Given this, I’m happy to close PR. It may still be possible to approach this in a different direction by adding the ability to include a time component in each task’s start or completion date (for example, `yyyy-MM-dd-HH-mm`), as this was the main functionality I was missing.

Thanks for the feedback and for taking the time to explain everything!

PR #2725, opened 2026-01-24
Cleaner UI update

harshad1:
This PR attempts to make the markor UI a little cleaner. Created a new branch with a more targeted set of changes (original here: https://github.com/gsantner/markor/pull/2703)

image
image
image

------
image
image
image

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
@@ -43,7 +43,6 @@ import android.widget.Toast;
 import androidx.annotation.NonNull;
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.appcompat.widget.SearchView;
-import androidx.core.content.ContextCompat;
 
 import net.gsantner.markor.ApplicationObject;
 import net.gsantner.markor.BuildConfig;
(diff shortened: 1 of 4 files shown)


harshad1:
Colors should mostly match the soft dark grey and soft white currently used for markor

gsantner:
Hello @harshad1 ,
sorry that I didn't have too much time for the existing PR. We can reopen it/take look again when this is merged, assume it's a smaller PR in amount of changes then.

Thanks, I installed it on my device and what I saw it looks like improvements.

Only one I noticed, that there is no divider anymore in the filemanager. Could you please enable it again? To me this results in the whole page being read like a text editor, with no separation between the items 😄 .

harshad1:
@gsantner Np. Everyone is busy :). The new PR is because the old one got a little messy with many changes.

No dividers is a big part of the visual modernization. Most apps do not use dividers any more (android built in files app for example). I have increased file item spacing to compensate and make the rows more distinct.

Options:
1. Increase item spacing a little more
2. Add an option to hide dividers (default on)

You'll also note that I have removed dividers in most other places (dialogs, share into fragment). Let me know if you want those handled

Thanks

gsantner:
> No dividers is a big part of the visual modernization. 

To me looks fine on the ost part except the file lists

> Most apps do not use dividers any more (android built in files app for example). 

It's OK to be different 😃 

> I have increased file item spacing to compensate and make the rows more distinct.

Less space for items = users complain that they need to scroll more, theres two sides

So for the non file listings, it's seems fine without dividers, but can it be enabled for files again?

PR #2727, opened 2026-01-26
Add option to toggle autosave

micaelmaciel:
PR to enable the option to toggle autosave. It will be on by default

### DIFF
--- a/app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
+++ b/app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
@@ -276,7 +276,10 @@ public class DocumentEditAndViewFragment extends MarkorBaseFragment implements F
 
     @Override
     public void onPause() {
-        saveDocument(false);
+        if(_appSettings.isAutoSaveEnabled()) {
+            saveDocument(false);
+        }
+
         if (_webView != null) {
             _webView.onPause();
         }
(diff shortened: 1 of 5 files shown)


gsantner:
Hello,
thank you, but I currently don't think about merging it.

micaelmaciel:
No problem. May I know why?

gsantner:
You swipe out, run into low ram, or Android decides in background that it could close the app.. and your changes are gone. Markor tries very hard to find a good saving solution, this just adds another condition to the big amount of rules.

micaelmaciel:
Oh, but it's optional. I'm sure most users wouldn't disable it because they agree with you. I wanted an option like this mainly because of the quicknote feature. I like to make changes quickly to reference something I'm currently dealing with and I like to keep my notes clean, so I didn't want to keep deleting lines every time I make changes. You may see it as a nitpick, but I saw there were some issues open to this idea. 

You think there's a specific way to implement this that you'd like or is it the feature itself you really don't want?

harshad1:
If you want to not save changes you can reload from disk (in the menu) before exiting.

guanglinn:
According to Android rules, when the user slides Markor out to the background, the changes should be automatically saved.
But when the user closes the note and return to the note directory (still in the app), perhaps it can prompt whether to save your changes (as an optional option).

@micaelmaciel Maybe you can make such a compromise solution.

micaelmaciel:
I don't think I'll have the time to do that, as I'm not that experienced with android development and I'm working on something else at the moment.

Since this is very much a me thing, I'll just keep my fork synched with the upstream repo. Thank you all for considering it

Issue #2681, opened 2025-10-29
Asciidoc highlight hide blinking text cursor


faustaahmad:
### ⚠️ This issue respects the following points: ⚠️

- [x] This is a **bug**. Not a [question or feature request](https://github.com/gsantner/markor/issues/new/choose).
- [x] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [x] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [x] The bug is still present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor settings are not touched. In case the issue is resolved there, you don't need to create a bug report. The change will be part of the next Markor update.)

### Description

![Image](https://github.com/user-attachments/assets/98876892-f664-461f-b239-90533343b8c0)
![Image](https://github.com/user-attachments/assets/453ff5bd-2c5a-480f-a8e4-6a52dfd7877a)

### Steps to reproduce

1. create asciidoc file
2. write something that enclosed with `[` and `]`
3. move text cursor inside

### Information

Android version: 11
Device: arm7a
App Version: 2.15.2

### Source

GitHub Releases

### Format / File type

Something else (-> Additional info)

### Additional info / Log

```shell
Asciidoc
```
Please add this file format into issue template too
~~fix `include::file.adoc` in view mode if possible~~

faustaahmad:
I thought im blind because not able to see the blinking text cursor since I have bit of partial blindness to red color

guanglinn:
There was a similar issue #2183 fixed in Markdown.
The reason is that the cursor is covered by EditText BackgroundColorSpan.

PR #2730, opened 2026-02-04
Fix AsciiDoc cursor being hidden and add support for copying links in view-mode

guanglinn:
Fix that issue #2681 - **Asciidoc highlight hide blinking text cursor**.

### Before

asciidoc_cursor_hidden

### Now

Screenshot_20260204_225341_net gsantner markor_test

Implement another feature from issue #1788 - **Make it possible to copy links in reading mode**.

https://github.com/user-attachments/assets/a766b365-e711-4d32-97d0-4382a227b539


guanglinn:
Okay, I have pushed new code.

PR #2745, opened 2026-02-25
Slovenian translation

janezpavelzebovec:
This adds Slovenian translation of the app.


gsantner:
Thanks, note that translations are done on translation platform Crowdin. Can be translated there using GitHub account too.
https://crowdin.com/project/markor

gsantner:
image

PR #2755, opened 2026-03-08
Choose todo archive based on todo name

kw217:
This makes it easier to use multiple .todo.txt files, by making the archive function pick a name for the archive based on the name of the todo file. If the todo file is `foo.todo.txt`, the proposed archive will be `.foo.todo.archive.txt`. For the default `todo.txt` it is `.todo.archive.txt`. This is just the default in the dialog - it is still possible to choose any name you like. The name is still remembered for each file.

The suggested default starts with a `.` to avoid clutter - this hides the archive file. I considered adding a setting to allow you to change this behaviour, but it seemed unnecessary clutter. Let me know if you think that would be useful and I will add it.

Let me know if there are any documents or tests I should update.

Thanks!


kw217:
Hi @gsantner - is this a reasonable approach? Do you need me to do anything before you review? Thanks.

gsantner:
Hello @kw217 ,
thank you!

I currently don't think about changing the default filename, even not to make the archive hidden by default (dotfile). ToDo is a prominent part of Markor and making it suddenly split off, without user explicitly changing something, isn't a good idea.

While you can have multiple todo.txt files - and this is just fine - majority use of todo.txt usage I see (and also at Markor) as >the one current todo list<.

kw217:
Interesting! I have lots of todo lists, for different categories - my main todo list, but also my shopping list (so I only see shopping items when I'm in the shop), my music to listen to list (when I want to choose what to listen to next), my DIY repairs list, etc. I was inspired in this by _Getting Things Done_ by David Allen.

If you only use one todo file, then having some features that apply when you have multiple todo files wouldn't impact your experience at all, but would make this kind of usage more convenient.

I can see that changing the name of the main archive and hiding it would be a bad idea, so I can certainly change the PR to leave that alone.

Would you consider accepting this, if it doesn't affect your mainline usage?

PR #2764, opened 2026-03-10
Fix invalid resource causing the app to be unable to open links in custom tabs.

xn-7492:
- Escape double quotes in build.gradle to avoid syntax error in BuildConfig.
- Fix invalid resource causing the app to be unable to open links in custom tabs.


gsantner:
I did similar fix today, but didn't notice this PR 😄

Issue #2716, opened 2026-01-06
Performance degradation from 2.14.1 to 2.15.2


pgorod:
### ⚠️ This issue respects the following points: ⚠️

- [x] This is a **bug**. Not a [question or feature request](https://github.com/gsantner/markor/issues/new/choose).
- [x] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [x] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [x] The bug is still present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor settings are not touched. In case the issue is resolved there, you don't need to create a bug report. The change will be part of the next Markor update.)

### Description

Using a big markdown file, I notice a significant performance degradation when I upgraded Markor from 2.14.1 (normal release) to 2.15.2 (using Marder, latest debug version).

I use a big file everyday, but I can't post it here, so I created a new sample one from a free eBook, which I tried to give the same characteristics as my file (size, number of lines, of headings, kind of markdown used). 

My file shows an even greater performance penalty, but I don't know exactly why and the sample seems enough to reproduce the problem:

[Chesterton.md](https://github.com/user-attachments/files/24453649/Chesterton.md)

### What sort of delay am I seeing?

I am using a OnePlus Nord CE 5G with 8GB of RAM which generally feels fast and snappy.

Syntax highlighting is turned on in all tests.
Turning it off doesn't seem to make a very noticeable difference.

- opening the file takes around 1.5s before upgrade, 3s or 4s after
- opening/closing the keyboard goes from almost immediate to noticeable (1 or 1.5s)
- navigating to a chapter by tapping on the top bar and selecting: no apparent performance hit

### Steps to reproduce

1. Open the file
2. Open keyboard or close it
3. Contrast delays in both versions

### Information

Android version: 11
Device:  OnePlus Nord CE 5G / 8G RAM / SnapDragon 750G
App Version: 2.14.1 / 2.15.2

### Source

Github apk

### Format / File type

Markdown

harshad1:
Three things:

1. Marder is a debug build and will be slower
[...]

Issue #2652, opened 2025-09-24
v2.15: App crashes and very slow while you edit a file


CREBOSO101:
### ⚠️ This issue respects the following points: ⚠️

- [x] This is a **bug**. Not a [question or feature request](https://github.com/gsantner/markor/issues/new/choose).
- [x] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.
- [x] Markor **is** up to date. See [Releases](https://github.com/gsantner/markor/tags) for the latest version. Updates are available from [F-Droid](https://f-droid.org/en/packages/net.gsantner.markor/) and GitHub.
- [x] The bug is still present in the latest development version (git master). (Please [download](https://nightly.link/gsantner/markor/workflows/build-android-project/master) and try the test version of Markor, named **Marder**. Don't worry; Markor and Marder appear as completely separate applications. You can install both side-by-side, and Markor settings are not touched. In case the issue is resolved there, you don't need to create a bug report. The change will be part of the next Markor update.)

### Description

When entering the .txt file that I usually edit the crashing is evident, when clicking on the screen it takes for the keyboard to open and also while typing locks occur frequently, not crashes of the type of closing the APP but it is as if the "fps" fall to zero if you understand me. My file have 172kb size.

This only occurs in versions after 2.14.1, such as 2.15.0 and 2.15.1, either from Github or F-Droid (I only test these so I don't know if it changes anything in the Play Store version).

It is a very serious problem that makes using the app impossible for me, I am having to use version 2.14.1 to edit my file.

It is basically the same thing that occurred in Issue #2478 

### Steps to reproduce

1. I think just opening and editing any file with many (MANY) characters. In my case I have a personal journal in which I write down everything I do or think during the day, then the file is a very extensive encrypted .txt.jenc

### Information

Android version: Android 13
Device: Xiaomi Poco X6 MIUI 14
App Version: 1.15.0 and 1.15.1, Github or F-Droid. Any version prior to this is without this problem.

### Source

GitHub Releases

### Format / File type

Not specific

### Additional info / Log

```shell
- My file only has pure text. And there are many, many characters
```

licaon-kter:
Confirmed slow downs here, F-Droid, Android 14

sivaraam:
Confirmed slowdown here too. Markor v2.15.1 on Android 12. File size: 101KB

guni77:
[...]

PR #2521, opened 2025-02-06
Quick filter all files, Share files into markor

harshad1:
This PR adds 2 new features:
1.  a quick filter - search for all files in the notebook by tapping on the title.
2. Share files into markor to attach them to documents. From download folder for example. Files will be copied into notebook like the usual file attachment system

Also included are a ton of fixes and improvements:

- File browser sorting and loading (especially special folders)
- Improved delete dialog
- Speed improvements (2-3x) for the search (on my device at least)
- undo / redo will track file load

PR #2769, opened 2026-03-14
fix: stabilize editor layout sizing to prevent large-file open slowdown

### LINKED ISSUE #2739: Creating new line in large files (>100 KB, 5000 lines) becomes increasingly slow.

- [x] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.

### Description

When dealing with large text files, with more than 100 KB and 5000 lines, a noticeable delay occurs when creating new lines. Other characters do not have delay, 

### Steps to reproduce

1. Create an artifically large .md file with many lines by copying something over and over again. The bigger, the more noticeable the delay will be.
2. Open this .md file with markor
3. Try to add a new line and notice a delay.

### Information

Android version: 16
Device: Samsung Galaxy
App Version: 2.15.2

### Source

F-Droid

### Format / File type

Markdown

### Additional info / Log

```shell
-
```

lrq3000:
## Summary
- Fixed a major speed performance regression on startup of opening large files and opening the keyboard in such large files, which was introduced by layout changes in PR #2521 (isolated via bisect to commit [64ff85b6](https://github.com/gsantner/markor/commit/64ff85b6) -- commit description "Better layout") merged in between v2.14.1 and v2.15.0.
- Root cause was layout-pass churn during editor initialization (unstable height sizing and repeated relayout pressure on large text/highlighting startup path).
- Restored stable one-shot min-height initialization and adjusted editor layout params to avoid reintroducing expensive startup relayouts.

## Details
- `DocumentEditAndViewFragment` now uses one-shot min-height synchronization during startup instead of persistent per-layout updates.
- Added concise guardrail comments in the hot path to explicitly prevent future reintroduction of layout-listener-based min-height updates.
- `document__fragment__edit.xml` editor sizing was adjusted to align with the stable pre-regression layout behavior.

## Related issues
Fixes #2739, #2716, #2652.

##  Test Build
The test build (Marder) is available here: https://github.com/lrq3000/markor/actions/runs/23170909092

## Authorship
This was made with an AI agentic co-development using git-bisect and OpenCode with ChatGPT Codex-5.3

## Additional comments
Please let me know if the fix is alright for you guys, I don't know the downstream consequences of these changes since I don't know the codebase.

I would suggest to add an automated performance test in the ci to ensure there is no more speed regression in the future. I can try to implement one as I was the one implementing these speed regression tests in the `tqdm` python module, so I have some experience on how to do that reliably (using AST decomposition instead of a timer, because otherwise the tests will randomly fail depending on the CI cloud server usage and that's very bothersome). But I am not sure this is possible in Android, but I can try. Just let me know if you would be interested in adding this to the project if my endeavor is successful.

Note that in any case, I would work on the automated performance unit test as a separate PR, because I know from experience that this is a very complex kind of unit test, very time consuming to make robustly, and so it's better in a separate PR (and I may not have the time to ever tackle it so the current fix should not depend on the performance unit test).

## Todo
* [x] Maintainer's review.
* [x] optionally, if #2770 is not going to be merged, show a toaster message when opening large files to notify the user that loading is underway and may take a few seconds, so the user knows the app is still working ([suggestion](https://github.com/gsantner/markor/issues/2716#issuecomment-4069012244) by @pgorod). -- UPDATE: Made as a follow-up PR in #2778 .


lrq3000:
I pushed a small update to get even closer to feature parity with the original intent in the 64ff85b6 commit that introduced the speed regression. Unfortunately, this commit's description is quite indescript ("Better layout"), so I used AI to try to infer the original intents and what features may be lost with the speed regression fix, here is the chatlog:

```
What `64ff85b6` likely intended
- Improve **editor layout consistency** in wrapped/unwrapped modes (especially when moving editor in/out of `HorizontalScrollView` in `setWrapState`).
- Make the editor occupy space more predictably using weighted params (`0dp + weight`) in `document__fragment__edit.xml`.
- Simplify bottom margin handling for the active edit container (`showHideActionBar` path).
- Remove visual jank from startup layout behavior (though the implementation chosen caused performance issues).

Does your current fix contradict that intent?
- **Mostly no.** Your fix removes the harmful part (persistent/per-layout min-height churn) but keeps the core UX behavior:
  - Wrap/unwrapped mode switching is still there in `app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java`.
  - Action-bar bottom spacing behavior remains.
  - Viewport fill for short docs is still handled via one-shot min-height sync.

----

what are the features we lose with this fix compared to the original intent in the commit 64ff85b6?

----

Very little is lost, and mostly in edge-cases.

Compared to original `64ff85b6` intent, what you potentially lose now:

- No continuous live min-height re-sync on every later container resize event.
  - Possible minor impact on rare cases like split-screen resize / keyboard timing / rotation after initial open.
- Initial wrapped layout no longer uses the earlier weighted `0dp` width experiment.
  - We kept the known-fast `match_parent` start config.

What you **keep** from the original intent:

- Wrap/unwrapped mode switching behavior.
- Editor reparenting into/out of horizontal scroll for no-wrap mode.
- Selection restore + smooth transition behavior.
- Action-bar bottom margin handling.
- Overall “better layout” UX, without the performance regression.

So the tradeoff is essentially:  
**we gave up continuous auto-resync behavior (which was costly) to keep startup/open performance stable.**

----

What does this imply in practice? What are scenarios, use cases, where we will see a difference, and what difference?

----

Great question. In practice, here’s where you may notice differences now:

- **Normal use (open/edit large files, toggle wrap, type, save):** no negative difference; this is where you gained the big speed win.
- **Keyboard show/hide after initial open:** mostly same; in rare devices you might see a bit more/less blank space below text until next explicit UI update.
- **Rotation / split-screen resize while the editor is already open:** this is the main edge case; height may not re-sync as “continuously” as before, so viewport fill can be slightly stale momentarily.
- **Very short documents:** occasionally you may see the editor area not perfectly “stretched” after dynamic container changes until another trigger (e.g. wrap toggle, action bar visibility change) runs.
- **Wrap toggle behavior itself:** still good, because we kept explicit one-shot resync on wrap transitions.

So the practical tradeoff is:
- **Before:** continuous auto-resync everywhere (but expensive, causing severe slowdown on big files).
- **Now:** explicit one-shot resync at key moments (fast and stable), with only rare post-resize visual edge cases.
```

I tested several of those edge cases, especially rotation after a large note is already opened in the editor, and I saw no noticeably slowdown compared to v2.14.1 nor UI discrepancies compared to v2.16.0, so in practice the "main edge case" is handled in my field tests.

lrq3000:
@gsantner Awesome, thank you so much for fast-tracking the merge of this PR!

Yes #2770 is riskier so I will do extensive testing myself before tagging it as ready, and even then I understand there will have to be a decision whether or not to merge it in. So far the progress is good and I think we may get a stable release with features parity for the large text editor soon-ish :-)

Also if you are open to getting donations, please let me know how you would like to proceed, you can also contact me at my e-mail LRQ3000 at gmail dot com :-) (PS: don't worry, my e-mail is already displayed on every commits messages so I don't care much at this point).

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
```diff
@@ -232,14 +232,9 @@ public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
...
-            final int height = _verticalScrollView.getHeight();
-            if (height > 0 && height != _hlEditor.getMinHeight()) {
-                _hlEditor.setMinHeight(height);
-            }
-        });
-        _verticalScrollView.addOnLayoutChangeListener((v, left, top, right, bottom, oldLeft, oldTop, oldRight, oldBottom) -> ensureMinHeight.run());
-        _verticalScrollView.post(ensureMinHeight);
+        // Keep this as a one-shot min-height sync. Do not use a persistent layout listener here,
+        // otherwise large documents trigger repeated relayout work during startup.
+        syncEditorMinHeightOnce(_verticalScrollView);
```
harshad1, line 237: what happens if the window is resized (i.e. in multi window mode)?

lrq3000: I did not test but it should work just like Markor v2.14.1 and earlier so it should be fine. I will give it a try.

PR #2778, opened 2026-03-19
feat(large-file): add toast notification when opening a large file

lrq3000:
Notifies the user that there may be a delay in UI responsiveness but the action was correctly registered.

The toast notification shows the file size and it displays when opening a file from the files browser, when switching to Quicknote, and when switching to Todo. The threshold is set empirically at 512 KiB.

Changes were kept as minimal as possible.

Follow-up to #2769 implementing the UX improvement suggested by @pgorod in #2716 [here](https://github.com/gsantner/markor/issues/2716#issuecomment-4069012244).

## Authorship
Stephen Karl Larroque

Using OpenCode with ChatGPT Codex-5.3 and plan mode


gsantner (approved):
Thank you for the improvement.

The amount of changes are low and obligatory for this case

gsantner:
Upon further using Markor with this change included, I need to say it's pretty ineffective currently:

For i.e. videos the popup shows up (EmbedBinary Format which not reads but makes \<video src=`. But it loads instant.

Then I had a not-small file which indeed takes a few seconds to load, there it not showed up.

lrq3000:
Oh ok thank you for the heads up, my apologies, I did not even know markor could embed videos, I'll pay attention to that for future PRs. Thank you for tackling these issues with your latest commits. Is there any remaining issue you would like me to take care of ? 

The problem with the threshold is that it is arbitrary and dependent on the phone, there is no really obvious way to dynamically size the large file threshold to ensure it does not show up unnecessarily or on the contrary that it shows up for medium sized files on phones with more restricted ressources. 

To definitely fix this issue, I think we would need to rewrite the loading function to use a thread to check how long it takes to loah in realtime and then show a toast message, but this seems overkill to me, so I decided to implemented a simple heuristic with a fixed threshold, at worse it does not show up when it should or it shows up despite the phone being fast enough to process larger files, but in any case it remains useful for files that are expected to be very long to load (think 15s+). But again in 10 years phones will likely be able to open files this large fast enough so the toast messqge will then show up all the time.

That's why I believe the true solution is through the bigtextedit editor in the long run.

gsantner:
Hello, thanks but no need to apologize 😄 . I just meant to inform that I quickly found bad cases, and I improved it.

> Is there any remaining issue you would like me to take care of ?

This topic specifically should be good for now. While it not generally fixes something or makes faster, it shows information to user.

> The problem with the threshold is that it is arbitrary and dependent on the phone, there is no really obvious way

I tried in past (`isDeviceGoodHardware()` function) to get a somewhat helpful distincation, but yes we can't really know constant statically at compile time. For the toast it's good enough.

As of central editing component, I am very unsure as majority of what makes Markor and it's features centers around the current EditText solution.

PR #2798, opened 2026-04-14
ViewMode: Add support for viewing link addresses

### LINKED ISSUE #2749: Add long-press menu for links in view mode
### Description

Would be very handy if long-pressing a link in view mode would yield a context popup menu with options to copy link, open the link (default behavior when tapping links), share, copy link text, and whatever other useful operations for links you can come up with.

### LINKED ISSUE #2800: Background highlighting of code spans omitted in list items
# IMPORTANT
* App version: 2.16.1
* Android version/rom: Lineage 22.2 for microG (based on Android 15)
* Device: Pixel 4

# Explain the Problem

Code spans (i.e. text delimited by backticks) lacks its distinctive background highlighting when appearing in list items. 

When similar code spans appear both in a regular paragraph and a nearby list, this omission makes code spans in lists less easily recognized as not regular text.

# Steps to Reproduce

```
My preferred pager is `less`.

- `less` is more advanced than `more`.
- Most systems' preinstalled packages include `less`.
- Otherwise, it easy to add with `apt` or `dnf`
```

guanglinn:
Most people just want to view the link address behind the link text instead of directly copying it. This feature supports long pressing the link text to view the link address and copy it.

Related discussion: #1788

### Screen recording

https://github.com/user-attachments/assets/77486950-dfe2-41cf-9e50-3d8b1e59a06a


app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
```diff
@@ -1029,14 +1031,19 @@ private void setupWebViewIfNeeded(final Activity activity) {
             _webViewClient = new MarkorWebViewClient(_webView, activity);
             _webView.setWebViewClient(_webViewClient);
 
-            // For copying link address to clipboard in view-mode
+            // For showing and copying link address in view-mode
             _webView.setOnLongClickListener(v -> {
                 WebView.HitTestResult hitResult = _webView.getHitTestResult();
```
harshad1, line 1036: IMO if something can be made final then it should be final

Liberal use of final makes the code easier to read imo

harshad1 (approved):

app/src/main/java/net/gsantner/markor/activity/DocumentEditAndViewFragment.java
guanglinn: Yes, I think so too. Not only security, `final` can also improve program performance in most cases

gsantner: I see final mostly as note to developer that the value won't change after first assignment (with little exception it being mandatory for passing certain variables to lambda etc).
As far I know and also searched a bit, adding final modifier not generally affects performance

gsantner:
Hello @guanglinn ,
thanks but it's having a merge conflict now due merging your other big PR 😄 
If it fits you, can you please take a look?

guanglinn:
Hello, @gsantner 
Now the conflict has been resolved, you can merge #2798

gsantner (approved):
Thanks, I tried and it worked fine!

I reverted that small word change, so it not resets the translation (and it's getting english everywhere again until someone translates)

PR #2808, opened 2026-04-24
Fix Wikitext (Zim wiki) multiline code block rendering

Self-Perfection:
Extract `{{{code: ...}}}` (Zim Source View plugin) and `'''...'''` (Zim standard preformatted) blocks before the per-line transformation so their content is not mangled by inline rules like IMAGE (`{{X}}` -> image embed). Extracted blocks are wrapped in Markdown fenced code blocks and re-inserted after the per-line pass.

Fixes: multiline code blocks containing `{{Variable}}` tokens being rendered as broken images and live HTML.

----

Tested by running added code in Java, separately from Markor. Relying on CI/CD in your repo for building APK for full integration test.

----

I've seen request to do auto-reformat. But I do not have Android studio. And for me generated code looks OKish? I suppose Claude is good enough in following conventions and this does not need passing through external tools for proper formatting?


Self-Perfection:
Tested APK with this file

[06_code_block_with_html.txt](https://github.com/user-attachments/files/27063318/06_code_block_with_html.txt)

```
Content-Type: text/x-zim-wiki
Wiki-Format: zim 0.6
Creation-Date: 2026-04-24T14:00:00+01:00

====== 06 code block with html ======

{{{code: id="Front Template" lang="html" linenumbers="False"
<span class="marker">🔄</span>
<button class="marker" popovertarget="Notes popup">🗒</button>
<dialog popover id="Notes popup">Some notes</dialog>
}}}
```

Works as expected!

gsantner (approved):
Thank you for the PR! Merging

PR #2823, opened 2026-05-13
GsFileBrowserFragment: auto-refresh on filesystem changes (FileObserver)

dsremo:
## Problem

When the user has the file browser open on a folder and an external process modifies that folder — another app saves a file, a sync tool (Syncthing, Nextcloud client, git pull) fetches new notes, the user moves files from a desktop file manager over MTP — the browser currently shows a stale listing until the user manually pull-to-refreshes or navigates away and back. This is especially jarring on a notes app where the expectation is that all the files in the folder are "live".

## Fix

Register a `FileObserver` on the current folder's absolute path in `onResume()`, unregister in `onPause()`.

Events watched: `CREATE | DELETE | MOVED_FROM | MOVED_TO | MODIFY` — anything that changes the listing.

When the observer fires, `_filesystemViewerAdapter.reloadCurrentFolder()` is dispatched to the existing main-thread handler with a **300 ms debounce** to coalesce bursts (e.g., a sync that touches 50 files reloads the view once, not 50 times).

## API compatibility

Uses `FileObserver(String, int)` — the API 1 constructor — rather than `FileObserver(File, int)` — the API 29 constructor — to stay compatible with Markor's `minSdk=18`.

## Lifecycle correctness

- Observer is only registered while the fragment is in the resumed state.
- `onPause()` always unregisters before the fragment becomes paused — no leaks across configuration changes or process death.
- `stopFolderObserver()` is also idempotent and called at the top of `startFolderObserver()` so navigation between folders cleanly tears down the previous watcher.

## Files touched

- `app/src/main/java/net/gsantner/opoc/frontend/filebrowser/GsFileBrowserFragment.java` (+44)

## Tested

Manual: opened the file browser on a folder, then from a separate `adb shell` did `touch newfile.md`, `rm oldfile.md`, `mv a.md b.md`. Each operation reflected in the browser within ~300 ms without any user action. Then verified rapid `for i in 1..20; do touch ${i}.md; done` produces a single reload, not 20.

## Notes

If you'd prefer this to be opt-in behind a setting (e.g., for users with very large folders where the kernel inotify watch is non-trivial), happy to add a toggle.


app/src/main/java/net/gsantner/opoc/frontend/filebrowser/GsFileBrowserFragment.java
```diff
@@ -351,6 +367,34 @@ public void onResume() {
...
+        stopFolderObserver();
+    }
+
+    private void startFolderObserver(final File folder) {
+        stopFolderObserver();
+        if (folder == null || !folder.isDirectory()) return;
+        _folderObserver = new FileObserver(folder.getAbsolutePath(), FOLDER_OBSERVER_MASK) {
+            @Override public void onEvent(int event, @Nullable String path) {
+                if (path == null) return;
+                _folderReloadHandler.removeCallbacks(_folderReloadRunnable);
```
harshad1: We have a system to debounce with a delay (search for makeDebounced)

harshad1:
This is very cool and remarkable simple. 

IMO we should try to move this into the adapter. That would simplify several things.

Will review and leave more comments if needed

gsantner:
Thanks for your PR!
Agree with harshad1 regarding debounce - if it's feasible, good to use.
Especially when selecting multiple files and deleting them at one go, you don't want to reload the list 50 times by 50 deleted file events.

dsremo:
Thanks both — pushed `9193e6a` addressing both points.

The watcher now lives in `GsFileBrowserListAdapter`. `rebindFolderObserver()` is called from `loadFolder()` only when `folderChanged` is true, so navigation and back-stack pops rebind the inotify watch automatically, and same-folder reloads (pull-to-refresh, the auto-refresh itself) don't churn it. `onDetachedFromRecyclerView` releases the watch.

Debounce uses `TextViewUtils.makeDebounced(300, this::reloadCurrentFolder)` — that also handles the multi-select-delete burst you mentioned, since `makeDebounced` posts on the main-thread `Handler` regardless of which `FileObserver` internal thread fires the event.

Virtual folders (Recent/Popular/Favourite/AppData/storage-emulated) and permission-denied scoped-storage paths skip the watcher via `isDirectory() && canRead()` — their listings are synthesised, so file-system events on them wouldn't be meaningful.

Side benefit: the fragment loses the `FileObserver` field, `Handler`/`Looper` plumbing, and the start/stop helpers entirely; `GsFileBrowserDialog` inherits the auto-refresh for free since it shares the adapter. Net -5 lines vs v1.

gsantner (approved):
Thank you @dsremo 

I tried it and it worked very well. This is a quite minor change with big effect.

I removed and created files using adb while filebrowser was open, as well selected things meanwhile - it didn't crash here.

Merging

PR #2825, opened 2026-05-16
Addition of Jetbrains Mono Font style

SireOmega:
It's your app. Just added Jetbrains Mono. One at the wrong side too (Home folder). Please fix that, I am on Android. Sorry for inconvenience


gsantner:
Hello, whats the liecense of it?

SireOmega:
> Hello, whats the liecense of it?

Apache 2.0 with copyrights in your name. This is just a fork.

gsantner:
Not Markor, the font. Adding it may require license attribution.

SireOmega:
> Not Markor, the font. Adding it may require license attribution.

If that's the case. I request for a feature where the user can upload their preferred font to the app through File Manager. This way you can avoid license issues.

gsantner:
.app/fonts folder. already supported.

SireOmega:
> .app/fonts folder. already supported.

It is true and it was my solution to this. But I would like to request in-app UI for the same. And Android apps now don't have accessible app folders and demand root access (Recent Android rules). So in app UI is the most convenient method for font addition.

PR #2842, opened 2026-06-10
markor:// links, Improvements to opening external links

harshad1:
This PR adds support for 

1. opening markor:// links
2. specifying line number and view mode in the link


app/src/main/java/net/gsantner/markor/activity/DocumentActivity.java
```diff
@@ -155,19 +167,17 @@ protected void onNewIntent(Intent intent) {
         handleLaunchingIntent(intent);
     }
 
-    private void handleLaunchingIntent(final Intent intent) {
-        if (intent == null) return;
+    private void handleLaunchingIntent(final Intent rawIntent) {
+        if (rawIntent == null) return;
 
+        final Intent intent = normalizeIntent(rawIntent, this);
```
harshad1, line 173: This function normalizes the incoming intent

gsantner (approved):
Thanks, not sure if a custom protocol is a good idea, but if it makes Markor internally better functioning, OK

PR #2878, opened 2026-08-22
FileManager: Fix Favourite state not in sync, closes #2786

### LINKED ISSUE #2786: Changing the state of the "Favourite" checkbox fails to update the Star icon's visual state.

- [x] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it)_.

### Description

In the file "Info" dialog, the **Star icon** and the **"Favourite" checkbox** are bound to the same underlying boolean attribute. However, the state propagation is asymmetrical: clicking the star icon updates the checkbox, but **clicking the checkbox does not update the star icon**.

https://github.com/user-attachments/assets/7bb1a3e6-3b8c-449c-b341-22f4021141b9

### Steps to reproduce

1. Long-press a file (e.g., `QuickNote.md`) and select **Info**.
2. **Test Case 1 (Forward Sync):** Tap the **Star icon** to active it (Filled).
   * Observe the "Favourite" checkbox: It will correctly sync to **Checked**.
3. **Test Case 2 (Backward Sync - BUG):** Tap the **"Favourite" checkbox** to change its state (Unchecked/Checked).
   * Observe the Star icon on the left: **It remains in its previous state and fails to synchronize.**

### Information

Android version: 14
Device: Honor 70
App Version: 2.16.1

### Source

F-Droid

### Format / File type

Not specific

### Additional info / Log

```shell
## Expected behaviour
The states of the **Star icon** and the **"Favourite" checkbox** should be tightly coupled and synchronize instantly in **both directions** upon any interaction.

---

## Current behaviour
The state propagation is **unidirectional**.
* When the checkbox is the source of the interaction, its paired Star icon **fails to receive or render the new data state**. This is a classic **UI Binding Defect**.

---

## Technical Problem
The issue is likely that the **Star icon**'s view model is set up as a **One-Way (Read-Only) Binding** to the "isFavourite" property, but lacks a binding to the checkbox's user-driven click event. Both components should have equivalent, mutual update listeners.
```

harshad1:
Prevents file info fav state from going out of sync with the toggle

Fixes #2786

### DIFF
--- a/app/src/main/java/net/gsantner/markor/format/wikitext/WikitextSyntaxHighlighter.java
+++ b/app/src/main/java/net/gsantner/markor/format/wikitext/WikitextSyntaxHighlighter.java
@@ -45,7 +45,7 @@ public class WikitextSyntaxHighlighter extends SyntaxHighlighterBase {
     public final static Pattern ZIMHEADER = Pattern.compile(
             "^Content-Type: text/x-zim-wiki(\r\n|\r|\n)" +
                     "Wiki-Format: zim \\d+\\.\\d+(\r\n|\r|\n)" +
-                    "Creation-Date: \\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}[.+:-\\d]+");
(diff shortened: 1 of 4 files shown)


gsantner (approved):
Thanks!

gsantner (approved):

PR #2879, opened 2026-08-22
Search/Replace: Fix regex capture replacement, closes #2872

### LINKED ISSUE #2872: Regex replace doesn't work with the new find toolbar

- [x] The topic is **not** already reported at [Issues](https://github.com/gsantner/markor/issues?q=#js-issues-search). _(I've searched it and checked most recent issues)_.

### Description

regex substitution doesnt work with the find toolbar but does work with the old find modal ("line based search", activated with the filter button). with regex enabled, searching works as it should but on replace special regex characters like `$1` are printed literally

### Steps to reproduce

1. open the built in reference.md file
2. in the new find toolbar search for `(\n\n```)\n` with regex enabled. search works
3. try replacing with `$1whatever\n`. `$1` and `\n` are printed literally
4. try the same action with "line based search". substitution works correctly

### Information

Android version: 11
Device (phone model): motorola moto g8
App Version: 2.16.1

### Source

F-Droid

### Format / File type

Not specific

### Additional info / Log

```shell
I could not test this in the development version as the download link doesnt work.
```

harshad1:
This allows us to use $1 etc in regex replace again

Addresses #2872


gsantner (approved):
Thanks! Merging

